PR #604, opened 2014-12-08
Make logstash a direct data source for bosun

kylebrandt:
image

With regards to alerting on logstash data, there are basically two routes we can take.

 1. Have logstash send data to bosun (which relays it to OpenTSDB to store it)
 2. Have bosun query logstash

I have been assuming we would do option one at Stack. There are certainly advantages, Bosun still only needs to know about OpenTSDB. Also, things are likely much faster because they have already been precomputed into the data format you are after (time series).

However, one of the goals of bosun is to reduce the friction to creating alerts. One of my complaints about the Nagios model is that to change your alert, you frequently had to go and change the collectors (check plugins). I think directly querying logstash is analogous here. You likely have to go into logstash and setup a logstash output (http://logstash.net/docs/1.4.2/outputs/opentsdb) for what you want to alert on. 

This highlights some duplication of code (i.e. to add this I would have to replicate the band func, factor out common code). Also, more extensive would be refactoring so that bosun is no longer acting on opentsdb objects, and is rather acting on a time series interface. I'm not entirely sure what this interface *must* have. I know it at least needs a group (although something could be the null group, that would make it not very powerful) and datapoints. I'm exploring how I can satisfy that with logstash.


kadaan:
@PeterGrace this seems a natural fit for the Elasticsearch percolator feature.

kylebrandt:
New branch started, rebuilt on top of the changes post graphite https://github.com/bosun-monitor/bosun/tree/logstash

PeterGrace:
@kadaan thank you for pointing me towards percolator, I had not heard of it before!  I'm going to fiddle with it and see what fun stuff I can do with that.

PR #627, opened 2014-12-25
Add HTTP Post template func

kylebrandt:
Three arguments as strings, url, content-type (i.e. application/json), and the post data

### DIFF
--- a/cmd/bosun/sched/template.go
+++ b/cmd/bosun/sched/template.go
@@ -352,15 +352,31 @@ func (c *Context) LeftJoin(q ...interface{}) (interface{}, error) {
 	return matrix, nil
 }
 
-func (c *Context) HTTPGet(u string) string {
-	resp, err := http.Get(u)
+func (c *Context) HTTPGet(url string) string {
+	resp, err := http.Get(url)
 	if err != nil {
 		return err.Error()
 	}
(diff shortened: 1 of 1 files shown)


cmd/bosun/sched/template.go
```diff
@@ -367,3 +367,19 @@ func (c *Context) HTTPGet(u string) string {
 	}
 	return string(body)
 }
+
+func (c *Context) HTTPPost(u, t, d string) string {
```
mjibson: Blasphemous, I know, but u, t, and d need longer names. c is fine.

cmd/bosun/sched/template.go
```diff
@@ -367,3 +367,19 @@ func (c *Context) HTTPGet(u string) string {
...
+	resp, err := http.Post(u, t, bytes.NewBufferString(d))
+	if err != nil {
+		return err.Error()
+	}
+	if resp.StatusCode >= 300 {
+		return fmt.Sprintf("%v: returned %v", u, resp.Status)
+	}
+	defer resp.Body.Close()
```
mjibson: This needs to be before the previous if block. Otherwise the body will remain open if the response is >= 300 (and the body can contain data in that case). This also needs to be fixed in the Get handler above.

PR #637, opened 2015-01-07
In some cases,after parsing expr's String() not equal input text,so use original input when filter is true

shelocks:


shelocks:
For example input:(query1-query2)/2,after expr parse e.String() will return query1-query2/2

PR #645, opened 2015-01-09
cmd/scollector: Split mssql.replica.delay and add cluster name to votes

gbrayut:


cmd/scollector/collectors/sql_windows.go
```diff
@@ -50,6 +61,7 @@ const (
 )
 
 var (
+	cluster        string
```
mjibson: Name this sqlCluster or something uniqueish. This file is part of a package with many other collectors and we don't want to have any file reserve use of a common name.

cmd/scollector/collectors/sql_windows.go
```diff
@@ -18,6 +18,17 @@ func init() {
 	c.init = wmiInit(c, func() interface{} { return &[]Win32_PerfRawData_MSSQLSERVER_SQLServerGeneralStatistics{} }, `WHERE Name <> '_Total'`, &sqlQuery)
 	collectors = append(collectors, c)
 
+	var dstCluster []MSCluster_Cluster
+	var q = wmi.CreateQuery(&dstCluster, ``)
+	if err := queryWmiNamespace(q, &dstCluster, rootMSCluster); err != nil {
+		cluster = "None"
+	}
+	if len(dstCluster) != 1 {
```
mjibson: Should this be an elseif? The "None" above will never be used, otherwise.

gbrayut:
Fixed issues

PR #648, opened 2015-01-12
cmd/Scollector: Add hw.chassis.power.reading in Watts

gbrayut:


cmd/scollector/collectors/dell_hw.go
```diff
@@ -93,16 +93,24 @@ func c_omreport_ps() (opentsdb.MultiDataPoint, error) {
...
+		if len(fields) == 2 {
+			i_fields := strings.Split(fields[0], "Current")
+			v_fields := strings.Fields(fields[1])
+			if len(i_fields) < 2 && len(v_fields) < 2 {
+				return
+			}
+			id := strings.Replace(i_fields[0], " ", "", -1)
+			Add(&md, "hw.chassis.current.reading", v_fields[0], opentsdb.TagSet{"id": id}, metadata.Gauge, metadata.A, descDellHWCurrent)
+		} else {
+			v_fields := strings.Fields(fields[3])
```
mjibson: How do you know len(fields) >= 3?

gbrayut: Not sure what you mean. There are 2 fields for the Current and 6 for the Power. This is what we are parsing:

PS C:\windows\system32> omreport chassis pwrmonitoring -fmt ssv
Power Consumption Information

Power Consumption

Index;Status;Probe Name;Reading;Warning Threshold;Failure Threshold
2;Ok;System Board Pwr Consumption;210 W;1792 W;1974 W

Amperage
PS1 Current 1;0.4 A
PS2 Current 2;0.6 A

gbrayut: I can wrap in else if if len(fields) == 6 { .. } if you prefer

mjibson: Let's split this whole section into two parts then:

```
if len(fields) == 2 && strings.Contains(fields[0], "Current") {
    [...]
    Add(&md, ...)
} else if len(fields) == 6 && [whatever] {
    [...]
}
```

gbrayut:
Updated with fix

PR #649, opened 2015-01-12
Add ability to query a logstash populated elasticsearch

mjibson:


cmd/bosun/conf/conf.go
```diff
@@ -27,30 +27,31 @@ import (
...
-	RawText         string
-	Macros          map[string]*Macro
-	Lookups         map[string]*Lookup
-	Squelch         Squelches `json:"-"`
-	Quiet           bool
+	Name                string        // Config file name
+	CheckFrequency      time.Duration // Time between alert checks: 5m
+	TSDBHost            string        // OpenTSDB relay and query destination: ny-devtsdb04:4242
+	GraphiteHost        string        // Graphite query host: foo.bar.baz
+	LogstashElasticHost string
```
mjibson: Let's move these three out into their own group. They are special.

cmd/bosun/expr/funcs.go
```diff
@@ -22,6 +27,18 @@ func graphiteTagQuery(args []parse.Node) (parse.Tags, error) {
 	return t, nil
 }
 
+func logstashTagQuery(args []parse.Node) (parse.Tags, error) {
+	n, ok := args[0].(*parse.StringNode)
```
mjibson: No need to check for ok here.

cmd/bosun/sched/template.go
```diff
@@ -383,3 +386,39 @@ func (c *Context) HTTPPost(url, bodyType, data string) string {
...
+	}
+	results, err := service.SearchSource(s).Do()
+	if err != nil {
+		return nil, err
+	}
+	r := make([]*simplejson.Json, len(results.Hits.Hits))
+	for i, h := range results.Hits.Hits {
+		s := string([]byte(*h.Source)[:])
+		var err error
+		r[i], err = simplejson.Loads(s)
```
mjibson: Why do we need this?

cmd/bosun/sched/template.go
```diff
@@ -383,3 +386,39 @@ func (c *Context) HTTPPost(url, bodyType, data string) string {
 	}
 	return string(body)
 }
+
+//func LSBaseQuery(e *State, index_root, keystring, filter, interval, sduration, eduration string) (*elastic.SearchService, *elastic.SearchSource, []keyMatch, error) {
+
+func (c *Context) LSContextQuery(index_root, filter, interval, sduration, eduration string) (interface{}, error) {
+	var ks []string
+	for k, v := range c.Group {
+		ks = append(ks, strings.Join([]string{k, v}, ":"))
```
mjibson: `k + ":" + v` will suffice.

cmd/bosun/expr/funcs.go
```diff
@@ -68,6 +85,15 @@ var Graphite = map[string]parse.Func{
 	},
 }
 
+var LogstashElastic = map[string]parse.Func{
```
mjibson: Let's put all the elastic stuff in its own file. I'm going to move the tsdb and graphite stuff around, also.

cmd/bosun/expr/funcs.go
```diff
@@ -787,3 +813,225 @@ func Transpose(e *State, T miniprofiler.Timer, d *Results, gp string) (*Results,
...
+		if keys[depth].Pattern != nil && !keys[depth].Pattern.MatchString(key) {
+			continue
+		}
+		tags[keys[depth].Key] = key
+		desc(item, tags, depth+1, keys, r, e, ds)
+	}
+}
+
+func GenLSIndexes(host, index_root string, start, end time.Time) (string, error) {
+	resp, err := http.Get("http://" + host + "/_cat/indices")
```
mjibson: There's no way to get this out of the elastic package?

cmd/bosun/expr/funcs.go
```diff
@@ -787,3 +813,225 @@ func Transpose(e *State, T miniprofiler.Timer, d *Results, gp string) (*Results,
...
+	lines := strings.Split(string(body), "\n")
+	existing_indicies := make(map[string]bool)
+	for _, line := range lines {
+		f := strings.Fields(line)
+		if len(f) > 1 {
+			existing_indicies[f[2]] = true
+		}
+	}
+	var indicies []string
+	for d := start; d.Before(end); d = d.AddDate(0, 0, 1) {
```
mjibson: Should this also include d.Equal(end)?

cmd/bosun/expr/funcs.go
```diff
@@ -787,3 +813,225 @@ func Transpose(e *State, T miniprofiler.Timer, d *Results, gp string) (*Results,
...
+		i := fmt.Sprintf("%s-%s", index_root, d.Format("2006.01.02"))
+		if _, ok := existing_indicies[i]; ok {
```
mjibson: My spidey sense suggests this loop should be merged with the loop on line 1020. That is, look through all the indicies only one time, and do the checking right then. Split on the `-`, check the index_root against [0], parse the time, and see if it's within range. If so, add it to indices.

cmd/bosun/expr/funcs.go
```diff
@@ -787,3 +813,225 @@ func Transpose(e *State, T miniprofiler.Timer, d *Results, gp string) (*Results,
...
+	}
+	lines := strings.Split(string(body), "\n")
+	existing_indicies := make(map[string]bool)
+	for _, line := range lines {
+		f := strings.Fields(line)
+		if len(f) > 1 {
+			existing_indicies[f[2]] = true
+		}
+	}
+	var indicies []string
```
mjibson: `indices` is the correct spelling.

cmd/bosun/expr/funcs.go
```diff
@@ -787,3 +813,225 @@ func Transpose(e *State, T miniprofiler.Timer, d *Results, gp string) (*Results,
...
+			var err error
+			k.Pattern, err = regexp.Compile(k.RawPattern)
+			if err != nil {
+				return nil, err
+			}
+		}
+		keys = append(keys, k)
+	}
+	var filters []elastic.Filter
+	for _, key := range keys {
```
mjibson: This loop can be merged with the loop above.

cmd/bosun/expr/funcs.go
```diff
@@ -787,3 +813,225 @@ func Transpose(e *State, T miniprofiler.Timer, d *Results, gp string) (*Results,
...
+		if key.RawPattern != "" {
+			re := elastic.NewRegexpFilter(key.Key, key.RawPattern)
+			filters = append(filters, re)
+		}
+	}
+	if len(filters) > 0 {
+		and := elastic.NewAndFilter(filters...)
+		f := filtered.Filter(and)
```
mjibson: Why not *filtered = filtered.Filter(and)?

cmd/bosun/expr/funcs.go
```diff
@@ -787,3 +813,225 @@ func Transpose(e *State, T miniprofiler.Timer, d *Results, gp string) (*Results,
...
+		*filtered = f
+	}
+	return keys, nil
+}
+
+// LScount takes 6 arguments.
+// index_root is the root name of the index to hit, the format is expected to be fmt.Sprintf("%s-%s", index_root, d.Format("2006.01.02"))
```
mjibson: These docs need to be wrapped at 80 chars.

cmd/bosun/expr/funcs.go
```diff
@@ -787,3 +813,225 @@ func Transpose(e *State, T miniprofiler.Timer, d *Results, gp string) (*Results,
...
+		}
+		r.Results = append(r.Results, &Result{
+			Value: series,
+			Group: make(opentsdb.TagSet),
+		})
+		return r, nil
+	}
+	aggregation := elastic.NewTermsAggregation().Field(keys[len(keys)-1].Key)
+	aggregation = aggregation.SubAggregation("ts", ts)
+	if len(keys) > 1 {
```
mjibson: No need for this check. The for loop just won't do anything in that case.

cmd/bosun/expr/funcs.go
```diff
@@ -787,3 +813,225 @@ func Transpose(e *State, T miniprofiler.Timer, d *Results, gp string) (*Results,
...
+		for i := len(keys) - 2; i > -1; i-- {
```
mjibson: Why - 2 and not - 1?

cmd/bosun/expr/funcs.go
kylebrandt: There is will fix

cmd/bosun/sched/template.go
kylebrandt: It makes for a good template interface for accessing members of the returned document. For example:

```
	{{$ls := .LSContextQuery .Alert.Vars.index_root "" .Alert.Vars.interval .Alert.Vars.start ""}}
	<table>
	{{range $h := $ls}}
	<pre>{{$h}}</pre>
	<tr>
		<td>{{$h.Get "TargetUserName"}}</td>
		<td>{{($h.Get "IpAddress").String}}</td>
	</tr>
	{{end}}
	</table>
```

cmd/bosun/conf/conf.go
kylebrandt: What do you mean by group?

cmd/bosun/expr/funcs.go
```diff
@@ -787,3 +813,225 @@ func Transpose(e *State, T miniprofiler.Timer, d *Results, gp string) (*Results,
...
+	if err != nil {
+		return nil, err
+	}
+	top, ok := result.Aggregations.Terms("g_" + keys[0].Key)
+	if !ok {
+		return nil, fmt.Errorf("top key g_%v not found in result", keys[0].Key)
+	}
+	for _, b := range top.Buckets {
+		tags := make(opentsdb.TagSet)
+		key := fmt.Sprintf("%v", b.Key)
```
mjibson: fmt.Sprint(b.Key)

cmd/bosun/expr/funcs.go
```diff
@@ -787,3 +813,225 @@ func Transpose(e *State, T miniprofiler.Timer, d *Results, gp string) (*Results,
...
+		if keys[0].Pattern != nil && !keys[0].Pattern.MatchString(key) {
+			continue
+		}
+		tags[keys[0].Key] = key
+		desc(b, tags, 1, keys, r, e, ds)
```
mjibson: Instead of passing keys and depth, just pass key[1:], and use keys[0] always (after making sure it's non-empty).

cmd/bosun/expr/funcs.go
```diff
@@ -787,3 +813,225 @@ func Transpose(e *State, T miniprofiler.Timer, d *Results, gp string) (*Results,
...
+	}
+	return r, nil
+}
+
+func desc(b *elastic.AggregationBucketKeyItem, tags opentsdb.TagSet, depth int, keys []keyMatch, r *Results, e *State, ds opentsdb.Duration) {
```
mjibson: I'd make this a closure inside of LSCount. This will allow you to not pass a bunch of stuff.

kylebrandt: How do you do that? When I tried to do that I got an error with the name referencing itself within the function.

cmd/bosun/expr/funcs.go
```diff
@@ -787,3 +813,225 @@ func Transpose(e *State, T miniprofiler.Timer, d *Results, gp string) (*Results,
...
+	if !ok {
+		panic(fmt.Sprintf("%v not found at depth %v", "g_"+keys[depth].Key, strconv.Itoa(depth)))
+	}
+	for _, item := range n.Buckets {
+		key := fmt.Sprintf("%v", item.Key)
+		if keys[depth].Pattern != nil && !keys[depth].Pattern.MatchString(key) {
+			continue
+		}
+		tags[keys[depth].Key] = key
+		desc(item, tags, depth+1, keys, r, e, ds)
```
mjibson: Need to create a new tags here, not modify in place. This is a depth-first search, so after the first one tags will be fully populated and you will be only changing certain nodes.

cmd/bosun/conf/conf.go
mjibson: Put newlines around them.

cmd/bosun/sched/template.go
mjibson: https://github.com/likexian/simplejson-go/blob/master/simplejson.go#L79

kylebrandt: @mjibson Not sure what I'm supposed to be getting from that link?

mjibson: That's the implementation of the function you are calling. It does almost nothing. And we are importing a package for it?

kylebrandt: No, mostly for the various type and get methods that it provides to the template as demonstrated in my example above.

mjibson: That is all trivial code. And I don't trust that package at all. It hasn't been gofmt'd and has zero godocs. Each function is just a few lines, so we can write our own wrappers if we want. We will not be importing it.

kylebrandt:
Need to part and add "github.com/olivere/elastic" before merge

cmd/bosun/expr/funcs.go
```diff
@@ -887,3 +925,294 @@ func Transpose(e *State, T miniprofiler.Timer, d *Results, gp string) (*Results,
 	}
 	return &r, nil
 }
+
+type keyMatch struct {
```
mjibson: Make this lsKeyMatch so it's clear it's just for the elastic logstash stuff.

cmd/bosun/expr/funcs.go
```diff
@@ -887,3 +925,294 @@ func Transpose(e *State, T miniprofiler.Timer, d *Results, gp string) (*Results,
...
+				Group: tags.Copy(),
+			})
+			return nil
+		}
+		if len(keys) < 1 {
+			return nil
+		}
+		n, _ := b.Aggregations.Terms("g_" + keys[0].Key)
+		for _, item := range n.Buckets {
+			key := fmt.Sprintf("%v", item.Key)
```
mjibson: key := fmt.Sprint(item.Key)

PR #660, opened 2015-01-15
cmd/scollector: Fix issue with exim collector

kylebrandt:
Did not handle case where size was zero or less than 1KB

### DIFF
--- a/cmd/scollector/collectors/exim_linux.go
+++ b/cmd/scollector/collectors/exim_linux.go
@@ -25,9 +25,10 @@ func c_exim_mailq() (opentsdb.MultiDataPoint, error) {
 		f := strings.Fields(line)
 		if len(f) == 5 && f[4] == "TOTAL" {
 			Add(&md, "exim.mailq_count", f[0], nil, metadata.Gauge, metadata.EMail, "The number of emails in exim's mail queue.")
-			if len(f[1]) > 3 {
+			var multi int64 = 1
+			size, err := strconv.ParseInt(f[1], 10, 64)
+			if err != nil && len(f[1]) > 3 {
 				unit := f[1][len(f[1])-2:]
-				var multi int64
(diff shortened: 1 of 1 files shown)


cmd/scollector/collectors/exim_linux.go
```diff
@@ -25,34 +25,36 @@ func c_exim_mailq() (opentsdb.MultiDataPoint, error) {
...
 				unit := f[1][len(f[1])-2:]
-				var multi int64
 				switch unit {
 				case "KB":
 					multi = 1024
 				case "MB":
 					multi = 1024 * 1024
 				default:
-					return fmt.Errorf("error parsing size unit of exim's mail queue")
+					multi = 1
```
mjibson: You already do this on line 28. Either don't set it or remove this default case.

cmd/scollector/collectors/exim_linux.go
```diff
@@ -25,34 +25,36 @@ func c_exim_mailq() (opentsdb.MultiDataPoint, error) {
...
 				}
-				size, err := strconv.ParseInt(f[1][:len(f[1])-2], 10, 64)
-				if err != nil {
-					return err
+				if multi > 1 {
```
mjibson: No need to check for this since it must be > 1.

cmd/scollector/collectors/exim_linux.go
mjibson: Actually I think you should keep the error because this is the case where f[1] was not a number, meaning it has some suffix. Thus, if we get to this default case we have hit a suffix we do not recognize, and should thus return an error.

cmd/scollector/collectors/exim_linux.go
kylebrandt: There is after I get rid of the default

PR #666, opened 2015-01-19
cmd/bosun: Add rename func to rename tag keys

kylebrandt:
Useful for when you want to join on a tag key, but the tag key from one underlying series does not match the tag key of the other.

image

### DIFF
--- a/cmd/bosun/expr/funcs.go
+++ b/cmd/bosun/expr/funcs.go
@@ -64,6 +64,29 @@ func tagTranspose(args []parse.Node) (parse.Tags, error) {
 	return tags, nil
 }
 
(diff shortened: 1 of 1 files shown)


cmd/bosun/expr/funcs.go
```diff
@@ -64,6 +64,30 @@ func tagTranspose(args []parse.Node) (parse.Tags, error) {
 	return tags, nil
 }
 
+func tagRename(args []parse.Node) (parse.Tags, error) {
+	tags, err := tagFirst(args)
+	fmt.Println(tags)
```
mjibson: remove debug

PR #667, opened 2015-01-19
cmd/scollector: Add a collector for memcached stats

kylebrandt:


cmd/scollector/collectors/collectors.go
```diff
@@ -110,6 +110,14 @@ func Run(cs []Collector) chan *opentsdb.DataPoint {
 	return ch
 }
 
+type MetricMeta struct {
+	MetricName string
+	opentsdb.TagSet
```
mjibson: I think these should not be embedded, but they should keep the same name.

cmd/scollector/collectors/collectors.go
```diff
@@ -110,6 +110,14 @@ func Run(cs []Collector) chan *opentsdb.DataPoint {
 	return ch
 }
 
+type MetricMeta struct {
+	MetricName string
```
mjibson: How about just "Metric"

cmd/scollector/collectors/memcached_unix.go
```diff
@@ -0,0 +1,216 @@
...
+	"bosun.org/metadata"
+	"bosun.org/opentsdb"
+	"bosun.org/util"
+)
+
+func init() {
+	collectors = append(collectors, &IntervalCollector{F: c_memcached_stats})
+}
+
+var memcached_meta = map[string]MetricMeta{
```
mjibson: memcachedMeta

PR #685, opened 2015-02-03
change UTC to Local

wkshare:
clean *.ts utc
clean *.go utc
add timeZone to dev.conf


cmd/bosun/conf/conf.go
```diff
@@ -28,6 +28,7 @@ import (
 type Conf struct {
 	Vars
 	Name            string        // Config file name
+        TimeZone        string        // Config timezone, default UTC
```
mjibson: Need to gofmt this file.

cmd/bosun/conf/conf.go
```diff
@@ -363,6 +365,14 @@ func New(name, text string) (c *Conf, err error) {
 func (c *Conf) loadGlobal(p *parse.PairNode) {
 	v := c.Expand(p.Val.Text, nil, false)
 	switch k := p.Key.Text; k {
+        case "timeZone":
+                _, err := time.LoadLocation(v)
+                if err != nil {
+                        c.error(err)
+                        c.TimeZone = "UTC"
+                } else {
+
```
mjibson: Missing an action here. If UTC is the default, need to set c.TimeZone here.

cmd/bosun/conf/conf.go
mjibson: This should be a *time.Location

cmd/bosun/conf/conf.go
```diff
@@ -314,6 +315,7 @@ func New(name, text string) (c *Conf, err error) {
 	defer errRecover(&err)
 	c = &Conf{
 		Name:           name,
+                TimeZone:       "UTC",
```
mjibson: time.UTC

cmd/bosun/conf/conf.go
```diff
@@ -363,6 +365,14 @@ func New(name, text string) (c *Conf, err error) {
 func (c *Conf) loadGlobal(p *parse.PairNode) {
 	v := c.Expand(p.Val.Text, nil, false)
 	switch k := p.Key.Text; k {
+        case "timeZone":
+                _, err := time.LoadLocation(v)
```
mjibson: Preserve the value of this and set it as c.TimeZone

cmd/bosun/conf/conf.go
```diff
@@ -363,6 +365,14 @@ func New(name, text string) (c *Conf, err error) {
...
+                if err != nil {
+                        c.error(err)
+                        c.TimeZone = "UTC"
```
mjibson: This is already the case and can be removed.

wkshare: Let me adjust:)  thanks for your suggest.

cmd/bosun/conf/conf.go
wkshare: Emm...  Modify and PR.

wkshare:
Excuse me ：）

wkshare:
@mjibson Setting env TZ, This project all use Local.

wkshare:
@mjibson  test case, Let me to modify?

wkshare:
@mjibson give me suggest, Please :)

mjibson:
This PR won't be merged because the implementation is too flawed. For example, there is a setting to configure the timezone, but it is then not used, and instead just uses local everywhere always.

PR #694, opened 2015-02-05
cmd/bosun: Add rule links to alert status panels

kylebrandt:


cmd/bosun/web/static/js/state.ts
```diff
@@ -111,6 +111,18 @@ bosunApp.factory('status', ['$http', '$q', function($http: ng.IHttpService, $q:
...
+						'alert=' + encodeURIComponent(btoa(v.AlertDef)) +
+						'&template=' + encodeURIComponent(btoa(v.TemplateDef)) +
+						'&fromDate=' + encodeURIComponent(v.last.Time.format("YYYY-MM-DD")) +
+						'&fromTime=' + encodeURIComponent(v.last.Time.format("HH:mm"));
+					var groups: string[] = [];
+					angular.forEach(v.Group, (v, k) => {
+						groups.push(k + "=" + v);
+					});
+					if (groups.length > 0) {
+						v.RuleUrl = v.RuleUrl + '&template_group=' + encodeURIComponent(groups.join());
```
mjibson: Join needs an argument, '&'

mjibson: v.RuleUrl += is shorter than v.RuleUrl = v.RuleUrl + ...

mjibson: Does join need an argument here? `','`, I think.

kylebrandt: Mozilla says the default is a comma, which is what I wanted: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join

PR #696, opened 2015-02-06
cmd/bosun: Sleep CheckFrequency before running checks on start

kylebrandt:
Since Bosun relays OpenTSBD data this allows it to relay some information before startup. Without this we get flooded with unknowns on startup if Bosun has been down for more than checkFrequency.


cmd/bosun/main.go
```diff
@@ -40,6 +40,7 @@ var (
 	flagWatch    = flag.Bool("w", false, "watch .go files below current directory and exit; also build typescript files on change")
 	flagReadonly = flag.Bool("r", false, "readonly-mode: don't write or relay any OpenTSDB metrics")
 	flagQuiet    = flag.Bool("q", false, "quiet-mode: don't send any notifications except from the rule test page")
+	flagNoSleep  = flag.Bool("no-sleep", false, "don't sleep the length of check frequency before running the first run of checks after startup.")
```
mjibson: Remove `.` at eol.

Issue #695, opened 2015-02-06
question about filters collectors


wkshare:
if I input "scollector -f collectors.c_procstats_linux",  it can listen collectors.c_procstats_linux,which includes so many  items for me , I only need some of them. for example I only need  linux.loadavg. How can I do this ?

kylebrandt:
The problem is that it is one giant function, that really should be refactored into a few smaller functions (that could still all go in the same file).

So for example in https://github.com/bosun-monitor/bosun/blob/master/cmd/scollector/collectors/procstats_linux.go we have

```
func init() {
	collectors = append(collectors, &IntervalCollector{F: c_procstats_linux})
}
```

And it should maybe be:

```
func init() {
	collectors = append(collectors, &IntervalCollector{F: c_procstats_linux_vmstat})
        collectors = append(collectors, &IntervalCollector{F: c_procstats_linux_stat})
        // ... etc
}
```

I don't have time to work on that anytime soon, but we can take a PR that changes this collector - I agree it is too much in one function.

wkshare:
Thanks ！
How to reduce the sampling frequency？

wkshare:
@kylebrandt

kylebrandt:
Fixed in #697

PR #701, opened 2015-02-07
Procstats classification

wkshare:
#695 PR like this ? @kylebrandt


kylebrandt:
Matt will review this more carefully - but first you are going to want to make a new PR that only has the changes specific to the procstats change (whereas this one contains the timezone stuff you are working on as well).

wkshare:
We have in the use of the change, will continue to observe.
closed this PR.
New PR: #703

PR #703, opened 2015-02-08
update procstats_linux.go

wkshare:
Procstats classification #695


mjibson:
I understand what you want but this is not the way to do it. Instead, we should have a configuration option to filter the metrics, not just the collectors. Then we could get this functionality for all collectors, not just these.

PR #704, opened 2015-02-08
cmd/bosun: Limit unknown emails sent in a single check

kylebrandt:
This helps limit unknown spam when something goes bad. If there are more than Threshold email groups in a single check run, it will roll the remaining unknown emails into a single email that lists all the groups and their alert instances.

image

image

### DIFF
--- a/cmd/bosun/conf/conf.go
+++ b/cmd/bosun/conf/conf.go
@@ -27,30 +27,31 @@ import (
 
 type Conf struct {
 	Vars
-	Name            string        // Config file name
-	CheckFrequency  time.Duration // Time between alert checks: 5m
-	HTTPListen      string        // Web server listen address: :80
-	Hostname        string
-	RelayListen     string // OpenTSDB relay listen address: :4242
-	SMTPHost        string // SMTP address: ny-mail:25
-	SMTPUsername    string // SMTP username
-	SMTPPassword    string // SMTP password
-	Ping            bool
-	EmailFrom       string
-	StateFile       string
-	TimeAndDate     []int // timeanddate.com cities list
-	ResponseLimit   int64
-	SearchSince     opentsdb.Duration
-	UnknownTemplate *Template
-	Templates       map[string]*Template
-	Alerts          map[string]*Alert
-	Notifications   map[string]*Notification `json:"-"`
-	RawText         string
-	Macros          map[string]*Macro
-	Lookups         map[string]*Lookup
-	Squelch         Squelches `json:"-"`
-	Quiet           bool
-	NoSleep         bool
+	Name             string        // Config file name
+	CheckFrequency   time.Duration // Time between alert checks: 5m
+	HTTPListen       string        // Web server listen address: :80
+	Hostname         string
+	RelayListen      string // OpenTSDB relay listen address: :4242
+	SMTPHost         string // SMTP address: ny-mail:25
+	SMTPUsername     string // SMTP username
+	SMTPPassword     string // SMTP password
+	Ping             bool
+	EmailFrom        string
+	StateFile        string
+	TimeAndDate      []int // timeanddate.com cities list
+	ResponseLimit    int64
+	SearchSince      opentsdb.Duration
+	UnknownTemplate  *Template
+	UnknownThreshold int
+	Templates        map[string]*Template
+	Alerts           map[string]*Alert
+	Notifications    map[string]*Notification `json:"-"`
+	RawText          string
+	Macros           map[string]*Macro
+	Lookups          map[string]*Lookup
+	Squelch          Squelches `json:"-"`
+	Quiet            bool
+	NoSleep          bool
 
 	TSDBHost            string // OpenTSDB relay and query destination: ny-devtsdb04:4242
 	GraphiteHost        string // Graphite query host: foo.bar.baz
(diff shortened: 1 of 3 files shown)


kylebrandt:
Note: Fixed spelling of "suppressed" in the code.

cmd/bosun/conf/conf.go
```diff
@@ -442,6 +449,7 @@ func (c *Conf) loadGlobal(p *parse.PairNode) {
 			c.error(err)
 		}
 	default:
+		c.UnknownThreshold = 5
```
mjibson: This is not the correct place to set a default. It is above around line 317.

cmd/bosun/conf/conf.go
```diff
@@ -27,30 +27,31 @@ import (
...
+	SMTPUsername     string // SMTP username
+	SMTPPassword     string // SMTP password
+	Ping             bool
+	EmailFrom        string
+	StateFile        string
+	TimeAndDate      []int // timeanddate.com cities list
+	ResponseLimit    int64
+	SearchSince      opentsdb.Duration
+	UnknownTemplate  *Template
+	UnknownThreshold int64
```
mjibson: time.Duration is better

mjibson: We can safely change this to an int and it'll make some other code simpler.

cmd/bosun/sched/notify.go
```diff
@@ -102,12 +103,46 @@ func (s *Schedule) sendNotifications(rh *RunHistory, silenced map[expr.AlertKey]
 				s.AddNotification(ak, n, time.Now().UTC())
 			}
 		}
+		var c int64
+		tHit := false
+		oTSets := make(map[string]expr.AlertKeys)
 		for name, group := range ustates.GroupSets() {
-			s.unotify(name, group, n)
+			c++
+			if !tHit && (int64(len(ustates.GroupSets()))-c == 1) {
```
mjibson: Does it make sense to just count at the beginning? If len(ustates.GroupSets()) > conf.UnknownThreshold, then just send the one email? Maybe.

cmd/bosun/sched/notify.go
```diff
@@ -131,6 +166,30 @@ func (s *Schedule) notify(rh *RunHistory, st *State, n *conf.Notification) {
 	n.Notify(subject.Bytes(), body.Bytes(), s.Conf, string(st.AlertKey()), attachments...)
 }
 
+// utnotify is single notification for N unknown groups into a single notification
+func (s *Schedule) utnotify(groups map[string]expr.AlertKeys, n *conf.Notification) {
+	var total int
+	now := time.Now().UTC()
+	for _, group := range groups {
+		// Don't know what the following line does, just copied from unotify
+		s.Group[now] = group
```
mjibson: This is a feature we don't use yet, but talked about. We should just delete s.Group and add it back if people ask.

cmd/bosun/sched/notify.go
```diff
@@ -131,6 +166,30 @@ func (s *Schedule) notify(rh *RunHistory, st *State, n *conf.Notification) {
...
+	if err := unknownMultiGroup.Execute(body, struct {
+		Groups    map[string]expr.AlertKeys
+		Threshold int64
+	}{
+		groups,
+		s.Conf.UnknownThreshold,
+	}); err != nil {
+		log.Println(err)
+	}
+	fmt.Println(subject, body.String())
```
mjibson: remove debug

cmd/bosun/sched/notify.go
kylebrandt: Okay, will do that one in another PR though so the changes for this are clear in the merge (don't know how big it will be to remove it)

cmd/bosun/sched/notify.go
```diff
@@ -102,12 +103,47 @@ func (s *Schedule) sendNotifications(rh *RunHistory, silenced map[expr.AlertKey]
...
+		}
+		if len(oTSets) > 0 {
+			s.utnotify(oTSets, n)
 		}
 	}
 }
 
+var unknownMultiGroup = template.Must(template.New("unknownMultiGroup").Parse(`
+	<p>Threshold of {{.Threshold}} reached for unknown notifications. The following unknown
+	group emails were not sent.
```
mjibson: Use spaces around `{{` and `}}` consistently in this template.

cmd/bosun/sched/notify.go
```diff
@@ -102,12 +103,47 @@ func (s *Schedule) sendNotifications(rh *RunHistory, silenced map[expr.AlertKey]
...
 		}
-		for name, group := range ustates.GroupSets() {
-			s.unotify(name, group, n)
+		var c int
+		tHit := false
+		oTSets := make(map[string]expr.AlertKeys)
+		groupSets := ustates.GroupSets()
+		for name, group := range groupSets {
+			c++
+			if !tHit && (len(groupSets)-c == 1) {
```
mjibson:
```
tHit := false
for name, group := range groupSets {
	c++
	if c >= s.Conf.UnknownThreshold {
		if !tHit && (len(groupSets)-c == 1) {
			// If the threshold is hit but only 1 email remains, just send the normal unknown
			s.unotify(name, group, n)
			break
		}
		tHit = true
		oTSets[name] = group
	} else {
		s.unotify(name, group, n)
	}
}
```

PR #723, opened 2015-02-16
WIP Dependencies

captncraig:


cmd/bosun/web/static/js/bosun.js
```diff
@@ -103,18 +103,12 @@ bosunControllers.controller('BosunCtrl', ['$scope', '$route', '$http', '$q', '$r
     $scope.panelClass = function (status, prefix) {
         if (prefix === void 0) { prefix = "panel-"; }
         switch (status) {
-            case "critical":
```
mjibson: Is this `go fmt`d?

cmd/bosun/conf/conf.go
```diff
@@ -814,36 +793,41 @@ func (c *Conf) loadAlert(s *parse.SectionNode) {
...
-			c.errorf("crit tags (%v) and warn tags (%v) must be equal", tags, wtags)
 		}
 	}
-	a.tags = tags
 	a.returnType = ret
 	c.Alerts[name] = &a
 }
 
+func (c *Conf) ensureTagsMatch(exprs ...*expr.Expr) {
+	var tags eparse.Tags = nil
```
mjibson: Can omit ` = nil`

cmd/bosun/conf/conf.go
```diff
@@ -814,36 +793,41 @@ func (c *Conf) loadAlert(s *parse.SectionNode) {
...
+			continue
+		}
+		etags, err := ex.Root.Tags()
+		if err != nil {
+			c.error(err)
+		}
+		if tags == nil {
+			tags = etags
+		} else if !tags.Equal(etags) {
+			c.errorf("crit, warn, and depends tags must be equal. (%v != %v)", tags, etags)
```
mjibson: remove the `.`

cmd/bosun/conf/conf.go
```diff
@@ -1134,7 +1118,16 @@ func (c *Conf) AlertTemplateStrings() (*AlertTemplateStrings, error) {
...
 
 func (c *Conf) NewExpr(s string) (*expr.Expr, error) {
-	return expr.New(s, c.Funcs())
+	ex, err := expr.New(s, c.Funcs())
+	if err == nil {
+		switch ex.Root.Return() {
+		case eparse.TypeNumber, eparse.TypeScalar:
+			break
+		default:
+			c.errorf("expression must return a number")
```
mjibson: This function should either return an error in this case, or remove the error return from its signature and instead always call c.error if it encounters one. I'd favor the latter, since that's how everything else in this package works.

cmd/bosun/conf/conf.go
```diff
@@ -1134,7 +1118,17 @@ func (c *Conf) AlertTemplateStrings() (*AlertTemplateStrings, error) {
 }
 
 func (c *Conf) NewExpr(s string) (*expr.Expr, error) {
```
mjibson: This should no longer return an error ever.

cmd/bosun/sched/check.go
```diff
@@ -218,26 +221,33 @@ func (s *Schedule) CheckUnknown() {
 func (s *Schedule) CheckAlert(T miniprofiler.Timer, r *RunHistory, a *conf.Alert) {
 	log.Printf("check alert %v start", a.Name)
 	start := time.Now()
-	var warns expr.AlertKeys
-	crits, err := s.CheckExpr(T, r, a, a.Crit, StCritical, nil)
-	if err == nil {
-		warns, _ = s.CheckExpr(T, r, a, a.Warn, StWarning, crits)
+	var warns, crits expr.AlertKeys
+	ignore := expr.AlertKeys{}
```
mjibson: Can we make this a `map[expr.AlertKey]bool`? Easier to do membership checks.

cmd/bosun/sched/check.go
```diff
@@ -256,8 +266,26 @@ func (s *Schedule) CheckExpr(T miniprofiler.Timer, rh *RunHistory, a *conf.Alert
 		rh.Events[ak] = &Event{
 			Status: StError,
 		}
+		return nil, err
+	}
+	return results, err
+}
+
+func (s *Schedule) CheckExpr(T miniprofiler.Timer, rh *RunHistory, a *conf.Alert, e *expr.Expr, checkStatus Status, ignore expr.AlertKeys) (alerts expr.AlertKeys, err error) {
```
mjibson: It looks like checking to see if results are in ignore got removed during an intervening diff.

captncraig: `CheckExpr` still eveluates that. `ExecuteExpr` is the subfunction I factored out, because calculating the dependencies only requires that.

mjibson: Oh, I see. I forgot that ignore was already present.

cmd/bosun/sched/check.go
```diff
@@ -218,26 +221,33 @@ func (s *Schedule) CheckUnknown() {
...
+	var warns, crits expr.AlertKeys
+	ignore := expr.AlertKeys{}
+	d, err := s.ExecuteExpr(T, r, a, a.Depends)
+	if err != nil {
+		collect.Add("check.errs", opentsdb.TagSet{"metric": a.Name}, 1)
+		log.Println(err)
+	} else {
+		for _, dep := range d.Results {
+			ak := expr.NewAlertKey(a.Name, dep.Group)
+			ignore = append(ignore, ak)
```
mjibson: Ok, I figured some stuff out. We need to leave the ignore stuff alone and make a new data structure for this because we need to tell ignore and depends apart, since depends have to mark the alertkey as Unevaluated, but ignore doesn't do that.

captncraig:
@mjibson If we are just going to ignore the entire event when sending notifications and things, why not skip the additional flag and just not generate the event in the first place. Ignore accomplishes that I think. 

If we want to deal with an unevaluated event in some other way that might not be true, but if we are just going to throw it away, why not nip it in the bud?

mjibson:
Because a notification may already exist. Notifications can also be chained (send an email every 5 minutes until acknowledged). Also, once the unevaluated goes away, we need to resume/start sending notifications. The notification creation logic could then stay the same instead of having to teach it about unevaluated. Doing a simple disable at send time reduces the complication, I think.

cmd/bosun/sched/notify.go
```diff
@@ -59,6 +59,12 @@ func (s *Schedule) CheckNotifications(rh *RunHistory) time.Duration {
 			if st == nil {
 				continue
 			}
+			//if alert is currently unevaluated because of a dependency,
```
mjibson: Space after //. Comment is a sentence and should start with caps: https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

cmd/bosun/sched/check.go
```diff
@@ -284,8 +311,7 @@ Loop:
 		}
 		event := rh.Events[ak]
 		if event == nil {
-			event = new(Event)
```
mjibson: We need this line because `event` is used below, and it'll still be `nil` if not set here.

captncraig:
@mjibson updated.

cmd/bosun/web/static/js/bosun.js
mjibson: I'm aware that typescript is doing this, but can you remove this file from the PR?

mjibson:
Code LGTM. Need to remove the bosun.js change and squish again. If you use `git gui`, you can select the amend option to easily modify the last commit then push -f. This is the main reason I'd like to try out gerrit and see if it fits our workflow.

cmd/bosun/sched/sched_test.go
```diff
@@ -172,3 +172,55 @@ func TestCount(t *testing.T) {
 		},
 	})
 }
+
+func TestDepends(t *testing.T) {
+	testSched(t, &schedTest{
+		conf: `alert a {
+			depends = avg(q("avg:n{a=*}", "5m", "")) > 5
```
mjibson: Could you write another test for the overlap stuff?

opentsdb/tsdb.go
```diff
@@ -144,6 +144,24 @@ func (t TagSet) Tags() string {
 	return b.String()
 }
 
+// Two tagsets overlap if they:
+// 1. Have at least one key/value pair that matches
+// 2. Have no keys in common where the values do not match
+func Overlap(a TagSet, b TagSet) bool {
```
mjibson: make this a method on TagSet instead

mjibson:
LGTM after these fixes.

kylebrandt:
image

image

```
stateFile = ../bosun.state
checkFrequency = 30s
responseLimit = 5242880
unknownThreshold = 2
ping = true

notification default {
	email = kyle@stackoverflow.com
	print = true
}

template unknown {
    subject = {{.Name}}: {{.Group | len}} unknown alerts
    body = `
    <p>Time: {{.Time}}
    <p>Name: {{.Name}}
    <p>Alerts:
    {{range .Group}}
        <br>{{.}}
    {{end}}`
}

unknownTemplate = unknown

template generic {
	body = `<a href="{{.Ack}}">Acknowledge alert</a>
	<p>Alert definition:
	<p>Name: {{.Alert.Name}}
	<p>Crit: {{.Alert.Crit}}
	
	<p>Tags
	
	<table>
		{{range $k, $v := .Group}}
			{{if eq $k "host"}}
				<tr><td>{{$k}}</td><td><a href="{{$.HostView $v}}">{{$v}}</a></td></tr>
			{{else}}
				<tr><td>{{$k}}</td><td>{{$v}}</td></tr>
			{{end}}
		{{end}}
	</table>

	<p>Computation
	
	<table>
		{{range .Computations}}
			<tr><td>{{.Text}}</td><td>{{.Value}}</td></tr>
		{{end}}
	</table>`
	subject = {{.Last.Status}}: {{.Alert.Name}}: {{.Eval .Alert.Vars.q}} on {{.Group.host}}
}

alert ping.host {
	template = generic
	# also dns resolution
	$q = max(q("sum:bosun.ping.timeout{dst_host=*,host=grove}", "1m", ""))
	warn = $q
	warnNotification = default
}

alert os.cpu {
	template = generic
	# Also DNS resolution
	depends = max(rename(q("sum:bosun.ping.timeout{dst_host=*,host=grove}", "1m", ""), "host=source,dst_host=host"))
	$q = avg(q("avg:os.cpu{host=*}", "1m", ""))
	warn = $q < 99
	warnNotification = default

}
```

Didn't get unknown for the email (good), but still shows up on the dashboard (bad)

captncraig:
See #772

PR #728, opened 2015-02-18
cmd/scollector: fix multiple service tags for blade servers in vsphere

gbrayut:
Fixes issue with multiple service tags on new VM servers (see ny-bosun01 logs)

### DIFF
--- a/cmd/scollector/collectors/vsphere.go
+++ b/cmd/scollector/collectors/vsphere.go
@@ -171,6 +171,8 @@ func vsphereHost(v *vsphere.Vsphere, md *opentsdb.MultiDataPoint) error {
 				switch p.Name {
 				case "summary.hardware.otherIdentifyingInfo":
 					d := xml.NewDecoder(bytes.NewBufferString(p.Val.Inner))
+					// Blade servers may have multiple service tags. We want to use the last one.
+					var lastServiceTag string
 					for {
 						var t HostSystemIdentificationInfo
 						err := d.Decode(&t)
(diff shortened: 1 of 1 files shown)


cmd/scollector/collectors/vsphere.go
```diff
@@ -181,9 +183,10 @@ func vsphereHost(v *vsphere.Vsphere, md *opentsdb.MultiDataPoint) error {
 							return err
 						}
 						if t.IdentiferType.Key == "ServiceTag" {
-							metadata.AddMeta("", tags, "serialNumber", t.IdentiferValue, false)
+							lastServiceTag = t.IdentiferValue
 						}
 					}
+					metadata.AddMeta("", tags, "serialNumber", lastServiceTag, false)
```
mjibson: lastServiceTag could be empty if the for loop never ran or something. Instead, we should just have  `serviceTagDone := false` before the for loop, then if the bool is false, set the metadata where it is now, then set the bool to true.

cmd/scollector/collectors/vsphere.go
```diff
@@ -171,6 +171,8 @@ func vsphereHost(v *vsphere.Vsphere, md *opentsdb.MultiDataPoint) error {
 				switch p.Name {
 				case "summary.hardware.otherIdentifyingInfo":
 					d := xml.NewDecoder(bytes.NewBufferString(p.Val.Inner))
+					//Blade servers may have multiple service tags. Use the last one.
```
mjibson: space after //

mjibson: and if we make that change below then this should say the first instead of last

gbrayut: We want the last one... first one is for the blade enclosure.

mjibson: In that case, only add it if lastServiceTag != ""

PR #729, opened 2015-02-18
Properly chain notifications.

captncraig:
Making sure to queue next notification instead of requeueing same one again


captncraig:
testing slack

PR #735, opened 2015-02-23
cmd/bosun: Make an IP whitelist for who can write metrics and metadata t...

### LINKED ISSUE #730: Provide a method to prevent accidental data being pushed to Bosun
Frequently at SE dev data accidentally gets written to our prod instance. The vast majority of this could be fixed by an IP access list for our put routes.

Long term eventually we will want to thing about auth for the API, but I think this is a pretty easy feature with at least big win at SE.

kylebrandt:
...o bosun

If no IPs are provided, all IPs are allowed. Format is comma separated CIDR. Loopback IPs are always permitted. Fixes #730

### DIFF
--- a/cmd/bosun/conf/conf.go
+++ b/cmd/bosun/conf/conf.go
@@ -6,6 +6,7 @@ import (
 	htemplate "html/template"
 	"io/ioutil"
 	"log"
+	"net"
 	"net/mail"
 	"net/url"
 	"os"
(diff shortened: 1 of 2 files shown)


cmd/bosun/conf/conf.go
```diff
@@ -65,6 +67,23 @@ type Conf struct {
...
+type AuthorizedIPs []*net.IPNet
+
+func (authorizedIPs AuthorizedIPs) Authorized(ip net.IP) (authorized bool) {
+	if len(authorizedIPs) == 0 {
+		return true
+	}
+	if ip.IsLoopback() {
+		return true
+	}
+	for _, ipnet := range authorizedIPs {
```
captncraig: do you know if the net package caches results for Contains? I would hate for this check to become a bottleneck in receiving data.

kylebrandt: I don't know the implementation. But we can see how it performs when we build it :-)

Issue #594, opened 2014-12-04
Set hostname explicitly


mjibson:
<a href="https://github.com/lukaspustina"><img src="https://avatars.githubusercontent.com/u/398967?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [lukaspustina](https://github.com/lukaspustina)**
_Wednesday Nov 26, 2014 at 13:30 GMT_
_Originally opened as https://github.com/bosun-monitor/scollector/issues/152_

----

As far as I understand scollector, the hostname is set by asking the OS for the hostname. I would appreciate a parameter to explicitly set the hostname in order to control the host and domain part.

For example, while the hostname might be `host1.at.my.subdomina.my.organization.internal` it seems more useful to set the scollector host tag to something like `host1.at.my.subdomain` only.

lukaspustina:
I compiled scollector from source yesterday and I'm using `-hostname` to set the hostname. This works for all metrics, but `scollector.*` metrics which still seem to transmit just the /etc/hostname.
Any ideas?

mjibson:
That sounds like a bug.

kylebrandt:
scollector.* metrics which still seem to transmit just the /etc/hostname - related is that the filters `-f` don't apply to these either.

gbrayut:
self metrics have a different flag to disable:

-m=false: Disable sending of metadata.
-n=false: Disable sending of scollector self metrics.

captncraig:
I agree, I don't think there is a significant advantage to filtering scollector metrics beyond on/off

kylebrandt:
Agreed. Hostname still needs to be set the same for both though.

kylebrandt:
Fixed in #594

PR #737, opened 2015-02-23
#594 - scollector metrics don't honor hostname

captncraig:
#594


collect/collect.go
```diff
@@ -133,6 +133,10 @@ func Init(tsdbhost *url.URL, root string) error {
 	return InitChan(tsdbhost, root, make(chan *opentsdb.DataPoint))
 }
 
+func SetHostname(host string) {
+	osHostname = host
```
kylebrandt: Need ti checkClean() here?

Issue #575, opened 2014-12-04
Windows Processes To Monitor Should be Specified in Conf File, Not Process_Windows.go


mjibson:
<a href="https://github.com/kylebrandt"><img src="https://avatars.githubusercontent.com/u/1692624?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [kylebrandt](https://github.com/kylebrandt)**
_Tuesday Oct 07, 2014 at 14:33 GMT_
_Originally opened as https://github.com/bosun-monitor/scollector/issues/70_

----

// These are silly processes but exist on my machine, will need to update KMB
var processInclusions = regexp.MustCompile("chrome|powershell|scollector|SocketServer")
var serviceInclusions = regexp.MustCompile("WinRM")

Prooo-lly time to update that ^^

PR #742, opened 2015-02-24
Allow user to supply processes to watch on windows.

captncraig:
Previously, only linux processes were able to be added via config file. Windows users only got the hard coded list.

This change allows windows users of scollector to add a regex to the conf file for windows processes and services to monitor.

One thing to consider is that on linux, the `process` line should be of format: `command,name,regex` while on windows, it just expects any regex that matches the process names you want. 

This is a bit difficult to unify across os's, but this is my attempt.


captncraig:
#575

cmd/scollector/collectors/processes_linux.go
```diff
@@ -1,18 +1,29 @@
 package collectors
 
 import (
+	"bosun.org/metadata"
+	"bosun.org/opentsdb"
```
kylebrandt: Shouldn't be moved to top, these are always after a space after the normal packages. Using goimports?

captncraig: My editor fmts, but apparently not in _linux files. 

I checked build in my linux vm, but guess I should fmt too.

gbrayut:
PR looks good to me. I built and ran on windows and linux and both worked using the conf file as expected. Windows probably should use a bit more strict regexes to prevent monitoring the wrong processes since it doesn't have argument match like linux. Also it is matching against the output of `gwmi Win32_PerfRawData_PerfProc_Process | ft name` which does strange things like `powershell` then 'powershell#1,powershell#2,...` for multiple instances with the same name. In fact, we previously were only monitoring the First powershell instance (Doh!)

Proposed baseline conf file for Windows (I'll put this in our sre/scollector-build repo for tracking)

> process=^chrome
> process=^powershell
> process=^scollector
> process=^WinRM
> process=^MSSQLServer
> process=^WmiPrvSE
> process=^java
> process=^snmp
> process=^mscorsvw
> process=^SocketServer
> process=^ChatServer
> process=^StackServer
> process=^LogStasher
> process=^TCBuildAgent
> process=^dsm_(sa|om)

We can then add one off additions to individual servers (like vpxd on vsphere)

Example of graphs on devbosun http://goo.gl/dAXPGw

cmd/scollector/main.go
```diff
@@ -193,11 +189,9 @@ func main() {
 			collectors.Vsphere(user, pwd, host)
 		}
 	}
-	if len(procs) > 0 {
-		if err := collectors.WatchProcesses(procs); err != nil {
-			log.Fatal(err)
-		}
-	}
+	// Start all process collectors. This is platform specific.
```
kylebrandt: They don't start here they are just registered. `collectors.Run(c)` starts them.

gbrayut:
LGTM! Confirmed still works on linux and windows. Also confirmed failure when bad regex in conf:

2015/02/25 13:06:59 fatal: main.go:117: error parsing regexp: missing closing ]: `[{)`

PR #757, opened 2015-03-03
CheckUnknown should not panic on nil alert.

captncraig:
I think this should only be needed if you remove an alert and restart bosun. This will prevent a panic.

### DIFF
--- a/cmd/bosun/sched/check.go
+++ b/cmd/bosun/sched/check.go
@@ -198,6 +198,10 @@ func (s *Schedule) CheckUnknown() {
 				continue
 			}
 			a := s.Conf.Alerts[ak.Name()]
+			//alert doesn't exist anymore. Maybe status loaded from old db state.
+			if a == nil {
+				continue
+			}
 			t := a.Unknown
 			if t == 0 {
 				t = s.Conf.CheckFrequency * 2


mjibson:
Here's the panic that caused this.

```
2015/03/03 17:22:32 checkUnknown
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x68 pc=0x4f62d4]

goroutine 130069 [running]:
bosun.org/cmd/bosun/sched.(*Schedule).CheckUnknown(0xf84900)
        /home/teamcity/buildAgent/work/73e1f84984469e15/src/bosun.org/cmd/bosun/sched/check.go:201 +0x354
created by bosun.org/cmd/bosun/sched.func·027
        /home/teamcity/buildAgent/work/73e1f84984469e15/src/bosun.org/cmd/bosun/sched/sched.go:597 +0x52
```

cmd/bosun/sched/check.go
```diff
@@ -198,6 +198,10 @@ func (s *Schedule) CheckUnknown() {
 				continue
 			}
 			a := s.Conf.Alerts[ak.Name()]
+			//alert doesn't exist anymore. Maybe status loaded from old db state.
```
mjibson: There should be no case where an alert doesn't exist where we still have a state. The RestoreState method should take care of those cases. I think that this PR is an correct workaround, but I'd like a better fix.

mjibson: Rather, I'd like to understand why this bug happened before signing off on this fix.

captncraig:
I can't reproduce this anymore. Perhaps we should leave it 'broken' until it rears its head again.

mjibson:
Closing until this happens again.

PR #767, opened 2015-03-06
WIP: AWS

kylebrandt:


cmd/scollector/collectors/aws.go
```diff
@@ -0,0 +1,98 @@
...
+	if ecc == nil {
+		slog.Warning("Unable to login to EC2")
+	}
+	//elb := elb.New(creds, region, nil)
+
+	cw := cloudwatch.New(creds, region, nil)
+	if cw == nil {
+		slog.Warning("Unable to login to CloudWatch")
+	}
+	instances, err := AWSGetInstances(*ecc)
```
kylebrandt: Errors are not bools but actually a native type. And then you check if they are equal to nil. So all this needs to be refactored, see error handling in other collectors.

cmd/scollector/collectors/aws.go
```diff
@@ -0,0 +1,98 @@
...
+	if err != false {
+		// Do something useful in error
```
kylebrandt: Besides the bool not being correct, you are being silent here so you could have a problem and not get a message

cmd/scollector/collectors/aws.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+	for _, instance := range instances {
+		AWSGetCPU(*cw, &md, instance)
+	}
+	return md, nil
+
+}
+
+func AWSGetInstances(ecc ec2.EC2) ([]ec2.Instance, bool) {
```
kylebrandt: So should be ([]ec2.Instance, error) for the return types. If you need to create you own error you can use fmt.Errorf() - see  http://blog.golang.org/error-handling-and-go

cmd/scollector/collectors/aws.go
```diff
@@ -0,0 +1,98 @@
...
+	"fmt"
+	"time"
+
+	"bosun.org/metadata"
+	"bosun.org/opentsdb"
+	"bosun.org/slog"
+
+	"github.com/awslabs/aws-sdk-go/aws"
+	"github.com/awslabs/aws-sdk-go/gen/ec2"
+	//"github.com/awslabs/aws-sdk-go/gen/elb"
```
kylebrandt: WIP, but as a heads up before commit you need to remove all commented out code. Also using goimports gets rid of the pain of manually removing and adding imports when you need them

cmd/scollector/collectors/aws.go
```diff
@@ -0,0 +1,98 @@
...
+		F: func() (opentsdb.MultiDataPoint, error) {
+			return c_aws(accessKey, secretKey, region)
+		},
+		name: fmt.Sprintf("aws-%s", region),
+	})
+}
+
+func c_aws(accessKey, secretKey, region string) (opentsdb.MultiDataPoint, error) {
+	var md opentsdb.MultiDataPoint
+
```
kylebrandt: Also when no longer WIP we don't have any spaces within a function.

cmd/scollector/collectors/aws.go
```diff
@@ -0,0 +1,98 @@
...
+	resp, err := cw.GetMetricStatistics(&search)
+	if err != nil {
+		slog.Warning(err)
+	}
+	tags := opentsdb.TagSet{
+		"instance": *instance.InstanceID,
+	}
+
+	for _, datapoint := range resp.Datapoints {
+		Add(md, awsCPU, *datapoint.Average, tags, metadata.Gauge, metadata.Pct, "")
```
kylebrandt: Add a desc for the last field instead of a empty string. Desc should be a complete sentences ending in a period. They should also be constants defined elsewhere in the collector. See the exim collector for an example. would be something like const descAWSCPU = "The percentage of time that the CPU is utilized."

cmd/scollector/main.go
```diff
@@ -50,6 +50,7 @@ var (
 	flagHostname        = flag.String("hostname", "", "If set, use as value of host tag instead of system hostname.")
 	flagFreq            = flag.String("freq", "15", "Set the default frequency in seconds for most collectors.")
 	flagConf            = flag.String("conf", "", "Location of configuration file. Defaults to scollector.conf in directory of the scollector executable.")
+	flagaws             = flag.String("aws", "", `AWS keys and region, format: "access_key:secret_key@region".`)
```
kylebrandt: flagAWS

cmd/scollector/collectors/aws.go
kylebrandt: Oh, also it looks like maybe it is "over the past 60 seconds" based on search struct. IF that is the case put that in the description.

cmd/scollector/main.go
PeterGrace: fixed in 4a4734 (yet to be pushed)

cmd/scollector/collectors/aws.go
PeterGrace: Fixed in ec15af7

cmd/scollector/collectors/aws.go
PeterGrace: Corrected in 86a8090

cmd/scollector/collectors/aws.go
```diff
@@ -0,0 +1,302 @@
...
+		Interval: 60 * time.Second,
+		name:     fmt.Sprintf("aws-%s", region),
+	})
+}
+
+func c_aws(accessKey, secretKey, region string) (opentsdb.MultiDataPoint, error) {
+	var md opentsdb.MultiDataPoint
+	creds := aws.Creds(accessKey, secretKey, "")
+	if creds == nil {
+		slog.Info("Unable to make creds")
```
mjibson: You probably want to replace all of these `slog.Info("X")` calls with `return nil, fmt.Errorf("X")`

cmd/scollector/collectors/aws.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+	return md, nil
+}
+
+func AWSGetInstances(ecc ec2.EC2) ([]ec2.Instance, error) {
+	instancelist := []ec2.Instance{}
+	resp, err := ecc.DescribeInstances(nil)
+	if err != nil {
+		var Error error
+		Error = fmt.Errorf("Unable to describe EC2 Instances")
```
mjibson: return nil, fmt.Errorf("unable to ...")

cmd/scollector/collectors/aws.go
```diff
@@ -0,0 +1,302 @@
...
+		MetricName: aws.String("CPUUtilization"),
+		Period:     aws.Integer(60),
+		Statistics: []string{"Average"},
+		Namespace:  aws.String("AWS/EC2"),
+		Unit:       aws.String("Percent"),
+		Dimensions: []cloudwatch.Dimension{{Name: aws.String("InstanceId"), Value: instance.InstanceID}},
+	}
+	resp, err := cw.GetMetricStatistics(&search)
+	if err != nil {
+		slog.Info("Error getting Metric Statistics: ", err)
```
mjibson: Probably need a return here, instead of a slog.Info. The function can either return an error (if it expected to succeed), or return some empty data (if failure is acceptable and should just be silently ignored).

cmd/scollector/collectors/aws.go
PeterGrace: I changed these from slog.Warning to return error as in previous comments.

cmd/scollector/collectors/aws.go
PeterGrace: No more uncommented code at this time

cmd/scollector/collectors/aws.go
PeterGrace: fixed in 42b874b03705295d53ad9363d4317e6dae32a124 and ddc4da550b66f0be113ddb0f088b12f6dbfebf86.

cmd/scollector/collectors/aws.go
PeterGrace: Fixed in recent commit.

cmd/scollector/collectors/aws.go
PeterGrace: fixed in recent commit.

cmd/scollector/collectors/aws.go
PeterGrace: fixed in previous commits

cmd/scollector/collectors/aws.go
PeterGrace: Added descriptions in commit f1f1f05940fcf5718fb238930b6de4256f75b710

cmd/scollector/collectors/aws.go
```diff
@@ -0,0 +1,314 @@
...
+	awsNetwork            string = "aws.ec2.net.bytes"
+	awsEC2DiskBytes       string = "aws.ec2.disk.bytes"
+	awsEC2DiskOps         string = "aws.ec2.disk.ops"
+	awsStatusCheckFailed  string = "aws.ec2.status.failed"
+	awsELBLatencyMin      string = "aws.elb.latency.minimum"
+	awsELBLatencyMax      string = "aws.elb.latency.maximum"
+	awsELBLatencyAvg      string = "aws.elb.latency.average"
+	awsELBHostsHealthy    string = "aws.elb.hosts.healthy"
+	awsELBHostsUnHealthy  string = "aws.elb.hosts.unhealthy"
+	awsEC2CPUDesc         string = "The average CPU Utilization, gathered at a 60 second interval and averaged over five minutes."
```
kylebrandt: Desc goes first, so DescAWSEC2CPU ... kind of ugly in this case but that is our standard.

cmd/scollector/collectors/aws.go
```diff
@@ -0,0 +1,314 @@
...
+	awsCPU                string = "aws.ec2.cpu"
+	awsNetwork            string = "aws.ec2.net.bytes"
+	awsEC2DiskBytes       string = "aws.ec2.disk.bytes"
+	awsEC2DiskOps         string = "aws.ec2.disk.ops"
+	awsStatusCheckFailed  string = "aws.ec2.status.failed"
+	awsELBLatencyMin      string = "aws.elb.latency.minimum"
+	awsELBLatencyMax      string = "aws.elb.latency.maximum"
+	awsELBLatencyAvg      string = "aws.elb.latency.average"
+	awsELBHostsHealthy    string = "aws.elb.hosts.healthy"
+	awsELBHostsUnHealthy  string = "aws.elb.hosts.unhealthy"
```
kylebrandt: We generally only use constants for metric names when they are used for multiple collectors, such as osCPU, so these can just be strings in the Add func

cmd/scollector/collectors/aws.go
```diff
@@ -0,0 +1,314 @@
...
+	}
+	cw := cloudwatch.New(creds, region, nil)
+	if cw == nil {
+		return nil, fmt.Errorf("Unable to login to CloudWatch")
+	}
+	instances, err := AWSGetInstances(*ecc)
+	if err != nil {
+		slog.Info("No EC2 Instances found.")
+	}
+	loadbalancers, err := AWSGetLoadBalancers(*elb)
```
kylebrandt: loadBalancer since it is not one word, same with loadBalancers below

cmd/scollector/collectors/aws.go
```diff
@@ -0,0 +1,314 @@
...
+		AWSGetCPU(*cw, &md, instance)
+		AWSGetNetwork(*cw, &md, instance)
+		AWSGetDiskBytes(*cw, &md, instance)
+		AWSGetDiskOps(*cw, &md, instance)
+		AWSGetStatusChecks(*cw, &md, instance)
+	}
+	return md, nil
+}
+
+func AWSGetInstances(ecc ec2.EC2) ([]ec2.Instance, error) {
```
kylebrandt: I think these should be things like awsGetInstances since they don't need to be exported. We haven't been consistent with this in the past thought ... @mjibson ?

cmd/scollector/collectors/aws.go
```diff
@@ -0,0 +1,314 @@
...
+	awsEC2CPUDesc         string = "The average CPU Utilization, gathered at a 60 second interval and averaged over five minutes."
+	awsEC2NetBytesDesc    string = "The average bytes transmitted or received via network, gathered at a 60 second interval and averaged over five minutes."
+	awsEC2DiskBytesDesc   string = "The average bytes written or read via disk, gathered at a 60 second interval and averaged over five minutes."
+	awsEC2DiskOpsDesc     string = "The average disk operations, either written or read, gathered at a 60 second interval and averaged over five minutes."
+	awsEC2StatusCheckDesc string = "The EC2 Status Check, which includes both instance-level and system-level drill-down, gathered every 60 seconds."
+	awsELBLatencyDesc     string = "The minimum, maximum and average latency as reported by the load balancer, gathered at a 60 second interval and averaged over five minutes."
+	awsELBHostCountDesc   string = "The number of instances in what the Elastic Load Balancer considers a healthy state, gathered every 60 seconds."
+)
+
+func AWS(accessKey, secretKey, region string) {
```
kylebrandt: This one would need to be caps though since we need it exported to call it in main.go

cmd/scollector/collectors/aws.go
kylebrandt: You don't need "string" in these, can just be `awsEC2CPUDesc = "The avera...`

cmd/scollector/collectors/aws.go
```diff
@@ -0,0 +1,314 @@
...
+	}
+	for _, loadbalancer := range resp.LoadBalancerDescriptions {
+		lblist = append(lblist, loadbalancer)
+	}
+	return lblist, nil
+}
+
+func AWSGetCPU(cw cloudwatch.CloudWatch, md *opentsdb.MultiDataPoint, instance ec2.Instance) error {
+	search := cloudwatch.GetMetricStatisticsInput{
+		StartTime:  time.Now().UTC().Add(time.Second * -600),
```
kylebrandt: You should this be 600 when the period is 60?

cmd/scollector/collectors/aws.go
```diff
@@ -0,0 +1,314 @@
...
+
+import (
+	"fmt"
+	"time"
+
+	"bosun.org/metadata"
+	"bosun.org/opentsdb"
+	"bosun.org/slog"
+
+	"github.com/awslabs/aws-sdk-go/aws"
```
kylebrandt: Before merging these will need to be vendored using `party`. @mjibson can help you with this.

cmd/scollector/collectors/aws.go
```diff
@@ -0,0 +1,314 @@
...
+	for _, reservation := range resp.Reservations {
+		for _, instance := range reservation.Instances {
+			instancelist = append(instancelist, instance)
+		}
+	}
+	return instancelist, nil
+}
+
+func AWSGetLoadBalancers(lb elb.ELB) ([]elb.LoadBalancerDescription, error) {
+	lblist := []elb.LoadBalancerDescription{}
```
kylebrandt: lbList

cmd/scollector/collectors/aws.go
```diff
@@ -0,0 +1,314 @@
...
+		Statistics: []string{"Average"},
+		Namespace:  aws.String("AWS/EC2"),
+		Unit:       aws.String("Percent"),
+		Dimensions: []cloudwatch.Dimension{{Name: aws.String("InstanceId"), Value: instance.InstanceID}},
+	}
+	resp, err := cw.GetMetricStatistics(&search)
+	if err != nil {
+		return fmt.Errorf("Error getting Metric Statistics: %s", err)
+	}
+	tags := opentsdb.TagSet{
```
kylebrandt: Could be one line

cmd/scollector/collectors/aws.go
```diff
@@ -0,0 +1,314 @@
...
+		MetricName: aws.String("NetworkIn"),
+		Period:     aws.Integer(60),
+		Statistics: []string{"Average"},
+		Namespace:  aws.String("AWS/EC2"),
+		Unit:       aws.String("Bytes"),
+		Dimensions: []cloudwatch.Dimension{{Name: aws.String("InstanceId"), Value: instance.InstanceID}},
+	}
+	resp, err := cw.GetMetricStatistics(&search)
+	if err != nil {
+		return fmt.Errorf("Error getting Metric Statistics: %s", err)
```
kylebrandt: error strigns should start with a lower case - https://github.com/golang/go/wiki/CodeReviewComments#error-strings

cmd/scollector/collectors/aws.go
PeterGrace: fixed in 1970aac181173ba61a256cec02775721e4b289b5

cmd/scollector/collectors/aws.go
PeterGrace: Generally as in I'm not allowed to do it, or generally as in "you don't have to do it this way, but it is OK as-is?"

cmd/scollector/collectors/aws.go
PeterGrace: Corrected in f999d88edf694911b25ed65d4f07b2b85e3686db

cmd/scollector/collectors/aws.go
PeterGrace: CloudWatch is weird about how you get the data.  It only reports a datapoint every 5 mins unless you buy the premium per-minute monitoring.  Period can only be a multiple of 60, but it will give you a measurement every 5 minutes.  I set the time to 600 to ensure that every check produced at least one datapoint (since we could theoretically land a check at the zeroth second and get no datapoints.  However, in thinking about it, that doesn't matter too terribly because the next check would pick the datapoint up, so I should probably change this to 300 or 301.

cmd/scollector/collectors/aws.go
PeterGrace: fixed in 740cbe218ac7ffdf143b00dd3e90dec50413ff4d

cmd/scollector/collectors/aws.go
PeterGrace: What?

cmd/scollector/collectors/aws.go
PeterGrace: fixed with 207a74130f97a84c8b26bd0f1725f9bfa69060b9

cmd/scollector/collectors/aws.go
kylebrandt: Makes sense - keeping it at 600 seems fine to me. As you use the timestamp from the data which it looks like you do, and the value for that timestamp does *not* change. Duplicate data points make OpenTSDB quite irate if the values differ. In our case, we have an option that fixes duplicates by taking the last one - but we shouldn't be relying on that. If the datapoint's value can change for the more recent one, that you want to do something to ensure you only get the datapoint after the time period has passed.

cmd/scollector/collectors/aws.go
```diff
@@ -0,0 +1,314 @@
...
+	}
+	search.MetricName = aws.String("UnhealthyHostCount")
+	resp, err = cw.GetMetricStatistics(&search)
+	if err != nil {
+		return fmt.Errorf("error getting Metric Statistics: %s", err)
+	}
+	if resp.Datapoints == nil {
+		AddTS(md, awsELBHostsUnHealthy, time.Now().UTC().Unix(), 0, opentsdb.TagSet{"loadbalancer": *loadBalancer.LoadBalancerName}, metadata.Gauge, metadata.Count, descAWSELBHostCount)
+	} else {
+
```
kylebrandt: No blank line

cmd/scollector/collectors/aws.go
```diff
@@ -0,0 +1,314 @@
...
+	}
+	tags := opentsdb.TagSet{
+		"instance": *instance.InstanceID,
+	}
+	for _, datapoint := range resp.Datapoints {
+		AddTS(md, awsCPU, datapoint.Timestamp.Unix(), *datapoint.Average, tags, metadata.Gauge, metadata.Pct, descAWSEC2CPU)
+	}
+	return nil
+}
+func AWSGetNetwork(cw cloudwatch.CloudWatch, md *opentsdb.MultiDataPoint, instance ec2.Instance) error {
```
kylebrandt: These and some of the following functions seem to share a lot of code - I bet you can refactor these into a generic function where you just pass the Unit and maybe the metadata / tag information. Or at least a func to initialize cloudwatch.GetMetricStatisticsInput since most of the values are the same.

cmd/scollector/collectors/aws.go
PeterGrace: In every case I've seen thus far, the duplicated data points have all been exactly the same, so I think we're safe from differing timeseries duplicates.

cmd/scollector/collectors/aws.go
PeterGrace: Fixed in 8384f3558035cc400f78d6b278f08c373787e695

cmd/scollector/collectors/aws.go
PeterGrace: I've put a bit of time into this thinking about it, and really the only field that will consistently stay the same is EndTime; Namespace changes by module, Unit changes by metric, statistics changes by metric.  It does seem like a lot of duplicated code, but if I make a separate function just to do it, I'll essentially have a function with 8 arguments doing functionally the same thing as the struct initialization which might add readability issues.  I'm open to doing it that way but it feels like neither way is a fantastic solution.

kylebrandt: Fair enough, this what at a glance so I'm okay with whatever conclusion.

cmd/scollector/collectors/aws.go
mjibson: Yes, these should all be awsXXX instead of AWSXXX so that they are unexported.

PeterGrace: Fixed in 85794f0bb624956985af1e8cd57ee43367b5ff39

cmd/scollector/collectors/aws.go
```diff
@@ -0,0 +1,313 @@
...
+	"bosun.org/slog"
+
+	"bosun.org/_third_party/github.com/awslabs/aws-sdk-go/aws"
+	"bosun.org/_third_party/github.com/awslabs/aws-sdk-go/gen/cloudwatch"
+	"bosun.org/_third_party/github.com/awslabs/aws-sdk-go/gen/ec2"
+	"bosun.org/_third_party/github.com/awslabs/aws-sdk-go/gen/elb"
+)
+
+const (
+	awsCPU                = "aws.ec2.cpu"
```
mjibson: Run this whole list through sort

cmd/scollector/collectors/aws.go
```diff
@@ -0,0 +1,313 @@
...
+		return nil, fmt.Errorf("unable to describe ELB Balancers")
+	}
+	for _, loadBalancer := range resp.LoadBalancerDescriptions {
+		lbList = append(lbList, loadBalancer)
+	}
+	return lbList, nil
+}
+
+func awsGetCPU(cw cloudwatch.CloudWatch, md *opentsdb.MultiDataPoint, instance ec2.Instance) error {
+	search := cloudwatch.GetMetricStatisticsInput{
```
mjibson: There are many calls to GetMetricStatisticsInput. They each make their own time, which means if one call takes 10 seconds, the results will be skewed with other results by that much. Would it be more useful to instead record now in the main c_aws function, and pass that down to all of these other functions? Then they could compute StartTime based on that, and each function here would be getting data from the same time point.

cmd/scollector/collectors/aws.go
```diff
@@ -0,0 +1,313 @@
...
+		MetricName: aws.String("Latency"),
+		Period:     aws.Integer(60),
+		Statistics: []string{"Average", "Minimum", "Maximum"},
+		Namespace:  aws.String("AWS/ELB"),
+		Unit:       aws.String("Seconds"),
+		Dimensions: []cloudwatch.Dimension{{Name: aws.String("LoadBalancerName"), Value: loadBalancer.LoadBalancerName}},
+	}
+	resp, err := cw.GetMetricStatistics(&search)
+	if err != nil {
+		return fmt.Errorf("error getting Metric Statistics: %s", err)
```
mjibson: `return err` is sufficient. same for all others.

cmd/scollector/collectors/aws.go
PeterGrace: corrected in d81bc05509d693a7f2b73915a5e4d4f93495a4e0

PeterGrace: Aaaand also bb29209c31f32714722f9186bbc933d37d967c4e to fix spacing (doh)

PeterGrace: reverted other two commits and committed 888548d727795c015d8e5d58abada23f7b594ad6 to alphabetize constants

cmd/scollector/collectors/aws.go
PeterGrace: fixed in 188b31ff9f8e3983eb2f9717598be47a045d5642

cmd/scollector/collectors/aws.go
PeterGrace: @mjibson re: your quote about GetMetricStatisticsInput, it doesn't actually matter after thinking about it -- cloudwatch has already digested the data and I get X data points every Y minutes, so the cloudwatch query retrieves a predigested set of data.  I could theoretically initialize Start and the search struct in c_aws and pass that along to all the functions but we get back to that earlier conversation of "is it more readable that way?"  I'm willing to go either way on that.

PR #772, opened 2015-03-09
Dependencies v2

captncraig:
The other one was only getting buggier and buggier.

This version is based off of newer code and should be a tad simpler, and have more tests.


captncraig:
This version still assumes alerts will run serially and in order. We can address the issue of running until consistent if/when we parallelize checks.

mjibson:
Looks like this needs a rebase.

kylebrandt:
Wouldn't rebase until my messed up vendoring is fixed.

cmd/bosun/expr/expr.go
```diff
@@ -38,6 +38,14 @@ type State struct {
...
 	logstashQueries      []elastic.SearchService
 	logstashElasticHosts []string
+
+	History AlertStatusProvider
+}
+
+// Alert Status Provider is used to provide information about alert results.
+// This facilitates alerts referencing other alerts, even when they go unknown or unevaluated.
+type AlertStatusProvider interface {
+	GetUnknownAndUnevaluatedAlertKeys(alertName string) (unknown, unevaluated []AlertKey)
```
mjibson: What do you think about inverting this API? Instead of this function, have two others: `IsUnknown(AlertKey)` and `IsUnevaluated(AlertKey)`. I think this would make the usage in conf.go simpler.

cmd/bosun/sched/check.go
```diff
@@ -198,58 +183,110 @@ func (s *Schedule) RunHistory(r *RunHistory) {
...
+	for _, a := range s.Conf.Alerts {
+		s.CheckAlert(T, r, a)
+	}
+	d := time.Since(start)
+	s.RunHistory(r)
+	<-s.checkRunning
+	return d, nil
+}
+
+func (s *Schedule) findStaleAlerts(now time.Time) []expr.AlertKey {
```
mjibson: Stale is a new term. What about `getUnknownAlerts` instead?

cmd/bosun/sched/check.go
```diff
@@ -198,58 +183,110 @@ func (s *Schedule) RunHistory(r *RunHistory) {
...
+		// Good, we've got the lock.
+	default:
+		return 0, fmt.Errorf("check already running")
+	}
+	r := s.NewRunHistory(now, cache.New(0))
+	start := time.Now()
+	for _, ak := range s.findStaleAlerts(now) {
+		r.Events[ak] = &Event{Status: StUnknown}
+	}
+	for _, a := range s.Conf.Alerts {
```
mjibson: How are recursive dependencies handled? I don't see anything that does it, but maybe there's some fancy algorithm that was not obvious to me.

kylebrandt:
I no longer have the the "ghost" unknowns from the rename function. However the dependency is failing to suppress the unknown:

image

Config:
```
alert ping.host {
    template = generic
    $q = max(rename(q("sum:bosun.ping.timeout{dst_host=*,host=ny-kbrandt02}", "1m", ""), "host=source,dst_host=host"))
    warn = $q
    warnNotification = default
}

alert os.cpu {
    template = generic
    depends = alert("ping.host", "warn")
    $q = avg(q("avg:os.cpu{host=*}", "1m", ""))
    warn = $q < 99
    warnNotification = default
}
```

kylebrandt:
So I think the next step would be to make a test for dependent alerts not going unknown if the dependency alert is "True".

A test case after that would be if the dependency goes "unknown" and the the dependent alert is an abnormal status (include unknown). In that case we should get only the dependency unknown notification and the dependent alert is suppressed. 

Note: In the above example, os.cpu is the dependent alert and ping.host is the dependency [alert].

kylebrandt:
Unknown still triggers, dashboard looks like it did before, here is the output of `/api/alerts`:
```
{
Groups: {
NeedAck: [
{
Active: true,
Status: "unknown",
Subject: "unknown - os.cpu",
Len: 1,
Children: [
{
Active: true,
Status: "unknown",
Alert: "os.cpu",
AlertKey: "os.cpu{host=ny-kbrandt01}",
Ago: "2015-03-12T14:05:02.561022842Z"
}
]
},
{
Active: true,
Status: "warning",
Subject: "warning - ping.host",
Len: 1,
Children: [
{
Active: true,
Status: "warning",
Subject: "warning: ping.host: on ny-kbrandt01",
Alert: "ping.host",
AlertKey: "ping.host{host=ny-kbrandt01,source=ny-kbrandt02}",
Ago: "2015-03-12T14:03:32.557630494Z"
}
]
}
]
},
TimeAndDate: null,
Silenced: { }
}
```

kylebrandt:
Oh! But it looks like I didn't get the os.cpu email. So that part worked, just not that dashboard. That is a progress

kylebrandt:
With the following config, I started Bosun (with an existing state file, but one generated from this revision) and I was in the following state:

image
(The crit for mem and warn for cpu were both for ny-kbrandt01). 

I then shutdown scollector and disabled ping and got the following: 
image

Which after more time became:

image

cmd/bosun/sched/sched.go
```diff
@@ -589,13 +589,6 @@ func (s *Schedule) Run() error {
 		go s.PingHosts()
 	}
 	go s.Poll()
-	go func() {
-		// Sleep before starting unknown checks. If Bosun was down for a little
-		// while then all the state.Touched times will be old and a bunch of
-		// unknowns will go off because it hasn't run all the checks yet.
-		time.Sleep(s.Conf.CheckFrequency)
```
mjibson: This feature needs to be added back. This is because if bosun is down for 10 minutes and scollectors all around have been queueing data, it will take a few minutes for that data to get back into bosun. Thus, when rules are run, they won't return anything, and then get marked as unknown.

kylebrandt: @captncraig The feature Matt means is delaying the unknown check at startup. Basically we want to give Bosun time to collect queued data before firing off a bunch of unknowns.

cmd/bosun/sched/check.go
```diff
@@ -200,58 +185,128 @@ func (s *Schedule) RunHistory(r *RunHistory) {
...
-		warns, _ = s.CheckExpr(T, r, a, a.Warn, StWarning, crits)
+		deps = filterDependencyResults(d)
+		crits, err := s.CheckExpr(T, r, a, a.Crit, StCritical, nil)
+		if err == nil {
+			warns, _ = s.CheckExpr(T, r, a, a.Warn, StWarning, crits)
+		}
 	}
+	markDependenciesUnevaluated(r.Events, deps, a.Name)
 	collect.Put("check.duration", opentsdb.TagSet{"name": a.Name}, time.Since(start).Seconds())
 	log.Printf("check alert %v done (%s): %v crits, %v warns", a.Name, time.Since(start), len(crits), len(warns))
```
kylebrandt: Know that we moved checkUnknown, can this also log the number of unknowns + unevaluated alerts?

cmd/bosun/sched/check.go
```diff
@@ -200,58 +186,135 @@ func (s *Schedule) RunHistory(r *RunHistory) {
...
+	s.RunHistory(r)
+	<-s.checkRunning
+	return d, nil
+}
+
+var bosunStartupTime = time.Now()
+
+func (s *Schedule) findUnknownAlerts(now time.Time) []expr.AlertKey {
+	keys := []expr.AlertKey{}
+	if time.Now().Sub(bosunStartupTime) < 1*time.Minute {
```
mjibson: I think we should use `s.Conf.CheckFrequency` instead of time.Minute, unless you've discussed something with Kyle already.

cmd/bosun/sched/check.go
```diff
@@ -200,58 +186,135 @@ func (s *Schedule) RunHistory(r *RunHistory) {
...
-	}()
-	results, _, err := e.Execute(rh.Context, rh.GraphiteContext, s.Conf.LogstashElasticHosts, rh.Cache, T, rh.Start, 0, a.UnjoinedOK, s.Search, s.Conf.AlertSquelched(a))
+		if !math.IsNaN(n) && n != 0 {
+			filtered = append(filtered, r)
+		}
+	}
+	return filtered
+}
+
+func markDependenciesUnevaluated(events map[expr.AlertKey]*Event, deps expr.ResultSlice, alert string) (depCount, unknownCount int) {
```
mjibson: `depCount` is used here, but `unevalCount` is returned. Change `depCount` to `unevalCount`.

cmd/bosun/sched/check.go
```diff
@@ -200,58 +186,135 @@ func (s *Schedule) RunHistory(r *RunHistory) {
...
+	unevalCount, unknownCount := 0, 0
+	for ak, ev := range events {
+		if ak.Name() == alert {
```
mjibson: Invert this if to avoid the indentation below.

cmd/bosun/sched/check.go
```diff
@@ -200,58 +186,135 @@ func (s *Schedule) RunHistory(r *RunHistory) {
...
 
 func (s *Schedule) CheckAlert(T miniprofiler.Timer, r *RunHistory, a *conf.Alert) {
 	log.Printf("check alert %v start", a.Name)
 	start := time.Now()
-	var warns expr.AlertKeys
-	crits, err := s.CheckExpr(T, r, a, a.Crit, StCritical, nil)
+	var warns, crits expr.AlertKeys
+	d, err := s.executeExpr(T, r, a, a.Depends)
+	var deps expr.ResultSlice
 	if err == nil {
```
mjibson: What about if err != nil?

cmd/bosun/sched/check.go
```diff
@@ -200,58 +186,135 @@ func (s *Schedule) RunHistory(r *RunHistory) {
...
-		warns, _ = s.CheckExpr(T, r, a, a.Warn, StWarning, crits)
+		deps = filterDependencyResults(d)
+		crits, err = s.CheckExpr(T, r, a, a.Crit, StCritical, nil)
+		if err == nil {
+			warns, _ = s.CheckExpr(T, r, a, a.Warn, StWarning, crits)
+		}
 	}
+	depCount, unknownCount := markDependenciesUnevaluated(r.Events, deps, a.Name)
```
mjibson: What happens in the case where deps and crit are the same?

```
alert a {
  $q = avg(q("avg:cpu{host=*}", "5m", "")) > 5
  deps = $q
  crit = $q
}
```

and then it goes unknown?

cmd/bosun/sched/check.go
```diff
@@ -200,58 +186,135 @@ func (s *Schedule) RunHistory(r *RunHistory) {
...
+	default:
+		return 0, fmt.Errorf("check already running")
+	}
+	r := s.NewRunHistory(now, cache.New(0))
+	start := time.Now()
+	for _, ak := range s.findUnknownAlerts(now) {
+		r.Events[ak] = &Event{Status: StUnknown}
+	}
+	for _, a := range s.Conf.Alerts {
+		s.CheckAlert(T, r, a)
```
mjibson: I don't understand how unknowns work with recursive dependencies. Say we have C depends on B depends on A. A goes unknown. C is checked. It should be marked uneval, but I can't understand how it would be.

opentsdb/tsdb.go
```diff
@@ -37,6 +45,21 @@ type Response struct {
 	DPS           map[string]Point `json:"dps"`
 }
 
+func (r *Response) Copy() *Response {
+	newR := Response{}
+	newR.Metric = r.Metric
+	newR.Tags = r.Tags.Copy()
+	newR.AggregateTags = make([]string, len(r.AggregateTags))
+	for i, t := range r.AggregateTags {
```
mjibson: `copy(newR.AggregateTags, r.AggregateTags)` is simpler

cmd/bosun/sched/check.go
mjibson: I've tried to add a test for this case:

https://github.com/bosun-monitor/bosun/commit/f4acb6a6d8a465581ed95ff214f3c2f9f92a8192

But I'm not sure the test is correct yet. I'll refine it tomorrow.

cmd/bosun/sched/check.go
captncraig: I didn't change how errors are handled. CheckAlert doesn't even have an error return. In executeExpr if there is an error, we put an error status on the schedule and skip further evaluation.

cmd/bosun/sched/check.go
captncraig: That is pretty silly to set up, and I would call it a mistake anytime I saw it in real life. That alert has two real states: Normal and unevaluated. It may go crit/unevaluated, warn/unevaluated or unknown/unevaluated, but none of those will notify or show up on the dashboard.

Short answer: don't do that.

cmd/bosun/sched/check.go
captncraig: Right now it depends on checks running in order, which now that I look at it, is not always going to happen, but usually does. As long as they are running sequentially, A gets checked, B gets checked and marked uneval, and then C gets checked and gets marked bbecause of B.

cmd/bosun/sched/check.go
mjibson: You changed the var name on the caller. I meant to change it in the named return values here.

mjibson:
Fix that little named return thing then LGTM.

PR #775, opened 2015-03-10
cmd/bosun: Fix silence logic

### LINKED ISSUE #769: Silence Didn't Suppress Notification
image

kylebrandt:
Fixes #769

### DIFF
--- a/cmd/bosun/sched/check_test.go
+++ b/cmd/bosun/sched/check_test.go
@@ -1,7 +1,12 @@
 package sched
 
 import (
+	"fmt"
+	"net/http"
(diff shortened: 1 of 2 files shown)


mjibson:
How about a test to go with this? I realize it might be longer than the code, but I think it's super important that we get into the habit of writing tests when bosun has bugs in its operational logic. Bosun is complicated enough that we will change some little thing somewhere else, and we need to make sure that change doesn't affect other parts. Tests are a good way to do that.

kylebrandt:
Do you have something I can look at it that will give me a starting point to write the test?

PR #776, opened 2015-03-10
cmd/scollector: Add duration to interval.go

gbrayut:


cmd/scollector/collectors/interval.go
```diff
@@ -48,10 +49,15 @@ func (c *IntervalCollector) Run(dpchan chan<- *opentsdb.DataPoint) {
 		}
 		next := time.After(interval)
 		if c.Enabled() {
+			timeStart := time.Now()
 			md, err := c.F()
+			timeFinish := time.Now()
```
mjibson: timeFinish := time.Since(timeStart)

cmd/scollector/collectors/interval.go
```diff
@@ -48,10 +49,15 @@ func (c *IntervalCollector) Run(dpchan chan<- *opentsdb.DataPoint) {
...
+			result := "success"
```
mjibson: Result needs to be a timeseries value we always report instead of a tag. If it's a tag, then it's too easy for unknowns to happen. So make another metric called `scollector.collector.error` and have it be 1 or 0 for error (1) or not (0).

cmd/scollector/collectors/interval.go
```diff
@@ -48,10 +49,15 @@ func (c *IntervalCollector) Run(dpchan chan<- *opentsdb.DataPoint) {
...
 			if err != nil {
 				slog.Errorf("%v: %v", c.Name(), err)
+				result = "error"
 			}
+			Add(&md, "scollector.collect.duration", float64(timeFinish.Sub(timeStart))/float64(time.Millisecond), opentsdb.TagSet{"collector": c.Name(), "result": result}, metadata.Gauge, metadata.MilliSecond, "Duration in milliseconds and result of collector runs in each interval.")
```
mjibson: Change name to `scollector.collector.duration`. The `scollector.collect` namespace is reserved for the collect package, which this isn't. Remove result from the tag, and use `timeFinish.Seconds()` as the value, and fix the description string to list seconds as the unit.

gbrayut:
Updated with requested changes. I also am running it as a second instance pointed at devbosun. See http://goo.gl/pzc9tA

cmd/scollector/collectors/interval.go
```diff
@@ -48,10 +49,16 @@ func (c *IntervalCollector) Run(dpchan chan<- *opentsdb.DataPoint) {
...
 		if c.Enabled() {
+			timeStart := time.Now()
 			md, err := c.F()
+			timeFinish := time.Since(timeStart)
+			result := 0
 			if err != nil {
 				slog.Errorf("%v: %v", c.Name(), err)
+				result = 1
 			}
+			Add(&md, "scollector.collector.duration", timeFinish.Seconds(), opentsdb.TagSet{"collector": c.Name()}, metadata.Gauge, metadata.Second, "Duration in seconds for each collector run.")
```
mjibson: Since the tags are the same here, we can set `tags := opentsdb.TagSet{"collector": c.Name()}` above the for loop on line 45.

Issue #724, opened 2015-02-17
Ability to monitor TCP and UDP listening connections on Windows


gbrayut:
We had an issue with source port exhaustion on Windows due to UDP listeners consuming all the listening ports. It appears there may be a WMI counter for [TCP sockets](http://serverfault.com/questions/567752/what-wmi-counters-can-indicate-port-exhaustion) (not exactly the same as listeners), but I am not able to find anything for UDP listening ports in WMI (basically the equivalent of `netstat -p UDP -a -b` )

We could try and parse the results of the netstat command, but it would be better if we were able to access the [IP Helper API](https://msdn.microsoft.com/en-us/library/aa366073.aspx) (aka IPHLPAPI) directly in Go and call the [GetExtendedUdpTable](https://msdn.microsoft.com/en-us/library/aa366033.aspx)/[GetExtendedUdpTable](https://msdn.microsoft.com/en-us/library/aa365930.aspx) methods or [GetUdpStatistics](http://www.geekpage.jp/en/programming/iphlpapi/getudpstatistics.php) (which has dwNumAddrs for "The number of entries in the UDP listener table.") Kyle was concerned about the size of the table being to large on some hosts (netstat can't run on our load balancers due to volume of connections), so in either case we would want a fast fail if the check was going to take too long.

Someone wrote a [C# wrapper for IP Helper API](http://www.timvw.be/2007/09/09/build-your-own-netstatexe-with-c/) and there is source code available for [Netstatp.exe](https://web.archive.org/web/20050212024125/http://www.sysinternals.com/files/netstatp.zip) (an old tool from Sysinternals that mimics netstat) but this is above my abilities.

I will be adding the TCP and UDP WMI metrics, but long term we need to monitor the available listening ports and alert when they are out of the expected range.

PR #777, opened 2015-03-10
cmd/scollector: add win.net.tcp.* metrics

gbrayut:


gbrayut:
Easy win for tracking TCP ports/segments. UDP requires using IP Helper API see #724

PR #779, opened 2015-03-11
cmd/scollector: fix issues with c_network_team_windows and NETCLRMemory

gbrayut:
I found some scollector errors in the Windows log files using Kibana:

The c.Enable variable in c_network_team_windows init function was not being updated correctly, probably due to a closure issue. This caused the collector to become active on or-dc02 and ny-dc05 (and others) and produce errors. Renaming the IntervalCollector variable (so it is not shared/reused) appears to solve the closure issue and stop the errors.

Also the PercentTimeinGC values were overflowing uint32, even though that is what WMI says they are. See http://goo.gl/x6om5l for example of error.

These changes are already running in a test build on a handful of systems pointed at prod bosun.

### DIFF
--- a/cmd/scollector/collectors/network_windows.go
+++ b/cmd/scollector/collectors/network_windows.go
@@ -16,24 +16,25 @@ import (
 func init() {
 	collectors = append(collectors, &IntervalCollector{F: c_network_windows, init: winNetworkInit})
 
-	c := &IntervalCollector{
+	c_winnetteam := &IntervalCollector{
 		F: c_network_team_windows,
 	}
 	// Make sure MSFT_NetImPlatAdapter and MSFT_NetAdapterStatisticsSettingData
 	// are valid WMI classes when initializing c_network_team_windows
-	c.init = func() {
+	c_winnetteam.init = func() {
 		var dstTeamNic []MSFT_NetLbfoTeamNic
 		var dstStats []MSFT_NetAdapterStatisticsSettingData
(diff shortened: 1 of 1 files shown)


gbrayut:
Hmm... just had the same error that I thought was PercentTimeinGC  happen on one of the servers with the uint64 code. Full details at http://goo.gl/sGdwC0

interval.go:58: bosun.org/cmd/scollector/collectors.c_dotnet_memory: strconv.ParseInt: parsing "18446744073709551448": value out of range'

Must be happening somewhere in the ole/wmi packages. I'll try and figure out where that.

gbrayut:
Interesting:

18446744073709551448 value from error message
18446744073709551615 uint64 max value

It may be that the call to `strconv.ParseInt(val, 10, 64)` in [StackExchange\wmi\wmi.go](https://github.com/StackExchange/wmi/blob/master/wmi.go#L292) is where the error is coming from. I'm going to try changing that to strconv.ParseUint and see if that stops the errors.

gbrayut:
Detecting [out of range errors](http://stackoverflow.com/questions/28977718) was more difficult than I though, but after asking on Stack Overflow I got the answer I was looking for. Testing the above changes to the WMI class on a few systems to see if it fixes the c_dotnet_memory parsing errors.

gbrayut:
The ParseInt errors are now fixed via https://github.com/StackExchange/wmi/pull/9 which is much cleaner.

I'll rebase this on master for the other fix and once both are approved we can party the wmi fix in

PR #781, opened 2015-03-11
cmd/bosun: Enable IsEmail in template subjects

mjibson:


mjibson:
Body and Subject are strings because they are sent over JSON to the dashboard. The others are []byte because they are written as an io.Reader.

wkshare:
@mjibson How to use？

gbrayut:
@wkshare here is an example template where the subject is rendered in markdown if it is being sent to our chat room via http post or as normal text if it sent via email.

template exceptions {
    subject = `
        {{ if .IsEmail }}
            {{.Last.Status}}: {{.Eval .Alert.Vars.exceptions_total | printf "%.f"}} Exceptions
        {{else}}
            **[{{.Eval .Alert.Vars.exceptions_total | printf "%.f"}} Exceptions](https://example.com/exceptions)**
        {{end}}`

wkshare:
thks, @gbrayut

PR #791, opened 2015-03-16
cmd/bosun: log to syslog

### LINKED ISSUE #768: Monitor to syslog when on Linux
Just like we do with scollector, we should update Bosun to be sending its logs to syslog

mjibson:
fixes #768


captncraig:
:thumbsup:

PR #795, opened 2015-03-17
cmd/scollector: win.service.* metrics

gbrayut:
Also cleanup default process and process_dotnet settings.
This is currently running as a second instance on ny-web01 pointed at devbosun

### DIFF
--- a/cmd/scollector/collectors/processes_windows.go
+++ b/cmd/scollector/collectors/processes_windows.go
@@ -23,8 +23,8 @@ func AddProcessConfig(line string) error {
 
 func WatchProcesses() {
 	if len(regexesProcesses) == 0 {
-		// if no processes configured in config file, use this set instead.
-		regexesProcesses = append(regexesProcesses, regexp.MustCompile("chrome|powershell|scollector|SocketServer|WinRM|MSSQLSERVER|StackServerProd|StackServerDev|LogStasher"))
+		// if no process settings configured in config file, use this set instead.
+		regexesProcesses = append(regexesProcesses, regexp.MustCompile("chrome|powershell|scollector|WinRM|MSSQLSERVER"))
 	}
 	collectors = append(collectors, &IntervalCollector{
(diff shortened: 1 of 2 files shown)


cmd/scollector/collectors/processes_windows.go
```diff
@@ -40,7 +40,7 @@ func c_windows_processes() (opentsdb.MultiDataPoint, error) {
 	}
 
 	var svc_dst []Win32_Service
-	var svc_q = wmi.CreateQuery(&svc_dst, `WHERE Started=true`)
+	var svc_q = wmi.CreateQuery(&svc_dst, "")
```
alienth: I'm not sure what is going on here. Would you care to drop the sick knowledge beats on me?

kylebrandt: @alienth It reflects the Object (svc_dst) and creates a WMI sql like query. You can specify a where filter as a string.

kylebrandt: See https://msdn.microsoft.com/en-us/library/aa394606%28v=vs.85%29.aspx for more about WQL.

kylebrandt: @alienth Oh also I use https://wmie.codeplex.com/ to explore these. Then we have some linqpad C# which generates the Go structs from C#. Somewhere in the repo

gbrayut: https://github.com/bosun-monitor/bosun/blob/master/util/Windows_WMI_Metadata.linq is a linqpad script that can do some simple code generation for us. Try running it from ny-rdp01.

Basically we use the github.com/stackexchange/wmi library to use reflection to create a WQL query based on a struct.

kylebrandt: https://github.com/StackExchange/wmi/blob/master/wmi.go#L393

PR #801, opened 2015-03-18
cmd/scollector: new collector - riak

elementai:
Took inspiration from memcached collector. Metrics are named similarly to tcollector's riak.py . Tested on docker's guillermo/riak which is 1.4.2 version.

### DIFF
--- a/cmd/scollector/collectors/riak.go
+++ b/cmd/scollector/collectors/riak.go
@@ -1,0 +1,506 @@
+package collectors
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strings"
+
+	"bosun.org/metadata"
+	"bosun.org/opentsdb"
+)
+
+var riakMeta = map[string]MetricMeta{
+	"pbc_connects_total": {
+		Metric:   "pbc_connections",
+		RateType: metadata.Counter,
+		Unit:     metadata.Connection,
+		Desc:     "Total number of Protocol Buffers connections made.",
+	},
+	"read_repairs_total": {
+		Metric:   "read_repairs",
+		RateType: metadata.Counter,
+		Unit:     metadata.Operation,
+		Desc:     "Total number of Read Repairs this node has coordinated.",
+	},
+	"read_repairs_primary_outofdate_count": {
+		Metric:   "read_repairs_primary_outofdate",
+		RateType: metadata.Counter,
+		Unit:     metadata.Operation,
+		Desc:     "Total number of read repair operations performed on primary vnodes due to stale replicas.",
+	},
+	"read_repairs_primary_notfound_count": {
+		Metric:   "read_repairs_primary_notfound",
+		RateType: metadata.Counter,
+		Unit:     metadata.Operation,
+		Desc:     "Total number of read repair operations performed on primary vnodes due to missing replicas.",
+	},
+	"read_repairs_fallback_outofdate_count": {
+		Metric:   "read_repairs_fallback_outofdate",
+		RateType: metadata.Counter,
+		Unit:     metadata.Operation,
+		Desc:     "Total number of read repair operations performed on fallback vnodes due to stale replicas.",
+	},
+	"read_repairs_fallback_notfound_count": {
+		Metric:   "read_repairs_fallback_notfound",
+		RateType: metadata.Counter,
+		Unit:     metadata.Operation,
+		Desc:     "Total number of read repair operations performed on fallback vnodes due to missing replicas.",
+	},
+	"coord_redirs_total": {
+		Metric:   "coord_redirs",
+		RateType: metadata.Counter,
+		Unit:     metadata.Operation,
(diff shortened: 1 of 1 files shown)


cmd/scollector/collectors/riak.go
```diff
@@ -0,0 +1,504 @@
...
+		if m, ok := riakMeta[k]; ok {
+			if strings.HasPrefix(m.Metric, "node.latency") {
+				v = v.(float64) / 1000000
+			}
+			Add(&md, "riak."+m.Metric, v, m.TagSet, m.RateType, m.Unit, m.Desc)
+		} else if k == "ring_members" {
+			nodes, ok := v.([]interface{})
+	                if !ok || len(nodes) < 1 {
+				slog.Error("unexpected content or type for 'ring_members' metric array")
+				return nil, nil
```
elementai: Shall I create new `error` and return it ?

kylebrandt: @elemental: Yes, instead of calling slog create an error with `fmt.Errorf` and return that - it will get logged higher up the stack.

cmd/scollector/collectors/riak.go
```diff
@@ -0,0 +1,504 @@
...
+			}
+			Add(&md, "riak.connected_nodes", len(nodes), nil, metadata.Gauge, metadata.Count, descConNodes)
+	}
+	}
+	return md, nil
+}
+
+
+const (
+        descConNodes = "The number of connected nodes, calculated as size of ring_members array"
```
kylebrandt: desc should always end in periods

cmd/scollector/collectors/riak.go
```diff
@@ -0,0 +1,504 @@
...
+	defer res.Body.Close()
+	var r map[string]interface{}
+	j := json.NewDecoder(res.Body)
+	if err := j.Decode(&r); err != nil {
+		return nil, err
+	}
+	for k, v := range r {
+		if m, ok := riakMeta[k]; ok {
+			if strings.HasPrefix(m.Metric, "node.latency") {
+				v = v.(float64) / 1000000
```
kylebrandt: The type assertion should be checked here, for example:

nl, ok := v.(float64); ok {
    v = nl / 10000
}

If you don't check and there is a failure in the assertion you will get a panic.

cmd/scollector/collectors/riak.go
```diff
@@ -0,0 +1,504 @@
...
+		Metric:   "pbc_connections",
+		RateType: metadata.Counter,
+		Unit:     metadata.Connection,
+		Desc:     "Total number of Protocol Buffers connections made",
+	},
+	"read_repairs_total": {
+		Metric:   "read_repairs",
+		RateType: metadata.Counter,
+		Unit:     metadata.Operation,
+		Desc: "Total number of Read Repairs this node has coordinated",
```
mjibson: Needs a `gofmt`

cmd/scollector/collectors/riak.go
```diff
@@ -0,0 +1,504 @@
...
+	"bosun.org/opentsdb"
+	"bosun.org/slog"
+)
+
+var riakMeta = map[string]MetricMeta{
+	"pbc_connects_total": {
+		Metric:   "pbc_connections",
+		RateType: metadata.Counter,
+		Unit:     metadata.Connection,
+		Desc:     "Total number of Protocol Buffers connections made",
```
mjibson: Period at end of sentence for all descriptions.

cmd/scollector/collectors/riak.go
```diff
@@ -0,0 +1,504 @@
...
+
+func c_riak() (opentsdb.MultiDataPoint, error) {
+	var md opentsdb.MultiDataPoint
+	res, err := http.Get(riakURL)
+	if err != nil {
+		return nil, err
+	}
+	defer res.Body.Close()
+	var r map[string]interface{}
+	j := json.NewDecoder(res.Body)
```
mjibson: Combine this with line below: `if err := json.NewDecoder(res.Body.Decode(&r))...`

mjibson:
Squish these 2 commits down into 1, then push -f so this PR is just 1 commit, then we'll merge. Thanks for the contribution.

cmd/scollector/collectors/riak.go
```diff
@@ -0,0 +1,506 @@
...
+		Desc:     "Total number of vnodes currently running in the Riak Search subsystem.",
+	},
+}
+
+func init() {
+	collectors = append(collectors, &IntervalCollector{F: c_riak, Enable: enableRiak})
+}
+
+const (
+	riakURL string = "http://localhost:8098/stats"
```
captncraig: Is this going to be constant for all riak installations, or should this be configurable?

elementai: That can be set in riak.config . I'm not yet sure how to better implement that.

kylebrandt: We assume localhost and default port in other collectors, so I think this
is fine for now

PR #803, opened 2015-03-18
cmd/scollector: fix for "The description for Event ID" in event logs

gbrayut:
See http://stackoverflow.com/questions/29130586
This will update the EventMessageFile to have the correct value when installing or running the service

### DIFF
--- a/cmd/scollector/service_windows.go
+++ b/cmd/scollector/service_windows.go
@@ -4,11 +4,13 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"syscall"
 	"time"
 
 	"bosun.org/_third_party/code.google.com/p/winsvc/debug"
 	"bosun.org/_third_party/code.google.com/p/winsvc/eventlog"
 	"bosun.org/_third_party/code.google.com/p/winsvc/mgr"
+	"bosun.org/_third_party/code.google.com/p/winsvc/registry"
 	"bosun.org/_third_party/code.google.com/p/winsvc/svc"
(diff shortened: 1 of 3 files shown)


cmd/scollector/service_windows.go
```diff
@@ -71,7 +73,9 @@ func installService(name, desc string) error {
 		return err
 	}
 	defer s.Close()
-	err = eventlog.InstallAsEventCreate(name, eventlog.Error|eventlog.Warning|eventlog.Info)
+	// Workaround for bug in eventlog.InstallAsEventCreate
```
mjibson: What exactly is the bug? Comparing this to the implementation of InstallAsEventCreate ([here](https://code.google.com/p/winsvc/source/browse/eventlog/install.go#61)), it is identical to your change except that this one calls ExpandEnv and the other one doesn't. Can you describe the bug more?

gbrayut:
Add patch from upstream https://code.google.com/p/winsvc/issues/detail?id=12

Also the registry key stuff could be swapped out with a call to reg.exe if we really don't want to do that in native go

cmd/scollector/service_windows.go
gbrayut: This workaround is no longer needed, as I included the upstream patch. We still need to fix any existing installs that are currently broken by manually expanding the %systemroot% variable in the REG_SZ key. Ping me when you have time to jump on a hangout if you want more details.

mjibson:
LGTM. Merge when you like.

PR #804, opened 2015-03-19
Add info commandstats metrics for redis

ptqa:
I've added metrics for "info commandstats" output. I'm new to Go, so not sure if this code is ok, but it works.

### DIFF
--- a/cmd/scollector/collectors/riak.go
+++ b/cmd/scollector/collectors/riak.go
@@ -1,0 +1,506 @@
+package collectors
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strings"
+
+	"bosun.org/metadata"
+	"bosun.org/opentsdb"
+)
+
+var riakMeta = map[string]MetricMeta{
+	"pbc_connects_total": {
+		Metric:   "pbc_connections",
+		RateType: metadata.Counter,
+		Unit:     metadata.Connection,
+		Desc:     "Total number of Protocol Buffers connections made.",
+	},
+	"read_repairs_total": {
+		Metric:   "read_repairs",
+		RateType: metadata.Counter,
+		Unit:     metadata.Operation,
+		Desc:     "Total number of Read Repairs this node has coordinated.",
+	},
+	"read_repairs_primary_outofdate_count": {
+		Metric:   "read_repairs_primary_outofdate",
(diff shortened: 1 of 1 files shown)


cmd/scollector/collectors/redis_unix.go
```diff
@@ -221,8 +233,9 @@ func c_redis() (opentsdb.MultiDataPoint, error) {
 				continue
 			}
 			sp := strings.Split(line, ":")
+			match, _ := regexp.MatchString("cmdstat_([a-z]+)",sp[0])
 			if len(sp) < 2 || !redisFields[sp[0]] {
-				continue
+				if !match {continue}
```
captncraig: why not move the `if match` block above this to remove the nested if stmt.

captncraig: This file also needs to be run through `go fmt`

cmd/scollector/collectors/redis_unix.go
```diff
@@ -236,6 +249,16 @@ func c_redis() (opentsdb.MultiDataPoint, error) {
 				Add(&md, "redis."+sp[0], status(sp[1]), tags, metadata.Unknown, metadata.None, "")
 				continue
 			}
+			if match {
+				cmdstats := strings.Split(sp[1],",")
+				cmdstats_calls := strings.Split(cmdstats[0],"=")
```
captncraig: We need bounds checking any time we start indexing into split strings.

cmd/scollector/collectors/redis_unix.go
```diff
@@ -182,13 +180,22 @@ func c_redis() (opentsdb.MultiDataPoint, error) {
...
+		}
+		info, info_err := c.Do("info")
+		if info_err != nil {
+			Error = info_err
 			continue
 		}
+
 		tags := instance.Copy()
-		infoSplit := strings.Split(string(lines.([]uint8)), "\n")
+		combined_lines := string(info.([]uint8)) + string(info_command.([]uint8))
```
captncraig: Why combine at all? Why not iterate over info_command separately? It seems like a lot of the conditional logic in later stages would be simpler by not combining.

ptqa:
If I leave a comment may be it's possible to reopen...

PR #813, opened 2015-03-20
Rule page dependencies

captncraig:
Allows rule page to handle more complicated scenarios with alerts that depend on other alerts. 

Will show results from the **_last_** alert specified.

Much of this code is to enable tracing alert dependencies to know what we need to show to load any given alert on the rule page.


kylebrandt:
Check comment style throughout

cmd/bosun/web/expr.go
```diff
@@ -279,11 +279,8 @@ func Rule(t miniprofiler.Timer, w http.ResponseWriter, r *http.Request) (interfa
 	if err != nil {
 		return nil, err
 	}
-	if len(c.Alerts) != 1 {
-		return nil, fmt.Errorf("exactly one alert must be defined")
-	}
 	var a *conf.Alert
-	// Set a to the first alert.
+	// Set a to the last alert.
```
mjibson: This for loop doesn't guarantee that because it's a map.

cmd/bosun/sched/check.go
```diff
@@ -293,7 +293,7 @@ func filterDependencyResults(results *expr.Results) expr.ResultSlice {
 }
 
 func markDependenciesUnevaluated(events map[expr.AlertKey]*Event, deps expr.ResultSlice, alert string) (unevalCount, unknownCount int) {
-	unevalCount, unknownCount := 0, 0
+	unevalCount, unknownCount = 0, 0
```
mjibson: This line is now unneeded, they are already 0.

cmd/bosun/conf/conf.go
```diff
@@ -865,20 +871,110 @@ func (c *Conf) loadAlert(s *parse.SectionNode) {
...
+				deps = append(deps, nextDep)
+			}
+		}
+		toCheck = newSet
+	}
+	text := ""
+	printed := map[*ConfItem]bool{}
+	//print dependencies right to left.
+	for i := len(deps) - 1; i >= 0; i-- {
+		fmt.Println(i, len(deps))
```
mjibson: remove debug

cmd/bosun/conf/conf.go
```diff
@@ -865,20 +871,110 @@ func (c *Conf) loadAlert(s *parse.SectionNode) {
...
+	c.findAllDependencies(&a)
 	c.Alerts[name] = &a
 	c.OrderedAlerts = append(c.OrderedAlerts, &a)
 }
 
+func (c *Conf) findAllDependencies(a *Alert) {
+	//Notifications and lookups are dependencies
+	var walkNotifications = func(n *Notifications) {
+		for _, l := range n.Lookups {
+			a.Dependencies = append(a.Dependencies, &l.ConfItem)
```
mjibson: Instead of being forced to do stuff like this (where a function dealing with Alerts has to know about looks and chains in notifications), what if we had a dependency interface, that each type (alert, lookup, notification, macro, template) implemented?

cmd/bosun/conf/conf.go
```diff
@@ -879,20 +885,110 @@ func (c *Conf) loadAlert(s *parse.SectionNode) {
...
+		walkExpr(a.Depends.Tree.Root)
+	}
+
+}
+
+func (c *ConfItem) textWithDependencies() string {
+	deps := []*ConfItem{}
+	// Recursively walk the tree, appending all other dependencies.
+	toCheck := []*ConfItem{}
+	for _, d := range c.Dependencies {
```
mjibson: `copy` will do this

cmd/bosun/conf/conf.go
```diff
@@ -879,20 +885,110 @@ func (c *Conf) loadAlert(s *parse.SectionNode) {
...
+		deps = append(deps, d)
+		toCheck = append(toCheck, d)
+	}
+	for len(toCheck) > 0 {
+		newSet := []*ConfItem{}
+		for _, dep := range toCheck {
+			for _, nextDep := range dep.Dependencies {
+				newSet = append(newSet, nextDep)
```
mjibson: copy

cmd/bosun/conf/dependencies_test.go
```diff
@@ -0,0 +1,61 @@
+package conf
+
+import (
+	"regexp"
+	"strings"
+	"testing"
+)
+
+var dependencyTestConfig = `
```
mjibson: const

mjibson:
Remove the changes to static.go and the .js files since they are unrelated to this change. (However they should be committed in their own, other commit.)

PR #814, opened 2015-03-20
cmd/bosun: Add blacklist for put IPs

kylebrandt:


cmd/bosun/conf/conf.go
```diff
@@ -69,21 +70,29 @@ type Conf struct {
 	squelch         []string
 }
 
-type AuthorizedIPs []*net.IPNet
-
-func (authorizedIPs AuthorizedIPs) Authorized(ip net.IP) (authorized bool) {
-	if len(authorizedIPs) == 0 {
+func (c *Conf) PutAuthorized(ip net.IP) bool {
+	// First process all blocked put ip ranges.
+	if len(c.BlockedPutIPs) > 0 {
```
captncraig: why is this if required? If it is empty, than the for loop will just do nothing.

cmd/bosun/conf/conf.go
```diff
@@ -432,14 +441,23 @@ func (c *Conf) loadGlobal(p *parse.PairNode) {
 		c.PingDuration = d
 	case "noSleep":
 		c.NoSleep = true
-	case "putIPs":
+	case "BlockedPutIPs":
```
captncraig: I dislike the duplication in these two cases. Can we share a bit more of this logic please?

cmd/bosun/conf/conf.go
kylebrandt: Because the default behavior is block which I want to keep in place in case this is ever extended. len of zero means (whitelist feature is not being used), so just allow if it isn't blocked

captncraig: Then should it read `if len(c.AllowedPutIPs) > 0 ` ?

The if as it is literally does nothing.

cmd/bosun/conf/conf.go
```diff
@@ -432,15 +439,11 @@ func (c *Conf) loadGlobal(p *parse.PairNode) {
...
-		for _, rc := range rawCIDRs {
-			_, ipnet, err := net.ParseCIDR(rc)
-			if err != nil {
-				c.error(err)
-			}
-			c.PutIPs = append(c.PutIPs, ipnet)
-		}
+	case "blockedPutIPs":
+		c.BlockedPutIPs = c.parseIPs(v, c.BlockedPutIPs)
+		fmt.Println(c.BlockedPutIPs)
```
mjibson: remove debug print

cmd/bosun/conf/conf.go
```diff
@@ -511,6 +514,18 @@ func (c *Conf) loadSection(s *parse.SectionNode) {
 	}
 }
 
+func (c *Conf) parseIPs(s string, nets []*net.IPNet) []*net.IPNet {
```
mjibson: This function should either take nets or return nets, not both.

Issue #806, opened 2015-03-19
Expression working only on historical data


vitorboschi:
I'm doing some tests on Bosun, but I can't make the following expression to work:

q("max:1m-max:ping.timeout{dst_host=reader*}", "1h", "")

The generated query is correct:
end=2015/03/19-17:33:11&m=max:1m-max:ping.timeout{dst_host=*}&start=2015/03/19-16:33:11

If I click the query link, the results are displayed correctly (and also if I specify a past date on the expression page). It just doesn't show any data on the expression page if I don't set any date/time

Is it a problem if most of the data currently in openTSDB was not sent through Bosun, but directly to openTSDB?

I'm on rev 236e300ee8a5fdd601e7ffc1768f47adff2d80d8

kylebrandt:
It maters if you use a wildcard to in the tag value.  OpenTSDB only
supports an asterisk as the only character (so it is all or nothing).
Expanding the glob is an extension bosun does, and it learns about possible
expansions based on the data relayed through it.

vitorboschi:
I don't think the problem is in the expansion, since it already seens to know possible tag values, because I can execute queries using * on it. Is there anything I can test, or some way to enable more debug info?

kylebrandt:
Hit alt-p tp bing up mini profiler, then query. There should be stuff in
the lower left and I believe it will show the opentsdb query issued.

vitorboschi:
I couldn't find the openTSDB query in the mini profiler, just the expression inside Bosun itself. The query generated seens correct, because it's shown by bosun as a link, and it indeed works as expected if I click it.
For some reason, it's just not querying openTSDB, or not processing its response

vitorboschi:
Profiler screenshot: 
![bosun](https://cloud.githubusercontent.com/assets/71193/6738106/93f0a0c8-ce4e-11e4-9cda-cb4df8efa63e.png)

kylebrandt:
Now that you have only a  `*` it should work. My next guess would be that the time on the Bosun host doesn't match the time of OpenTSDB.

vitorboschi:
Time shouldn't be a problem, since both Bosun and OpenTSDB are on the same machine. I'm trying to add some debug info in the code. 
I can see that func Query in expr/funcs.goexpr/funcs.go is executing without errors, So I'm trying to go add some debug to the function which queries OpenTSDB. I've tried

func (r *Request) Query(host string) (ResponseSet, error) { 

[...]

PR #815, opened 2015-03-20
use unix timestamps on tsdb request

vitorboschi:
this avoids issues when OpenTSDB is not running on UTC timezone. 
Fixes issue #806

### DIFF
--- a/cmd/bosun/sched/sched_test.go
+++ b/cmd/bosun/sched/sched_test.go
@@ -112,7 +112,7 @@ func testSched(t *testing.T, st *schedTest) (s *Schedule) {
 }
 
 var queryTime = time.Date(2000, 1, 1, 12, 0, 0, 0, time.UTC)
-var window5Min = `"2000/01/01-11:55:00", "2000/01/01-12:00:00"`
+var window5Min = `"9.467277e+08", "9.46728e+08"`
 
 func TestCrit(t *testing.T) {
 	testSched(t, &schedTest{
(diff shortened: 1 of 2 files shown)


kylebrandt:
@mjibson See any issue with querying this way? I don't....

mjibson:
Bosun does a lot dealing with time, so I'd like to test this a bit more before it goes in.

mjibson:
This diff breaks our tests (`go test ./...` in the root directory) so it can't go in yet. It could be because the tests need to be updated for this new format

mjibson:
Add https://github.com/bosun-monitor/bosun/commit/84c14cf0d0f0aaac3d640c55c441870b254288ec and we'll merge.

vitorboschi:
@mjibson I've just updated the PR. Thanks for pointing and fixing the test

cmd/bosun/sched/sched_test.go
```diff
@@ -112,7 +112,7 @@ func testSched(t *testing.T, st *schedTest) (s *Schedule) {
 }
 
 var queryTime = time.Date(2000, 1, 1, 12, 0, 0, 0, time.UTC)
-var window5Min = `"2000/01/01-11:55:00", "2000/01/01-12:00:00"`
+var window5Min = `"9.467277e+08", "9.46728e+08"`
```
captncraig: This makes the tests pretty hard to follow. Perhaps we can add a helper function to format dates properly, but still view them more sensibly?

mjibson: True. However, the tests were already doing this, except for the first query. At least they are consistently bad. We can make this improvement in another PR.

Issue #573, opened 2014-12-04
Ability to add metadata to external collectors


mjibson:
<a href="https://github.com/PeterGrace"><img src="https://avatars.githubusercontent.com/u/967365?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [PeterGrace](https://github.com/PeterGrace)**
_Tuesday Sep 16, 2014 at 13:15 GMT_
_Originally opened as https://github.com/bosun-monitor/scollector/issues/61_

----

I would really like to be able to specify metadata for some of my external checks.  Can some sort of interface be added which would allow me to set metadata for custom metrics?

gbrayut:
I've had this come up as an issue a few times in the past week so I wanted to bump it to give it some better visibility. Especially now that we show metadata on the graph page, external collectors should be able to submit metadata just like regular collectors.

kylebrandt:
done http://bosun.org/scollector/external-collectors

PR #817, opened 2015-03-22
cmd/scollector: add meta for extenal collectors

ptqa:
It's related to this [issue](https://github.com/bosun-monitor/bosun/issues/573). I've implemented metadata for external collectors, so now you can use it this way:
```metric timestamp value tag1=val1 tag2=val2 meta.Unit=unitname meta.Desc=description meta.Rate=ratename ... ```
There is a problem that description should be a single word, but it's hard to change this without changing all output format.

### DIFF
--- a/cmd/scollector/collectors/program.go
+++ b/cmd/scollector/collectors/program.go
@@ -2,6 +2,7 @@ package collectors
 
 import (
 	"bufio"
+	"errors"
 	"io"
 	"os"
 	"os/exec"
@@ -11,6 +12,7 @@ import (
 	"strings"
 	"time"
 
+	"bosun.org/metadata"
 	"bosun.org/opentsdb"
 	"bosun.org/slog"
 	"bosun.org/util"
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: cmd/scollector/collectors/program.go [hunk 1]
```diff
@@ -151,12 +178,20 @@ Loop:
 			Value:     val,
 			Tags:      opentsdb.TagSet{"host": util.Hostname},
 		}
-		for _, tag := range sp[3:] {
-			tags, err := opentsdb.ParseTags(tag)
+		for _, field := range sp[3:] {
+			if strings.HasPrefix(field, "meta.") {
```
mjibson: This is an interesting solution. @kylebrandt thoughts? The other kind of solution to consider is just a line completely devoted to metadata. This would allow the metadata to be printed, say, just once on collector startup instead of being forced to output with every or just the first point, which is annoying for external collector authors. It would also allow the meta namespace to be used as is.

gbrayut: I like the idea of a separate line for metadata. Works better with grep when debugging.

ptqa: No problem, should I implement smthng like
    metric meta.Unit=unitname meta.Desc=description meta.Rate=ratename?

mjibson: What about:

```
metadata unit seconds
metadata type rate
metadata desc The description of the metric.
```

This makes parsing quite easy.

ptqa: But there should be metric name somewhere, isn't it?

kylebrandt: I thought I commented on this already, sorry. My idea was to have a `|` after everything that exists already, and then you can do `Metakey1="value"  Metakey2="value"` on the same line. I thought about adding another line, but I thought that might be problematic because the order of lines being flushed needs to be ensured, not awesome if you have something with multiple threads.

mjibson: Why does the order of lines matter for metadata?

kylebrandt: @mjibson How do you identify which metric + tagset the metadata belongs to without repeating most of the information on the original line anyways?

mjibson: Right, I forgot about that. We should discuss over hangout.

mjibson:
We're working on another way to do this and will have some code up soon.

ptqa:
Ok then

mjibson:
PR at https://github.com/bosun-monitor/bosun/pull/843 if you want to comment on the design.

PR #819, opened 2015-03-22
util functions to read last line(s) from file

elementai:
Two utility functions to read last line or lines from a file. Adapted from https://gist.github.com/kylebrandt/b2906d990d0e762ad87e , I've used `bufio.Scanner` for cross-platform compatibility (so I hope). 

P.S. Had to include https://github.com/stretchr/testify for tests, for my changes scroll to the very end.
P.P.S. Sorry for the second PR, messed up 3rd party packages


elementai:
@mjibson In response to your commit in my previous PR, I need this function for Chef collector to read Chef's logfile which can be rather large.

elementai:
It could probably be better to implement ReadLastN -something function, but I dunno anyone really needs it.

mjibson:
The util package is only for functions that are needed by both scollector and bosun. If an individual collector needs additional functionality, it must be implemented in the collector's .go file, and party'd if external packages are used.

elementai:
Ok.

PR #826, opened 2015-03-23
cmd/scollector: Chef collector

elementai:
Simple collector for Chef, reads and parses last line from `/var/chef/reports/run_status_json.log`.

### DIFF
--- a/cmd/scollector/collectors/chef_linux.go
+++ b/cmd/scollector/collectors/chef_linux.go
@@ -1,0 +1,138 @@
+package collectors
+
+import (
+	"bufio"
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+	"strings"
+	"time"
+
+	"bosun.org/metadata"
+	"bosun.org/opentsdb"
+)
+
+type crSummary struct {
+	StartTime   string  `json:"start_time"`
+	EndTime     string  `json:"end_time"`
+	ElapsedTime float32 `json:"elapsed_time"`
+	Status      string  `json:"status"`
+}
+
+func init() {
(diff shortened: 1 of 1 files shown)


cmd/scollector/collectors/chef_linux.go
```diff
@@ -0,0 +1,139 @@
...
+	chefRunLog = "/var/chef/reports/run_status_json.log"
+)
+
+func chefEnable() bool {
+	_, err := os.Stat(chefRunLog)
+	return err == nil
+}
+
+func c_chef_linux() (opentsdb.MultiDataPoint, error) {
+	md, err := chef_linux(chefRunLog)
```
elementai: Additional function is for tests convenience. Will include tests later, need to discuss a bit.

cmd/scollector/collectors/chef_linux.go
```diff
@@ -0,0 +1,139 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+	"strings"
+	"time"
+
+	"bufio"
```
mjibson: This should be grouped with other stdlib imports.

cmd/scollector/collectors/chef_linux.go
```diff
@@ -0,0 +1,139 @@
...
+
+	"bosun.org/metadata"
+	"bosun.org/opentsdb"
+)
+
+type CRSummary struct {
```
mjibson: Make this unexported.

mjibson:
@kylebrandt ok adding this? LGTM.

PR #816, opened 2015-03-21
cmd/scollector: commandstats metrics for redis #2

ptqa:
This adds metrics for "info commandstats" output. I accidentally closed previous pull request (https://github.com/bosun-monitor/bosun/pull/804) and I'm unable to reopen it for some reason, so here is new one. I've made changes as @captncraig adviced.

PR #831, opened 2015-03-23
  cmd/scollector: commandstats metrics for redis #3

ptqa:
And again I forgot to commit before push. [Previous try] (https://github.com/bosun-monitor/bosun/pull/816)

Changed if logic as @mjibson advised.

### DIFF
--- a/cmd/scollector/collectors/redis_unix.go
+++ b/cmd/scollector/collectors/redis_unix.go
@@ -182,13 +182,13 @@ func c_redis() (opentsdb.MultiDataPoint, error) {
 			continue
 		}
 		defer c.Close()
-		lines, err := c.Do("INFO")
+		info, err := c.Do("info", "all")
 		if err != nil {
 			Error = err
 			continue
 		}
 		tags := instance.Copy()
-		infoSplit := strings.Split(string(lines.([]uint8)), "\n")
+		infoSplit := strings.Split(string(info.([]uint8)), "\n")
 		for _, line := range infoSplit {
 			line = strings.TrimSpace(line)
 			sp := strings.Split(line, ":")
@@ -221,7 +221,7 @@ func c_redis() (opentsdb.MultiDataPoint, error) {
 				continue
 			}
 			sp := strings.Split(line, ":")
-			if len(sp) < 2 || !redisFields[sp[0]] {
+			if len(sp) < 2 || !(redisFields[sp[0]] || strings.HasPrefix(sp[0], "cmdstat_")) {
 				continue
 			}
 			if sp[0] == "master_link_status" {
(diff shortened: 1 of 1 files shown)


mjibson:
FYI you can `git push -f` to force push after modify your commits, instead of opening a new PR every change.

ptqa:
I know and I did it before, but if you `git reset --soft` and then `git push -f` pull request is closed automatically. :disappointed:

cmd/scollector/collectors/redis_unix.go
```diff
@@ -239,10 +239,53 @@ func c_redis() (opentsdb.MultiDataPoint, error) {
 			Add(&md, "redis."+sp[0], sp[1], tags, metadata.Unknown, metadata.None, "")
 		}
 		Add(&md, "redis.key_count", keys, tags, metadata.Gauge, metadata.Key, descRedisKeyCount)
+
+		info_command, info_command_err := c.Do("info", "commandstats")
```
kylebrandt: Since the default info gets almost all the information anyways, I think we should only call info once with the all argument, and parse that.

ptqa: @kylebrandt Ok, I did that in new commit.

cmd/scollector/collectors/redis_unix.go
```diff
@@ -236,6 +236,33 @@ func c_redis() (opentsdb.MultiDataPoint, error) {
...
 			}
+			if strings.HasPrefix(sp[0], "cmdstat_") {
+				cmdStats := strings.Split(sp[1], ",")
+				if len(cmdStats) < 3 {
+					continue
+				}
+				cmdStatsCalls := strings.Split(cmdStats[0], "=")
+				cmdStatsUsec := strings.Split(cmdStats[1], "=")
+				var cmdStatsMsec, cmdStatsMsecPc float64
+				microsec, err := strconv.ParseFloat(cmdStatsUsec[1], 64)
```
kylebrandt: Need to check length here to avoid possible panic, don't know that cmdStatusUsec is of len > 1

cmd/scollector/collectors/redis_unix.go
```diff
@@ -236,6 +236,33 @@ func c_redis() (opentsdb.MultiDataPoint, error) {
...
+				if err != nil {
+					continue
+				}
+				cmdStatsMsec = microsec / 1000
+				cmdStatsUsecPc := strings.Split(cmdStats[2], "=")
+				microsec, err = strconv.ParseFloat(cmdStatsUsecPc[1], 64)
```
kylebrandt: Check len here as well

cmd/scollector/collectors/redis_unix.go
```diff
@@ -236,6 +236,33 @@ func c_redis() (opentsdb.MultiDataPoint, error) {
...
+				if err != nil {
+					continue
+				}
+				cmdStatsMsecPc = microsec / 1000
+				if shortTag := strings.Split(sp[0], "_"); len(shortTag) == 2 {
+					tags["cmd"] = shortTag[1]
```
kylebrandt: len here as well

cmd/scollector/collectors/redis_unix.go
```diff
@@ -236,6 +236,33 @@ func c_redis() (opentsdb.MultiDataPoint, error) {
...
+				}
+				Add(&md, "redis.cmdstats_msec_pc", cmdStatsMsecPc, tags, metadata.Gauge, metadata.MilliSecond, descRedisCmdMsecPc)
+				Add(&md, "redis.cmdstats_msec", cmdStatsMsec, tags, metadata.Counter, metadata.MilliSecond, descRedisCmdMsec)
+				Add(&md, "redis.cmdstats_calls", cmdStatsCalls[1], tags, metadata.Counter, metadata.Operation, descRedisCmdCalls)
```
kylebrandt: len here as well

ptqa:
@kylebrandt So every time I split there should be if. Ok, done.

kylebrandt:
Split will always have at least one element, so need to check when using an
index > 0

PR #840, opened 2015-03-25
cmd/bosun: Refactor logstash elastic provider

kylebrandt:


cmd/bosun/conf/conf.go
```diff
@@ -111,6 +112,14 @@ func (c *Conf) GraphiteContext() graphite.Context {
 	return graphite.Host(c.GraphiteHost)
 }
 
+// LogstashContext returns a Logstash context
+func (c *Conf) LogstashContext() logstash.Context {
+	if len(c.LogstashElasticHosts) < 1 {
+		return nil
```
captncraig: Are we checking for nil everywhere we create one of these? Perhaps it would be better to return a no-op instance rather than nil?

cmd/bosun/expr/funcs.go
```diff
@@ -691,11 +692,16 @@ func timeTSDBRequest(e *State, T miniprofiler.Timer, req *opentsdb.Request) (s o
...
+func timeLSRequest(e *State, T miniprofiler.Timer, req *logstash.Request) (resp *elastic.SearchResult, err error) {
+	e.logstashQueries = append(e.logstashQueries, *req.Source)
+	b, _ := json.MarshalIndent(req.Source.Source(), "", "  ")
 	T.StepCustomTiming("logstash", "query", string(b), func() {
-		resp, err = service.SearchSource(source).Do()
+		getFn := func() (interface{}, error) {
+			return e.logstashContext.Query(req)
+		}
+		var val interface{}
+		val, err = e.cache.Get(string(b), getFn)
```
captncraig: Are we going to have caching issues here if we don't copy the results? Will downstream functions modify these results?

cmd/bosun/conf/conf.go
kylebrandt: Not sure I just modeled it after the existing ones.

cmd/bosun/expr/funcs.go
kylebrandt: @Dieterbe Do you know, I don't have a great grasp on how the caching works..... ?

Dieterbe: cache stores pointer to data. but the whole expression package was made with the decision that changing data is ok, so when avg() modifies the result it's working on, we don't want that to affect the data in cache, so we must copy the data as we pull it out of cache.  i'm not sure where we agreed that should be done, i think the caller of the cache (the query functions in funcs.go) have to take care of this, but @captncraig probably remembers this better?

kylebrandt: Either way I don't think it matters. I'm not caching the results that get returned from the query, rather I'm just caching the elastic response for identical queries and then processing it

Dieterbe: what is the response you're caching, if not the results from the query?
and what does "processing" mean?

kylebrandt: Elastic Response -> MakeIntoBosunResult -> BosunResult

I'm caching the elastic response, not the BosunResult. By processing I mean
transforming an elastic response to a Bosun Result.

cmd/bosun/conf/conf.go
```diff
@@ -111,6 +111,16 @@ func (c *Conf) GraphiteContext() graphite.Context {
 	return graphite.Host(c.GraphiteHost)
 }
 
+// LogstashContext returns a Logstash service
+func (c *Conf) Logstash() expr.LogstashElasticHosts {
```
mjibson: Remove this function. Make c.LogstashElasticHosts of type expr.LogstashElasticHosts instead of []string.

cmd/bosun/expr/logstash.go
```diff
@@ -0,0 +1,394 @@
...
+	},
+}
+
+// This is an array of Logstash hosts and exists as a type for something to attach
+// methods to.  The elasticsearch library will use the listed to hosts to discover all
+// of the hosts in the config
+type LogstashElasticHosts []string
+
+// InitClient sets up the elastic client. If the client has already been
+// initalized it is a noop
```
mjibson: . at eol

mjibson:
Do we also need to party the new elastic package with this?

kylebrandt:
No

PR #842, opened 2015-03-25
Don't auto-close errors

captncraig:


cmd/bosun/sched/check.go
```diff
@@ -287,8 +287,10 @@ func removeUnknownEvents(evs map[expr.AlertKey]*Event, alert string) {
 
 func (s *Schedule) closeErrorEvent(evs map[expr.AlertKey]*Event, alert string) {
 	ak := expr.NewAlertKey(alert, nil)
-	if s.Status(ak).Status() == StError && evs[ak] == nil { //if run history already has event for this key do nothing
-		evs[ak] = &Event{Status: StNormal}
+	if st, ok := s.status[ak]; ok && st.Status() == StError {
+		if evs[ak] == nil { //if run history already has event for this key do nothing
+			st.Forgotten = true //supress unkowns on this key
```
mjibson: You cannot mutate the status field here, you can only do it in RunHistory.

mjibson:
LGTM

PR #847, opened 2015-03-26
cmd/bosun: add goo.gl API key support

mjibson:


cmd/bosun/web/web.go
```diff
@@ -227,6 +228,33 @@ func JSON(h func(miniprofiler.Timer, http.ResponseWriter, *http.Request) (interf
...
 }
 
+func Shorten(w http.ResponseWriter, r *http.Request) {
+	u := url.URL{
+		Scheme: "https",
+		Host:   "www.googleapis.com",
+		Path:   "/urlshortener/v1/url",
+	}
+	if schedule.Conf.ShortURLKey != "" {
+		u.RawQuery = "key=" + schedule.Conf.ShortURLKey
```
kylebrandt: Should this be encoded?

mjibson: Nope. The docs say: "The API key is safe for embedding in URLs; it doesn't need any encoding."

PR #851, opened 2015-03-27
cmd/bosun: Instrument alerts with collect

kylebrandt:


cmd/bosun/sched/check.go
```diff
@@ -185,6 +186,52 @@ func (s *Schedule) RunHistory(r *RunHistory) {
...
+	}
+	metadata.AddMeta("bosun.alerts.current_severity", nil, "rate", metadata.Gauge, false)
+	metadata.AddMeta("bosun.alerts.current_severity", nil, "unit", metadata.Alert, false)
+	metadata.AddMeta("bosun.alerts.current_severity", nil, "desc", "The number of open alerts by current severity.", false)
+	for k, v := range lastAbnormalCount {
+		collect.Put("alerts.last_abnormal_severity", opentsdb.TagSet{"severity": k}, v)
+	}
+	metadata.AddMeta("bosun.alerts.bosun.alerts.last_abnormal_severity", nil, "rate", metadata.Gauge, false)
+	metadata.AddMeta("bosun.alerts.bosun.alerts.last_abnormal_severity", nil, "unit", metadata.Alert, false)
+	metadata.AddMeta("bosun.alerts.bosun.alerts.last_abnormal_severity", nil, "desc", "A number of open alerts by last abnormal severity.", false)
```
alienth: Full description should be in a const, no?

kylebrandt: First time we ever had done metadata in Bosun, so I don't know. Adding a
new function anyways

kylebrandt: Actually in this case I don't think we should use constants since the description documents the particular instrumentation in the code (so proximity wins over neatness). That is different than the way scollector uses metadata.

cmd/bosun/sched/check.go
```diff
@@ -185,6 +186,44 @@ func (s *Schedule) RunHistory(r *RunHistory) {
 	if checkNotify && s.nc != nil {
 		s.nc <- true
 	}
+	// Instrument alert states.
+	severityCount := make(map[string]int64)
+	lastAbnormalCount := make(map[string]int64)
+	unAckedCount := map[bool]int64{true: 0, false: 0}
+	activeCount := map[bool]int64{true: 0, false: 0}
```
captncraig: Using a map for these seems a bit confusing. Why not just have two variables activeCount/inactiveCount? / ackedCount / unAckedCount?

cmd/bosun/sched/check.go
```diff
@@ -185,6 +186,44 @@ func (s *Schedule) RunHistory(r *RunHistory) {
 	if checkNotify && s.nc != nil {
 		s.nc <- true
 	}
+	// Instrument alert states.
```
captncraig: Can we factor all of this into a seperate method? RunHistory is already too big IMO. `s.CollectMetrics()` or something may be preferable to inlining it here.

cmd/bosun/sched/check.go
```diff
@@ -185,6 +186,44 @@ func (s *Schedule) RunHistory(r *RunHistory) {
...
+	severityCount := make(map[string]int64)
```
captncraig: perhaps `countBySeverity` is a better name for this?

cmd/bosun/sched/check.go
```diff
@@ -185,6 +186,44 @@ func (s *Schedule) RunHistory(r *RunHistory) {
...
+		activeCount[state.IsActive()]++
+	}
+	for k, v := range severityCount {
+		collect.Put("alerts.current_severity", opentsdb.TagSet{"severity": k}, v)
+	}
+	metadata.AddMetricMeta("bosun.alerts.current_severity", metadata.Gauge, metadata.Alert, "The number of open alerts by current severity.")
+	for k, v := range lastAbnormalCount {
+		collect.Put("alerts.last_abnormal_severity", opentsdb.TagSet{"severity": k}, v)
+	}
+	metadata.AddMetricMeta("bosun.alerts.last_abnormal_severity", metadata.Gauge, metadata.Alert, "The number of open alerts by last abnormal severity.")
```
captncraig: Would it be a horrible idea for some of these to include alert name in the tags and emit one point per alert? Could be cool to graph status of an individual alert over history.

cmd/bosun/sched/check.go
kylebrandt: I did it that was first and I didn't think it looked nice.

kylebrandt:
@captncraig This ready for a review again

cmd/bosun/sched/check.go
```diff
@@ -185,9 +186,63 @@ func (s *Schedule) RunHistory(r *RunHistory) {
...
+		"The number of open alerts by active status.")
+	for _, state := range s.status {
+		if !state.Open {
+			return
+		}
+		// The tagset of the alert is not included because there is no way to
+		// store the string of a group in OpenTSBD in a parsable way. This is
+		// because any delimiter we chose could also be part of a tag key or tag
+		// value.
+		ts := opentsdb.TagSet{"alert": state.Alert}
```
mjibson: Won't this cause data with the same tagset but different values?

cmd/bosun/sched/check.go
```diff
@@ -185,9 +186,63 @@ func (s *Schedule) RunHistory(r *RunHistory) {
 	if checkNotify && s.nc != nil {
 		s.nc <- true
 	}
+	s.CollectStates()
 	s.Save()
 }
 
+// CollectStates sends various state information to bosun with collect.
+func (s *Schedule) CollectStates() {
+	metadata.AddMetricMeta(
```
mjibson: These could be moved to an init block.

cmd/bosun/sched/check.go
```diff
@@ -185,9 +186,63 @@ func (s *Schedule) RunHistory(r *RunHistory) {
...
+		}
+		// The tagset of the alert is not included because there is no way to
+		// store the string of a group in OpenTSBD in a parsable way. This is
+		// because any delimiter we chose could also be part of a tag key or tag
+		// value.
+		ts := opentsdb.TagSet{"alert": state.Alert}
+		severity := state.Status().String()
+		lastAbnormal := state.AbnormalStatus().String()
+		var err error
+		err = collect.Put("alerts.current_severity", ts.Copy().Merge(opentsdb.TagSet{"severity": severity}), 1.0)
```
mjibson: Why are we using tags to indicate data instead of values? Alert statuses have integer values. This avoids the problem where you're only sending data when something is present instead of sending whether something is present (which causes unknowns). The argument I can think of to use tags, though, is that the alert status can change if we add others, and the tag strings won't. I'm still leaning toward the integer way, though.

cmd/bosun/sched/check.go
```diff
@@ -185,9 +186,63 @@ func (s *Schedule) RunHistory(r *RunHistory) {
...
+		}
+		err = collect.Put("alerts.last_abnormal_severity", ts.Copy().Merge(opentsdb.TagSet{"severity": lastAbnormal}), 1.0)
+		if err != nil {
+			log.Println(err)
+		}
+		ackString := "acknowledged"
+		if state.NeedAck {
+			ackString = "unacknowledged"
+		}
+		err = collect.Put("alerts.acknowledgement_status", ts.Copy().Merge(opentsdb.TagSet{"status": ackString}), 1.0)
```
mjibson: Again, why not use a value of 0 or 1 to indicate ack'd, instead of tags?

cmd/bosun/sched/check.go
kylebrandt: Will do: "These could be moved to an init block."

cmd/bosun/sched/check.go
kylebrandt: "Won't this cause data with the same tagset but different values?" Yes that was pretty stupid, think I will map[alert.name]int64 for the counts, and init all alertnames to zero.

cmd/bosun/sched/check.go
kylebrandt: "Why are we using tags to indicate data instead of values? Alert statuses have integer values. This avoids the problem where you're only sending data when something is present instead of sending whether something is present (which causes unknowns). The argument I can think of to use tags, though, is that the alert status can change if we add others, and the tag strings won't. I'm still leaning toward the integer way, though."

I want the aggregation of these things to provide counts. Severity is mostly categorical:

``` 
const (
	StNone Status = iota
	StNormal
	StWarning
	StCritical
	StUnknown
	StError
)
```

So in isn't really true that `StNormal + StNormal = StWarning`, or `StUnknown / StWarning == StNormal`. This would only really make sense as some sort of abstract "severity score" and that isn't what I'm after.

cmd/bosun/sched/check.go
kylebrandt: "Again, why not use a value of 0 or 1 to indicate ack'd, instead of tags? Same for active below." I want to count them, and if I use zero I can't count via sum aggregation things with a zero value.

mjibson: Ok.

cmd/bosun/sched/check.go
mjibson: Ok.

cmd/bosun/sched/check.go
```diff
@@ -185,9 +201,87 @@ func (s *Schedule) RunHistory(r *RunHistory) {
...
+	// [AlertName][Severity]Count
+	severityCounts := make(map[string]map[string]int64)
+	abnormalCounts := make(map[string]map[string]int64)
+	ackStatusCounts := make(map[string]map[bool]int64)
+	activeStatusCounts := make(map[string]map[bool]int64)
+	// Initalize the Counts
+	for _, alert := range s.Conf.Alerts {
+		severityCounts[alert.Name] = make(map[string]int64)
+		abnormalCounts[alert.Name] = make(map[string]int64)
+		for _, severity := range StatusNames {
```
mjibson: I think it's bad to do this this way. I would instead recommend making a statusLast status, and then loop from 0 to statusLast and convert it to a string. This way you don't need that silly slice.

cmd/bosun/sched/check.go
```diff
@@ -185,9 +201,88 @@ func (s *Schedule) RunHistory(r *RunHistory) {
...
+	severityCounts := make(map[string]map[string]int64)
+	abnormalCounts := make(map[string]map[string]int64)
+	ackStatusCounts := make(map[string]map[bool]int64)
+	activeStatusCounts := make(map[string]map[bool]int64)
+	// Initalize the Counts
+	for _, alert := range s.Conf.Alerts {
+		severityCounts[alert.Name] = make(map[string]int64)
+		abnormalCounts[alert.Name] = make(map[string]int64)
+		var i Status
+		for i = 0; i.String() != "none"; i++ {
```
mjibson: Needs to start at 1 because 0 is already "none".

PR #857, opened 2015-03-31
Add HAProxy collector

alienth:
Much of this code is from @kylebrandt, however I have done despicable things to it under the guise of "finishing" it.

I've modeled much of the multiple-fetch off of processes_linux. The config file will look something like the following:

```
haproxy = username:password
haproxy_instance = 1:http://lb01:7001/haproxy;csv
haproxy_instance = 2:http://lb01:7002/haproxy;csv
```

Splitting the instances up is primarily to improve readability and prevent very long lines.

This original code made some assumptions regarding haproxy layout, and I've furthered piled on more assumptions leading to a very assumptive tower of pancakes. The assumptions are as follows:

1. haproxy stats will always be secured with user/pass HTTP auth.
1. Multiple haproxy instances on one server are tagged as 'tiers', a bit of a misnomer in almost all cases (but existing terminology in ours).
1. A single set of credentials is used for all haproxy instances running. This is designed this way primarily to prevent repeating the credentials over and over, and also to make most of the config file viewable after it has been blackboxed.
1. If you have an 'haproxy' config line, you'll also have at least one 'haproxy_instance' config line.

As a side note, another reason I discarded @kylebrandt's config line method is that we cannot guarantee what characters may exist in the password, and as such a credential line can be a bit scary to parse (unless you use a connection-string like format).

So, there you have it. Let the code evisceration commence.

:eyeglasses: @kylebrandt  @mjibson

### DIFF
--- a/cmd/scollector/collectors/haproxy_unix.go
+++ b/cmd/scollector/collectors/haproxy_unix.go
@@ -1,0 +1,486 @@
+package collectors
+
+import (
+	"encoding/csv"
+	"fmt"
+	"net/http"
+	"strconv"
+	"strings"
+
+	"bosun.org/metadata"
+	"bosun.org/opentsdb"
+)
+
+// HAProxy registers an HAProxy collector.
+func HAProxy(user, pwd, url, tier string) {
+	collectors = append(collectors, &IntervalCollector{
+		F: func() (opentsdb.MultiDataPoint, error) {
+			return c_haproxy_csv(user, pwd, url)
+		},
+		name: fmt.Sprintf("haproxy-%s", url, tier),
+	})
+}
+
+func c_haproxy_csv(user, pwd, url, tier string) (opentsdb.MultiDataPoint, error) {
+	var md opentsdb.MultiDataPoint
+	const metric = "haproxy"
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return nil, err
+	}
+	req.SetBasicAuth(user, pwd)
+	client := &http.Client{}
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+	//csvFile, err := os.Open("/ha.csv")
+	if err != nil {
+		return nil, err
+	}
+	//defer csvFile.Close()
+	reader := csv.NewReader(resp.Body)
+	records, err := reader.ReadAll()
+	if err != nil {
+		return nil, err
+	}
+	if len(records) < 2 {
+		return nil, nil
+	}
+	parse := func(v string) (int64, error) {
+		var i int64
+		if v != "" {
+			i, err = strconv.ParseInt(v, 10, 64)
+			if err != nil {
+				return 0, err
+			}
+			return i, nil
+		}
+		return i, nil
+	}
+	for _, rec := range records[1:] {
+		// There is a trailing comma in haproxy's csv
+		if len(rec) != len(haproxyCsvMeta)+1 {
+			return nil, fmt.Errorf("expected %v lines. got: %v",
+				len(haproxyCsvMeta)+1, len(rec))
+		}
+		hType := haproxyType[rec[32]]
+		pxname := rec[0]
+		svname := rec[1]
+		ts := opentsdb.TagSet{"pxname": pxname, "svname": svname}
+		// TODO MERGE IN INSTANCE TAG
+		for i, field := range haproxyCsvMeta {
+			m := strings.Join([]string{metric, hType, field.Name}, ".")
+			switch i {
+			case 0, 1, 26, 27, 28, 31, 32, 56, 57:
+				continue
+			case 39, 40, 41, 42, 43, 44:
+				sp := strings.Split(field.Name, "_")
+				if len(sp) != 2 {
+					return nil, fmt.Errorf("unexpected field name %v in hrsp", field.Name)
+				}
+				ts := ts.Copy().Merge(opentsdb.TagSet{"status_code": sp[1]})
+				m = strings.Join([]string{metric, hType, sp[0]}, ".")
+				v, err := parse(rec[i])
+				if err != nil {
+					return nil, err
+				}
+				Add(&md, m, v, ts, metadata.Counter, metadata.Response,
+					fmt.Sprintf("The number of http responses with a %v status code.", sp[1]))
+			case 17:
+				v, ok := haproxyStatus[rec[i]]
+				// Not distinging between MAINT and MAINT via...
+				if !ok {
+					v = 3
+				}
+				Add(&md, m, v, ts, field.RateType, field.Unit, field.Desc)
+			case 36:
+				if rec[i] == "" {
+					continue
+				}
+				v, ok := haproxyCheckStatus[rec[i]]
+				if !ok {
+					return nil, fmt.Errorf("unknown check status %v", rec[i])
+				}
+				Add(&md, m, v, ts, field.RateType, field.Unit, field.Desc)
+			default:
+				v, err := parse(rec[i])
+				if err != nil {
+					return nil, err
+				}
+				Add(&md, m, v, ts, field.RateType, field.Unit, field.Desc)
+			}
+		}
+	}
+	return md, nil
+}
+
+type MetricMetaName struct {
+	Name string
+	MetricMeta
+}
+
+var haproxyType = map[string]string{
+	"0": "frontend",
+	"1": "backend",
+	"2": "server",
+	"3": "listen",
+}
+
+var haproxyCheckStatus = map[string]float64{
+	"UNK":     0,
+	"INI":     1,
+	"SOCKERR": 2,
+	"L4OK":    3,
+	"L4TMOUT": 4,
+	"L4CON":   5,
+	"L6OK":    6,
+	"L6TOUT":  7,
+	"L6RSP":   8,
+	"L7OK":    9,
+	"L7OKC":   10,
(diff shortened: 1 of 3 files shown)


cmd/scollector/main.go
```diff
@@ -39,6 +39,7 @@ var (
 	flagBatchSize       = flag.Int("b", 0, "OpenTSDB batch size. Used for debugging bad data.")
 	flagSNMP            = flag.String("s", "", "SNMP host to poll of the format: \"community@host[,community@host...]\".")
 	flagICMP            = flag.String("i", "", "ICMP host to ping of the format: \"host[,host...]\".")
+	flagHAProxy         = flag.String("haproxy", "", `haproxy credentials, of the format: "user:password"`)
```
captncraig: if `haproxy_instance` can only be supplied via config file, why do we need a flag for the credentials? Is it just to keep sensitive info out of the config file? I would rather not add flags if there is no way to use this feature without a config file anyway.

alienth: Yeah, this was part of the original design before I added the extra config lines. I'm fine with killing this as a flag. Will do so.

cmd/scollector/collectors/haproxy_unix.go
```diff
@@ -0,0 +1,518 @@
...
+
+type trackedInstance struct {
+	Tier     string
+	URL      string
+	Username string
+	Password string
+}
+
+// HAProxy registers an HAProxy collector.
+func HAProxy(user, pwd string) {
```
captncraig: I like method names to be good verbs. This is a noun, and I am not sure what it does when I See it in main. Maybe `StartHAProxyCollector` or something along those lines.

cmd/scollector/collectors/haproxy_unix.go
```diff
@@ -0,0 +1,518 @@
...
+	"bosun.org/metadata"
+	"bosun.org/opentsdb"
+)
+
+var trackedInstances = []*trackedInstance{}
+
+type trackedInstance struct {
+	Tier     string
+	URL      string
+	Username string
```
captncraig: Why does each instance have a username and password? You are already passing the un/pw to most methods, just use it via parameters rather than saving it to each instance struct.

cmd/scollector/collectors/haproxy_unix.go
```diff
@@ -0,0 +1,518 @@
...
+	instance.URL = sp[1]
+	trackedInstances = append(trackedInstances, &instance)
+
+	return nil
+}
+
+func cHAProxyCSV(user, pwd string) (opentsdb.MultiDataPoint, error) {
+	var md opentsdb.MultiDataPoint
+	var err error
+	client := http.Client{}
```
captncraig: i don't see a reason the http client can't be created inside of `fetchHAProxyData`

cmd/scollector/collectors/haproxy_unix.go
```diff
@@ -0,0 +1,518 @@
...
+}
+
+var haproxyStatus = map[string]int{
+	"UP":    0,
+	"DOWN":  1,
+	"NOLB":  2,
+	"MAINT": 3,
+}
+
+var haproxyCSVMeta = []MetricMetaName{
```
captncraig: maybe we could come up with a data structure that combines the bosun metadata with the haproxy field index metadata and drive all of the logic from data rather than the magic number switch thing.

cmd/scollector/collectors/haproxy_unix.go
```diff
@@ -0,0 +1,518 @@
...
+				len(haproxyCSVMeta)+1, len(rec))
+		}
+		hType := haproxyType[rec[32]]
+		pxname := rec[0]
+		svname := rec[1]
+		ts := opentsdb.TagSet{"pxname": pxname, "svname": svname, "tier": instance.Tier}
+		for i, field := range haproxyCSVMeta {
+			m := strings.Join([]string{metric, hType, field.Name}, ".")
+			switch i {
+			case 0, 1, 26, 27, 28, 31, 32, 56, 57:
```
captncraig: maybe these ignored ones could be implemented by an `ignore: true` field on the `MetricMetaName` struct.

cmd/scollector/collectors/haproxy_unix.go
```diff
@@ -0,0 +1,518 @@
...
+					return err
+				}
+				Add(md, m, v, ts, field.RateType, field.Unit, field.Desc)
+			}
+		}
+	}
+	return nil
+}
+
+type MetricMetaName struct {
```
captncraig: horrible struct name

cmd/scollector/collectors/haproxy_unix.go
```diff
@@ -0,0 +1,518 @@
...
+			return fmt.Errorf("expected %v lines. got: %v",
+				len(haproxyCSVMeta)+1, len(rec))
+		}
+		hType := haproxyType[rec[32]]
+		pxname := rec[0]
+		svname := rec[1]
+		ts := opentsdb.TagSet{"pxname": pxname, "svname": svname, "tier": instance.Tier}
+		for i, field := range haproxyCSVMeta {
+			m := strings.Join([]string{metric, hType, field.Name}, ".")
+			switch i {
```
captncraig: This might make sense as an if/else chain rather than a switch. Then you can look at more data than just the index for the conditionals and make them make a bit more sense.

cmd/scollector/collectors/haproxy_unix.go
```diff
@@ -0,0 +1,518 @@
...
+			case 0, 1, 26, 27, 28, 31, 32, 56, 57:
+				continue
+			case 39, 40, 41, 42, 43, 44:
+				sp := strings.Split(field.Name, "_")
```
captncraig: this could be `if field.Name startsWith (hrsp) for example

cmd/scollector/collectors/haproxy_unix.go
captncraig: maybe a comment linking to http://www.haproxy.org/download/1.5/doc/configuration.txt or similar document as the source of this information?

cmd/scollector/collectors/haproxy_unix.go
```diff
@@ -0,0 +1,518 @@
...
+				}
+				ts := ts.Copy().Merge(opentsdb.TagSet{"status_code": sp[1]})
+				m = strings.Join([]string{metric, hType, sp[0]}, ".")
+				v, err := parse(rec[i])
+				if err != nil {
+					return err
+				}
+				Add(md, m, v, ts, metadata.Counter, metadata.Response,
+					fmt.Sprintf("The number of http responses with a %v status code.", sp[1]))
+			case 17:
```
captncraig: and these last two cases can be picked out by name as well.

cmd/scollector/collectors/haproxy_unix.go
```diff
@@ -0,0 +1,518 @@
...
+		if len(rec) != len(haproxyCSVMeta)+1 {
+			return fmt.Errorf("expected %v lines. got: %v",
+				len(haproxyCSVMeta)+1, len(rec))
+		}
+		hType := haproxyType[rec[32]]
+		pxname := rec[0]
+		svname := rec[1]
+		ts := opentsdb.TagSet{"pxname": pxname, "svname": svname, "tier": instance.Tier}
+		for i, field := range haproxyCSVMeta {
+			m := strings.Join([]string{metric, hType, field.Name}, ".")
```
captncraig: maybe pull save `rec[i]` into a variable at the top of this loop so we don't have to pull it out that way in every case.

cmd/scollector/main.go
kylebrandt: Was mostly for development, probably would have killed it as well.

cmd/scollector/collectors/haproxy_unix.go
kylebrandt: This is consistent with all our other collectors that that register themselves. So this should stay the same name - if we want to change it that is fine, but it should be in a commit that changes all of them at once.

cmd/scollector/collectors/haproxy_unix.go
```diff
@@ -0,0 +1,518 @@
...
+		}
+		return i, nil
+	}
+
+	req, err := http.NewRequest("GET", instance.URL, nil)
+	if err != nil {
+		return err
+	}
+	req.SetBasicAuth(instance.Username, instance.Password)
+	resp, err := client.Do(req)
```
mjibson: Just use `http.DefaultClient.Do(req)` here if you don't need to change any defaults.

alienth:
:haircut:

Suggested changes made. Will squash down before I merge.

cmd/scollector/collectors/haproxy_unix.go
```diff
@@ -0,0 +1,528 @@
...
+				}
+				v, ok := haproxyCheckStatus[value]
+				if !ok {
+					return fmt.Errorf("unknown check status %v", value)
+				}
+				Add(md, m, v, ts, field.RateType, field.Unit, field.Desc)
+			} else {
+				v, err := parse(value)
+				if err != nil {
+					return err
```
captncraig: Do we want to return errors in these cases, or continue to the next row?

alienth: I believe the intent there is to throw an error and stop processing. If any field fails to parse, then something odd is happening somewhere.

captncraig: Fair enough.

captncraig:
:+1:

PR #860, opened 2015-04-01
Implement function DropGe, to complement the already implemented DropLe

vitorboschi:
Add a missing function to Bosun. I plan to send a PR updating the docs for DropGe and DropLe (which is missing documentation) soon


vitorboschi:
You're right, e implies equal. The operator are that way because the code inside the if actually copies the element to the return series. I found it confusing too, and just followed the same style.
I will update the PR to merge the two functions, and try to make the code a bit more clear.

kylebrandt:
This looks okay to @mjibson or @captncraig can you sign off?

PR #862, opened 2015-04-01
cmd/bosun: move more state mutation into RunHistory

mjibson:
Currently GetOrCreateStatus (only used in the web package) is the only
known place outside of RunHistory where state is mutated. This change moves
things like Touched and Result mutation into RunHistory.

This should prevent the recent panics because now the web package won't
create alert keys that shouldn't exist.

### DIFF
--- a/cmd/bosun/sched/check.go
+++ b/cmd/bosun/sched/check.go
@@ -40,7 +40,14 @@ func NewStatus(ak expr.AlertKey) *State {
 	}
 }
 
-func (s *Schedule) Status(ak expr.AlertKey) *State {
+func (s *Schedule) GetStatus(ak expr.AlertKey) *State {
+	s.Lock()
+	state := s.status[ak]
+	s.Unlock()
+	return state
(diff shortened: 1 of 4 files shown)


cmd/bosun/sched/check.go
```diff
@@ -40,7 +40,14 @@ func NewStatus(ak expr.AlertKey) *State {
 	}
 }
 
-func (s *Schedule) Status(ak expr.AlertKey) *State {
+func (s *Schedule) GetStatus(ak expr.AlertKey) *State {
```
captncraig: Should this just return the status to remove the possibility of thinking you can do stuff with it?

mjibson: Sadly, no. The web package is the only user of GetStatus and GetOrCreateStatus, and both of them need the full structure, as far as I can tell. Our test rule page does some crazy things.

captncraig:
lgtm

PR #870, opened 2015-04-03
cmd/bosun: Ensure web instances have their result set

mjibson:
Fixes a recent panic. Correct fix would be to refactor the web package to
use RunHistory, which already handled this correctly.


captncraig:
:+1:

PR #874, opened 2015-04-07
Adding configuration option to set content-type on post notification.

### LINKED ISSUE #873: `application/x-www-form-urlencoded` is hardcoded
HTTP POST/PUT can handle different 'Content-Type'-s, the above is hard-coded in [several places](https://github.com/bosun-monitor/bosun/search?utf8=%E2%9C%93&q=x-www-form-urlencoded).

captncraig:
Fixes #873

### DIFF
--- a/cmd/bosun/conf/conf.go
+++ b/cmd/bosun/conf/conf.go
@@ -316,13 +316,14 @@ type Template struct {
 type Notification struct {
 	ConfItem
 	Vars
-	Name      string
-	Email     []*mail.Address
-	Post, Get *url.URL
-	Body      *ttemplate.Template
-	Print     bool
-	Next      *Notification
-	Timeout   time.Duration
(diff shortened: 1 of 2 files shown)


cmd/bosun/conf/conf.go
```diff
@@ -1080,6 +1082,8 @@ func (c *Conf) loadNotification(s *parse.SectionNode) {
 			n.Get = get
 		case "print":
 			n.Print = true
+		case "contentType":
+			n.ContentType = v
```
captncraig: We could potentially do a `mime.ParseMediaType(v)` for validation here, but I am not sure it is necessary.

gbrayut: Would recommend against as custom mime types are possible. So long as we default to a good value let the user pick whatever they want.

PR #877, opened 2015-04-08
Adding dash to the metric-name regex

### LINKED ISSUE #876: Valid metric names produce 'Bad Query Format` errors
Metrics with dashes fail validation and cause: `opentsdb: bad query format: sum:ipv4-bytes.bytes.counter.value` errors.

captncraig:
Fixes #876


IzakMarais:
Wow, thanks for the very fast fix on this.

PR #868, opened 2015-04-02
Config/Rule page unification

captncraig:
New route at `/config2` (temporary) for new combined config/rule page.

PR #880, opened 2015-04-09
cmd/bosun: check for template before using

mjibson:


mjibson:
I merged this because it needs to go out, but I'd appreciate a review anyways. Thanks.

captncraig:
Looks fine. That function will be deleted with #868 anyway.

PR #881, opened 2015-04-09
cmd/bosun: add UA-Compat meta tag for IE

mjibson:
This prevents it going into IE7 mode on intranets.


mjibson:
See: http://stackoverflow.com/questions/13949858/why-does-ie9-opens-in-document-mode-as-ie7-standards

PR #884, opened 2015-04-10
Use TCP to insert data into OpenTSDB along with HTTP

santosh-d3vpl3x:
Hi Guys

We use OpenTSDB 2.0.1 for our needs and we could not use scollector's HTTP data GZIPed and chunked insertion. So we decided to tweek it to send stats over TCP by default. Also we have modified the code to fetch configurations related to frequency from `scollector.conf`

Please let me know if there are any issues.

### DIFF
--- a/cmd/scollector/main.go
+++ b/cmd/scollector/main.go
@@ -150,7 +150,7 @@ func readConf() {
 
 func main() {
 	flag.Parse()
-	if *flagPrint || *flagDebug  {
+	if *flagPrint || *flagDebug {
 		slog.Set(&slog.StdLog{Log: log.New(os.Stdout, "", log.LstdFlags)})
 	}
 	if *flagVersion {
--- a/cmd/scollector/main.go
+++ b/cmd/scollector/main.go
@@ -34,7 +34,7 @@ var (
 	flagFilter          = flag.String("f", "", "Filters collectors matching this term, multiple terms separated by comma. Works with all other arguments.")
 	flagList            = flag.Bool("l", false, "List available collectors.")
 	flagPrint           = flag.Bool("p", false, "Print to screen instead of sending to a host")
-	flagTCP             = flag.Bool("tcp", false, "Use TCP instead of HTTP")
+	flagTelnet          = flag.Bool("telnet", false, "Use telnet instead of HTTP")
 	flagHost            = flag.String("h", "", `Bosun or OpenTSDB host. Ex: "http://bosun.example.com:8070".`)
 	flagColDir          = flag.String("c", "", `External collectors directory.`)
 	flagBatchSize       = flag.Int("b", 0, "OpenTSDB batch size. Used for debugging bad data.")
@@ -271,9 +271,8 @@ func main() {
 	if *flagPrint {
 		collect.Print = true
 	}
-
-	if *flagTCP {
-		collect.TCP = true
+	if *flagTelnet {
+		collect.Telnet = true
 	}
 	if !*flagDisableMetadata {
 		if err := metadata.Init(u, *flagDebug); err != nil {
--- a/cmd/scollector/main.go
+++ b/cmd/scollector/main.go
@@ -49,7 +49,7 @@ var (
 	flagVersion         = flag.Bool("version", false, "Prints the version and exits.")
 	flagDisableDefault  = flag.Bool("n", false, "Disable sending of scollector self metrics.")
 	flagHostname        = flag.String("hostname", "", "If set, use as value of host tag instead of system hostname.")
-	flagFreq            = flag.String("freq", "15", "Set the default frequency in seconds for most collectors.")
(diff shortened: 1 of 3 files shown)


cmd/scollector/main.go
```diff
@@ -48,7 +49,7 @@ var (
 	flagVersion         = flag.Bool("version", false, "Prints the version and exits.")
 	flagDisableDefault  = flag.Bool("n", false, "Disable sending of scollector self metrics.")
 	flagHostname        = flag.String("hostname", "", "If set, use as value of host tag instead of system hostname.")
-	flagFreq            = flag.String("freq", "15", "Set the default frequency in seconds for most collectors.")
```
captncraig: Why did you remove this default? Scollector now *cannot* be run without setting a `freq` flag. That is not good.

cmd/scollector/main.go
```diff
@@ -34,6 +34,7 @@ var (
 	flagFilter          = flag.String("f", "", "Filters collectors matching this term, multiple terms separated by comma. Works with all other arguments.")
 	flagList            = flag.Bool("l", false, "List available collectors.")
 	flagPrint           = flag.Bool("p", false, "Print to screen instead of sending to a host")
+	flagHTTP            = flag.Bool("http", false, "Use HTTP instead of TCP")
```
captncraig: The default behaviour should not be changed. Reverse this logic and add a `tcp` flag that defaults to false.

cmd/scollector/main.go
```diff
@@ -149,7 +150,7 @@ func readConf() {
 
 func main() {
 	flag.Parse()
-	if *flagPrint || *flagDebug {
+	if *flagPrint || *flagDebug || *flagHTTP {
 		slog.Set(&slog.StdLog{Log: log.New(os.Stdout, "", log.LstdFlags)})
```
captncraig: I don't think http mode should setup logging this way. This is another breaking change to the default behaviour.

santosh-d3vpl3x: Thanks for such a quick review. 

The changes made above were influenced by inability to make respective flags configurable through configuration file instead of arguments. I faced this issue with both frequency and tcp/http switch and it is reflected by the code. 

If you could let me know, the way to do so , I will make those changes in our code base as they are desired.

Meanwhile, changes would be made for:

1. restore `freq` default to `15`.
1. make `http` as default way to send metrics. 


`*flagHTTP` is a mistake on my part, I will fix it.

captncraig: The general idea is that most settings are available via flag **and** conf file. As it reads the config file it simply overrides the flag value if it finds that in the conf file. 

Generally all we need to do is add a flag and a matching case in `readConf()`.

Feel free to join us in our slack room. I sent you an invite.

santosh-d3vpl3x:
As discussed on Slack, I have removed conf/command line arg related settings as there are few issues with it's current implementation. 

###### Changes Done
- Removed changes related to `freq` as it would have lead to change in default implementation.
- Switched to HTTP as default mode for sending data from TCP mode.

Usage:
add `-tcp` as a flag to send data over TCP instead of through HTTP request.

kylebrandt:
What does sending over TCP mean? Http uses TCP as well, so what is the
protocol on top of TCP here?

nickman:
I guess the correct term would be Telnet.

kylebrandt:
Okay that makes sense, please rename things accordingly.

santosh-d3vpl3x:
Well, I knew this will confuse someone. 

Here is the clarification:
scollector uses http endpoint provided by OpenTSDB to transfer data. Whereas, this branch opens TCP sockets to transfer data directly. TCP is the correct term over here. 

telnet and http both are application level protocols[High level] where as TCP is a transport level protocol[Low level]. 

Could you please let me know why there is any confusion.

kylebrandt:
Agreed, I'm not sure of the details of OpenTSDB's implementation here, but they do call it the telnet API. TCP is confusing since both application level protocols (Telnet and HTTP) use TCP as their transport protocol. Therefore I think we should just stick with OpenTSDB's terminology.

Reference:
http://opentsdb.net/docs/build/html/user_guide/writing.html#input-methods

collect/queue.go
```diff
@@ -68,66 +71,110 @@ func sendBatch(batch []json.RawMessage) {
...
-	resp, err := client.Do(req)
-	d := time.Since(now).Nanoseconds() / 1e6
-	if err == nil {
-		defer resp.Body.Close()
-	}
-	Add("collect.post.total_duration", Tags, d)
-	Add("collect.post.count", Tags, 1)
-	// Some problem with connecting to the server; retry later.
-	if err != nil || resp.StatusCode != http.StatusNoContent {
+	if Telnet {
```
mjibson: This diff is not very readable as it stands. Would you instead make a telnet func that does the sending, and just `if Telnet { thatFunc(); return }`?

PR #887, opened 2015-04-10
TravisCI: exclude all . and _ folders from gofmt checks.

gbrayut:


build/validate.sh
```diff
@@ -1,14 +1,17 @@
...
-if [ "$GOFMTOUT" != '' ]; then 
-    echo "The following files need gofmt -s:";
-    echo "$GOFMTOUT";
-    exit 1;
+echo -e "\nChecking gofmt -l -s -w for all folders that don't start with . or _"
+cd $GOPATH/src/bosun.org
+GOFMTRESULT=0
+GOFMTOUT=$(find . -maxdepth 1 -type d -iregex './[^._].*' -exec gofmt -l -s -w {} +);
+if [ "$GOFMTOUT" != '' ]; then
+    echo "The following files need 'gofmt -l -s -w':"
```
mjibson: `gofmt -s -w` is correct, no `-l`

build/validate.sh
```diff
@@ -1,14 +1,17 @@
...
-echo -e "\nChecking go fmt bosun.org/..."
-GOFMTOUT=$(go fmt -s bosun.org/...);
-if [ "$GOFMTOUT" != '' ]; then 
-    echo "The following files need gofmt -s:";
-    echo "$GOFMTOUT";
-    exit 1;
+echo -e "\nChecking gofmt -l -s -w for all folders that don't start with . or _"
+cd $GOPATH/src/bosun.org
+GOFMTRESULT=0
+GOFMTOUT=$(find . -maxdepth 1 -type d -iregex './[^._].*' -exec gofmt -l -s -w {} +);
```
mjibson: can we save the output of find here to a variable, which we can then use with other programs, like vet?

build/validate.sh
gbrayut: we need to use -l to get the list of files that changed:

> C:\Code\Go\src\bosun.org [travis_fmt +0 ~1 -0]> gofmt -?
> flag provided but not defined: -?
> usage: gofmt [flags] [path ...]
>   -cpuprofile="": write cpu profile to this file
>   -d=false: display diffs instead of rewriting files
>   -e=false: report all errors (not just the first 10 on different lines)
>   -l=false: list files whose formatting differs from gofmt's
>   -r="": rewrite rule (e.g., 'a[b:len(a)] -> a[b:]')
>   -s=false: simplify code
>   -w=false: write result to (source) file instead of stdout

build/validate.sh
gbrayut: I'd say just always use find. Probably is a different flag we could use besides -exec, but there may be different regexes we want to use for other commands.

this is a pretty standard bash approach to running commands on specific files/folders taken from http://stackoverflow.com/a/13319190/17373

build/validate.sh
mjibson: But we do not need to tell the user that they should use `-l`.

build/validate.sh
mjibson: Well I'd like to be able to do this:

```
DIRS=find . -maxdepth 1 -type d -iregex './[^._].*'
GOFMTOUT=gofmt -l -s -w $DIRS
GOVETOUT=vet $DIRS
```

gbrayut: @mjibson I fixed the -l, added the DIRS variable, and force pushed, but `go vet $DIRS` is complaining about some folders. Should I exclude those when calling go vet?

```
[root@ny-gbraylx01 bosun.org]# go vet $DIRS
can't load package: package bosun.org/cmd: no buildable Go source files in /home/gbray/go/src/bosun.org/cmd
can't load package: package bosun.org/plugins: no buildable Go source files in /home/gbray/go/src/bosun.org/plugins
can't load package: package bosun.org/build: no buildable Go source files in /home/gbray/go/src/bosun.org/build
```

Issue #754, opened 2015-03-03
Scollector c_apache_mod_info error due to IIS web server listening on all IPs


gbrayut:
ny-promoweb01 and or-promoweb01 are IIS servers that have the following errors from scollector:

'interval.go:53: bosun.org/cmd/scollector/collectors.c_apache_mod_info: fewer than 4 fields found on connection:/keep-alive line'

Looks like the apache collector is enabled [via a URL](https://github.com/bosun-monitor/bosun/blob/master/cmd/scollector/collectors/apache_mod_info.go#L25) which on this server happens to have a 302 redirect served by IIS.

Probably best to extend the [enableURL](https://github.com/bosun-monitor/bosun/blob/master/cmd/scollector/collectors/interval.go#L80) method to use a regex to validate the results or return headers and/or allow a [CheckRedirect](http://golang.org/pkg/net/http/#Client) function to say it shouldn't follow 302 redirects.

kylebrandt:
In addition I would make this a Unix collector until someone requests support for apache run on windows

gbrayut:
This is also happening on unix systems.
```
Mar  5 22:48:22 ny-realog01 /opt/scollector/scollector[26207]: error: interval.go:53: bosun.org/cmd/scollector/collectors.c_apache_mod_info: fewer than 4 fields found on connection:/keep-alive line
Mar  5 23:18:22 ny-realog01 /opt/scollector/scollector[26207]: error: interval.go:53: bosun.org/cmd/scollector/collectors.c_apache_mod_info: fewer than 4 fields found on connection:/keep-alive line
```

PR #890, opened 2015-04-10
cmd/scollector: more lenient apache_mod_info collector

mjibson:
Some HTTP servers, like Go's, do a best-match for routing. Thus, when
handling "/" locally, it will match "/server-info", triggering the
apache_mod_info collector to enable. But since it's not that page,
extracting the info lines fails. Search for the a string unique to that
URL on Apache servers.


gbrayut:
you can mark this a fixing https://github.com/bosun-monitor/bosun/issues/754

mjibson:
Hold off on merging this PR. I think we have a better solution.

PR #892, opened 2015-04-10
cmd/scollector: use a regex to check for apache info

### LINKED ISSUE #754: Scollector c_apache_mod_info error due to IIS web server listening on all IPs

mjibson:
fixes #754
closes #890

### DIFF
--- a/cmd/scollector/collectors/apache_mod_info_linux.go
+++ b/cmd/scollector/collectors/apache_mod_info_linux.go
@@ -22,7 +22,7 @@ func init() {
 		collectors,
 		&IntervalCollector{
 			F:        c_apache_mod_info,
-			Enable:   enableURL(apacheModInfoBaseURL),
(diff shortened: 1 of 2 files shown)


captncraig:
:+1:

PR #894, opened 2015-04-13
cmd/scollector: Add dsc.mof file metrics

gbrayut:
This is a first pass at adding basic DSC metrics to scollector. Right now it is just monitoring the size and last modified time of the configuration files. Next I need to add the WMI data from the MSFT_DSCLocalConfigurationManager, but that requires calling a WMI function, so it may or may not work in go yet. I'd like to merge this first to start getting basic stats.

Currently running on two systems pointed at devbosun. See http://goo.gl/FZB5i0 for example query

### DIFF
--- a/cmd/scollector/collectors/dsc_windows.go
+++ b/cmd/scollector/collectors/dsc_windows.go
@@ -1,0 +1,62 @@
+package collectors
+
+import (
+	"os"
+	"time"
+
+	"bosun.org/metadata"
+	"bosun.org/opentsdb"
+)
+
+func init() {
+	collectors = append(collectors, &IntervalCollector{F: c_dsc_mof, Interval: time.Minute * 5})
+}
+
+const (
+	dscLCM = "dsc.lcm."
+	dscMof = "dsc.mof."
+)
+
+var (
+	dscpath     = os.ExpandEnv(`${SYSTEMROOT}\system32\Configuration\`)
+	mapMofFiles = map[string]string{
+		"MetaConfig.mof":       "Meta_Config",
+		"Current.mof":          "Current_Config",
+		"backup.mof":           "Backup_Config",
+		"pending.mof":          "Pending_Config",
+		"DSCStatusHistory.mof": "DSC_History",
+		"DSCEngineCache.mof":   "DSC_Cache",
+	}
+)
+
+// c_dsc_mof monitors the size and last modified time of each mof file.
+// These out of band metrics can be used to verify the DSC WMI Status metrics.
+func c_dsc_mof() (opentsdb.MultiDataPoint, error) {
+	var md opentsdb.MultiDataPoint
+	c := 0
+	if _, err := os.Stat(dscpath + "MetaConfig.mof"); os.IsNotExist(err) {
+		c = 1
+	}
+	Add(&md, dscLCM+"configured", c, nil, metadata.Gauge, metadata.StatusCode, descDSCLCMConfigured)
+	if c == 1 {
+		return md, nil
+	}
+	for filename, filetype := range mapMofFiles {
+		tags := opentsdb.TagSet{"type": filetype}
+		s := int64(-1)
+		l := int64(-1)
+		if fi, fierr := os.Stat(dscpath + filename); fierr == nil {
+			s = fi.Size()
+			l = time.Now().Unix() - fi.ModTime().Unix()
+		}
+		Add(&md, dscMof+"size", s, tags, metadata.Gauge, metadata.Bytes, descDSCMofSize)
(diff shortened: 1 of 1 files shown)


cmd/scollector/collectors/dsc_windows.go
```diff
@@ -0,0 +1,63 @@
...
+		"DSCEngineCache.mof":   "DSC_Cache",
+	}
+)
+
+// c_dsc_mof monitors the size and last modified time of each mof file.
+// These out of band metrics can be used to verify the DSC WMI Status metrics.
+func c_dsc_mof() (opentsdb.MultiDataPoint, error) {
+	var md opentsdb.MultiDataPoint
+	if _, err := os.Stat(dscpath + "MetaConfig.mof"); os.IsNotExist(err) {
+		Add(&md, dscLCM+"configured", 1, nil, metadata.Gauge, metadata.StatusCode, descDSCLCMConfigured)
```
mjibson: For both this and the dscMof stuff, set a value then call Add once:

```
c := 0
if _, err := os.Stat(...); os.IsNotExist(err) {
  c = 1
}
Add(&md, dscLCM+"configured", c, nil, ...)
if c == 1 {
  return
}
```

cmd/scollector/collectors/dsc_windows.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+const (
+	descDSCLCMConfigured = "Indicates if DSC Local Configuration Manager is configured: 0=configured, 1=not configured. If the LCM is not configured then the rest of the dsc.* metrics will be skipped on that server."
+	descDSCMofSize       = "Size of the mof file in bytes or -1 if file does not exist."
+	descDSCMofModified   = "Number of seconds since the mof file was last modified or -1 if file does not exist."
+)
+
+// TODO: implement status metrics using MSFT_DSCLocalConfigurationManager class
+//func c_dsc_wmi_status() (opentsdb.MultiDataPoint, error) {
```
mjibson: remove this line

gbrayut:
Updated with requested changes

cmd/scollector/collectors/dsc_windows.go
```diff
@@ -0,0 +1,63 @@
...
+	Add(&md, dscLCM+"configured", c, nil, metadata.Gauge, metadata.StatusCode, descDSCLCMConfigured)
+	if c == 1 {
+		return md, nil
+	}
+	for filename, filetype := range mapMofFiles {
+		tags := opentsdb.TagSet{"type": filetype}
+		fi, fierr := os.Stat(dscpath + filename)
+		s := int64(-1)
+		l := int64(-1)
+		if fierr == nil {
```
mjibson: merge the `fi, fierr := ...` line with into this if

cmd/scollector/collectors/dsc_windows.go
```diff
@@ -0,0 +1,63 @@
...
+		c = 1
+	}
+	Add(&md, dscLCM+"configured", c, nil, metadata.Gauge, metadata.StatusCode, descDSCLCMConfigured)
+	if c == 1 {
+		return md, nil
+	}
+	for filename, filetype := range mapMofFiles {
+		tags := opentsdb.TagSet{"type": filetype}
+		fi, fierr := os.Stat(dscpath + filename)
+		s := int64(-1)
```
mjibson: Is it better to not report at all instead of `-1`? If someone is doing aggregation, the -1 will throw them off instead of indicate the non-existence problem.

gbrayut: Well some of these files are only present for a short period of time. Like pending, is used to stage before moving to the current.mof. I could use 0 instead of -1, but there is always a possibility that a file size could actually be zero.

Problem with not reporting is it is hard to see in the time line and you can't create alerts on absent/unknown metrics very easily.

mjibson: You could have a separate metric that is 0/1 for existence, and the others just for size. Also, if the MetaConfig.mof file doesn't exist, then none of these will ever run since we return early in that case. Though, maybe that file isn't as short-lived as the others. Up to you. I'm ok leaving it as is.

gbrayut: That is the "master file", so if it doesn't exist the LCM doesn't exist or isn't configured (like a linux system not running puppet). If that file exists, then we need to track these files, and I'd rather not use a separate "exist" metric for each. I don't see any issues with using -1, as even with aggregations it will make sense (-1 will act just like 0, which is conceptually what I would expect)

mjibson: Ok.

PR #895, opened 2015-04-14
collect: Add os tag to scollector.collect.* metrics

gbrayut:
I'd also like to add metadata to these metrics but not sure exactly where to do that. I assume they can just use a goroutine to send the metric metadata every hour, or whatever our current design is.

### DIFF
--- a/cmd/scollector/main.go
+++ b/cmd/scollector/main.go
@@ -266,6 +266,7 @@ func main() {
 	}
 	collectors.DefaultFreq = time.Second * time.Duration(freq)
 	collect.Freq = time.Second * time.Duration(freq)
+	collect.Tags = opentsdb.TagSet{"os": runtime.GOOS}
 	if *flagPrint {
 		collect.Print = true
 	}
--- a/collect/collect.go
+++ b/collect/collect.go
@@ -15,6 +15,7 @@ import (
 	"sync"
 	"time"
 
+	"bosun.org/metadata"
 	"bosun.org/opentsdb"
 )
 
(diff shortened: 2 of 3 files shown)


gbrayut:
Looks like this package is also used by bosun and realog. Those should work fine with this extra tag, or we could default to nil and only set the OS tag for scollector.

gbrayut:
Changed to use an exported collect.Tags variable so that the os tag is only set when used in scollector

mjibson:
The build breaking here was because the go team made a mistake and pushed a bad change to the tools repo. Nothing you did.

mjibson:
LGTM

mjibson:
Metadata uses the metadata package. It's already configured to do the correct thing, just call http://godoc.org/bosun.org/metadata#AddMetricMeta

gbrayut:
Yeah I saw the issue someone filed and linked it in the slack room. Hopefully that doesn't happen too often :-p

I'll try and add metadata to this PR tomorrow. Thanks

gbrayut:
Added metadata. Currently running on ny-winbuild01|ny-rdp01 as a second instance pointing at http://devbosun/

cmd/scollector/main.go
```diff
@@ -266,6 +266,7 @@ func main() {
 	}
 	collectors.DefaultFreq = time.Second * time.Duration(freq)
 	collect.Freq = time.Second * time.Duration(freq)
+	collect.Tags = opentsdb.TagSet{"os": runtime.GOOS}
```
captncraig: Do you need to add host in here, or does collect do that for you?

gbrayut: I think that is done by the check function. It is working on devbosun so something is adding it for us.

captncraig:
LGTM

mjibson:
lgtm

PR #909, opened 2015-04-20
go get -u for esc

captncraig:
This should help alleviate the problem where a local version of esc is older than current. Travis always uses latest.

With this running validate.sh locally should update esc and make sure the output matches.


gbrayut:
LGTM although travis is complaining that it still needs go generate.

gbrayut:
Probably due to the recent changes in https://github.com/mjibson/esc/commits/master

PR #911, opened 2015-04-21
cmd/tsdbrelay: Initial relay implementation

mjibson:


captncraig:
I do think it would be cool to have this available via the `bosun` command rather than as a separate executable. Other than that, this looks good to me.

PR #918, opened 2015-04-23
cmd/bosun: Add syntax highlighting to the rule editor

kylebrandt:
image


cmd/bosun/web/static/js/config.ts
```diff
@@ -54,6 +57,8 @@ bosunControllers.controller('ConfigCtrl', ['$scope', '$http', '$location', '$rou
 	$scope.template_group = search.template_group || '';
 	$scope.items = parseItems();
 	$scope.tab = search.tab || 'results';
+	$scope.aceTheme = 'chrome';
+	$scope.aceMode = 'bosun';
```
captncraig: do these need to be on the scope? Why not just hard code them in the markup?

kylebrandt: They get used again in the toggle function. Plus might want to make theme changeable at some point.

cmd/bosun/web/static/js/ace/mode-bosun.js
```diff
@@ -0,0 +1,167 @@
+ace.define("ace/mode/bosun_highlight_rules",["require","exports","module","ace/lib/oop","ace/mode/text_highlight_rules"], function(require, exports, module) {
```
captncraig: Do any of these requires depend on files we didn't include? I'm not seeing 404s but it looks like maybe something is missing?

PR #921, opened 2015-04-27
WIP Incident Ids

captncraig:
The basic design is that events will be given a unique incident id.

- All sequantial, non-normal events in for an alertkey will share an incident id. 
- A normal event will have an incident id of 0.

There is also a schedule-wide incidents lookup so that we can:
- View all events for an incident via api / ui.
- Lookup all incidents in a given timeframe
- Create incidents at load time from historic event data.


captncraig:
Very basic incident view

screen shot 2015-05-01 at 1 35 09 pm

kylebrandt:
Can probably use the timeline directive as well?

captncraig:
Things matt and I discussed:

- Change incidentId to uint64
- Change signature of Append back to how it was, add a schedule append func
- Add actions to incident view
- Make incidents ids time based historically

cmd/bosun/sched/bolt.go
```diff
@@ -166,6 +171,12 @@ func (s *Schedule) RestoreState() error {
 	if err := decode(dbSilence, &s.Silence); err != nil {
 		log.Println(dbSilence, err)
 	}
+	if err := decode(dbIncidents, &s.Incidents); err != nil {
+		log.Println(dbIncidents, err)
+	}
+	if err := decode(dbNextIncident, &s.nextIncidentId); err != nil {
```
mjibson: Could we just compute this instead of store it? Storing an int seems silly.

cmd/bosun/sched/sched.go
```diff
@@ -605,9 +614,20 @@ func (s *State) Touch() {
...
+	if len(s.History) == 0 || last.Status != event.Status {
+		if sched != nil {
+			if last.IncidentId != 0 {
+				//if last event has incident id and is not closed, we continue it.
+				if incident, ok := sched.Incidents[last.IncidentId]; ok && incident.End.IsZero() {
+					event.IncidentId = last.IncidentId
+				}
+			} else if event.Status != StNormal {
+				//otherwise create new incident on first non-normal event
+				event.IncidentId = sched.createIncident(s.AlertKey(), time.Now().UTC()).Id
```
mjibson: Instead of using `time.Now()` here, move setting `event.Time` from below this block to above it, then use `event.Time`. This will guarantee the event and incidents share exactly the same time.

cmd/bosun/sched/sched.go
```diff
@@ -576,6 +579,12 @@ func (s *Schedule) Action(user, message string, t ActionType, ak expr.AlertKey)
 			return fmt.Errorf("cannot close active alert")
 		}
 		st.Open = false
+		last := st.Last()
+		if last.IncidentId != 0 {
+			if incident, ok := s.Incidents[last.IncidentId]; ok {
```
mjibson: Access to shared map: needs to use incidentLock

cmd/bosun/sched/sched.go
```diff
@@ -704,6 +725,111 @@ func (a ActionType) MarshalJSON() ([]byte, error) {
...
+	Id       uint32
+	Start    time.Time
+	End      time.Time
+	AlertKey expr.AlertKey
+}
+
+func (s *Schedule) createIncident(ak expr.AlertKey, start time.Time) *Incident {
+	s.incidentLock.Lock()
+	defer s.incidentLock.Unlock()
+	if s.Incidents == nil {
```
mjibson: How about we do this during Init instead?

cmd/bosun/sched/sched.go
```diff
@@ -704,6 +725,111 @@ func (a ActionType) MarshalJSON() ([]byte, error) {
...
+		s.Incidents = map[uint32]*Incident{}
```
mjibson: I'm aware that Go has 2 ways of initing memory. But we use make more than `{}`. Could we continue to use make everywhere?

cmd/bosun/sched/sched.go
```diff
@@ -704,6 +725,111 @@ func (a ActionType) MarshalJSON() ([]byte, error) {
...
+	list := []*Incident{}
+	for _, i := range s.Incidents {
+		if alert != "" && i.AlertKey.Name() != alert {
+			continue
+		}
+		if i.Start.Before(from) || i.Start.After(to) {
+			continue
+		}
+		list = append(list, i)
+		if len(list) >= 200 {
```
mjibson: Why is this limiting done here? It should perhaps go into the web controller itself.

cmd/bosun/sched/sched.go
```diff
@@ -704,6 +725,111 @@ func (a ActionType) MarshalJSON() ([]byte, error) {
...
+	}
+	return list
+}
+
+func (s *Schedule) GetIncidentEvents(id uint32) (*Incident, []Event, error) {
+	s.incidentLock.Lock()
+	incident, ok := s.Incidents[id]
+	s.incidentLock.Unlock()
+	if !ok {
+		return nil, nil, fmt.Errorf("Incident %d not found.", id)
```
mjibson: Errors should be lowercase and not end in a period: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

cmd/bosun/sched/sched.go
```diff
@@ -704,6 +725,111 @@ func (a ActionType) MarshalJSON() ([]byte, error) {
...
+	list := []Event{}
+	state := s.GetStatus(incident.AlertKey)
+	if state == nil {
+		return incident, list, nil
+	}
+	found := false
+	for i, e := range state.History {
+		if e.IncidentId == id {
+			if !found && i > 0 {
+				list = append(list, state.History[i-1]) //add event right before incident
```
mjibson: Events should be assigned incident IDs so that this isn't needed.

cmd/bosun/web/static/js/incident.ts
```diff
@@ -0,0 +1,25 @@
+interface IIncidentScope extends ng.IScope {
+	error: string;
+	incident: any;
+	events: any;
+}
+bosunControllers.controller('IncidentCtrl', ['$scope', '$http', '$location', '$route', function($scope: IIncidentScope, $http: ng.IHttpService, $location: ng.ILocationService, $route: ng.route.IRouteService) {
+	var search = $location.search();
+	var id = search.id;
+	if (!id){
```
mjibson: space after `)`

cmd/bosun/web/static/js/incident.ts
```diff
@@ -0,0 +1,25 @@
...
+		$scope.error = "must supply incident id as query parameter"
+		return
+	}
+	$http.get('/api/incidents/events?id='+id)
+		.success((data) => {
+			$scope.incident = data.Incident;
+			console.log(moment($scope.incident.End).year());
```
mjibson: remove debug

cmd/bosun/web/static/js/incident.ts
```diff
@@ -0,0 +1,25 @@
...
+			if (moment($scope.incident.End).year() == 0){
```
mjibson: space after `)`

mjibson: The Go should be change so it sends null instead of the zero time.

cmd/bosun/web/static/partials/incident.html
```diff
@@ -0,0 +1,24 @@
...
+		<div class="alert alert-danger" ng-bind="error"></div>
+	</div>
+</div>
+<div ng-show="incident">
+	<div class="row">
+		<h3> Incident {{incident.Id}}: {{incident.AlertKey}} </h3>
+	</div>
+	<div class="row">
+		<h4 ng-show="incident.End"><span ts-time="incident.Start" ts-end-time="incident.End" no-link="true"></span></h4>
+		<h4 ng-hide="incident.End"><span ts-time="incident.Start" no-link="true" /> (Open)</h4>
```
mjibson: Why `no-link`? Linking to the event start time is useful. Or did I forget what that directive does?

cmd/bosun/web/web.go
```diff
@@ -346,6 +348,41 @@ func Alerts(t miniprofiler.Timer, w http.ResponseWriter, r *http.Request) (inter
...
+	incident, events, err := schedule.GetIncidentEvents(uint32(num))
+	if err != nil {
+		return nil, err
+	}
+	return struct {
+		Incident *sched.Incident
+		Events   []sched.Event
+	}{incident, events}, nil
+}
+func Incidents(t miniprofiler.Timer, w http.ResponseWriter, r *http.Request) (interface{}, error) {
```
mjibson: newline between funcs

cmd/bosun/web/web.go
```diff
@@ -346,6 +348,41 @@ func Alerts(t miniprofiler.Timer, w http.ResponseWriter, r *http.Request) (inter
 	return schedule.MarshalGroups(t, r.FormValue("filter"))
 }
 
+func IncidentEvents(t miniprofiler.Timer, w http.ResponseWriter, r *http.Request) (interface{}, error) {
+	id := r.FormValue("id")
+	if id == "" {
+		return nil, fmt.Errorf("id must be specified")
+	}
+	num, err := strconv.ParseUint(id, 10, 32)
```
mjibson: Remember to change the 32 to a 64 during the uint64 change.

cmd/bosun/web/web.go
```diff
@@ -346,6 +348,41 @@ func Alerts(t miniprofiler.Timer, w http.ResponseWriter, r *http.Request) (inter
...
+		return nil, err
+	}
+	return struct {
+		Incident *sched.Incident
+		Events   []sched.Event
+	}{incident, events}, nil
+}
+func Incidents(t miniprofiler.Timer, w http.ResponseWriter, r *http.Request) (interface{}, error) {
+	alert := r.FormValue("alert")
+	fromTime := time.Now().UTC().Add(-14 * 24 * time.Hour)
```
mjibson:
```
	toTime := time.Now().UTC()
	fromTime := toTime.Add(-14 * 24 * time.Hour) // 2 weeks
```

cmd/bosun/web/web.go
```diff
@@ -346,6 +348,41 @@ func Alerts(t miniprofiler.Timer, w http.ResponseWriter, r *http.Request) (inter
...
+	toTime := time.Now().UTC()
+	if from := r.FormValue("from"); from != "" {
+		if t, err := time.Parse(tsdbFormatSecs, from); err != nil {
+			fromTime = t
```
mjibson: return nil, err

cmd/bosun/web/web.go
```diff
@@ -346,6 +348,41 @@ func Alerts(t miniprofiler.Timer, w http.ResponseWriter, r *http.Request) (inter
...
+		}
+	}
+	if to := r.FormValue("to"); to != "" {
+		if t, err := time.Parse(tsdbFormatSecs, to); err != nil {
+			toTime = t
```
mjibson: return nil, err

cmd/bosun/web/web.go
mjibson: Do we need a `err == nil` case here, where we set `fromTime = t`?

cmd/bosun/web/web.go
mjibson: same as above with `err == nill`

mjibson:
This appears to also need a rebase as it can't be merged.

cmd/bosun/sched/sched.go
captncraig: The untention was to grab the events before and after the incident as well. After talking a bit, I don't think that is necessary anymore. Unless you think it is.

mjibson: I don't think it is either. Let's take it out.

cmd/bosun/sched/sched.go
captncraig: Because it is a field on the scedule, and we do `s := &Schedule{}` friggin everywhere.

mjibson: Look at https://github.com/bosun-monitor/bosun/blob/502e6253e846bd97b5ea1cb1998c4c7fa95fa776/cmd/bosun/sched/sched.go#L366

We already initialize a whole pile of maps there.

cmd/bosun/web/static/partials/incident.html
captncraig: It just links to timeanddate.com. I don't care for it, but will leave it in.

cmd/bosun/sched/check.go
```diff
@@ -111,7 +111,20 @@ func (s *Schedule) RunHistory(r *RunHistory) {
 		if event.Unevaluated {
 			continue
 		}
+		prev := state.Last()
 		state.Append(event)
+		if prev.IncidentId != 0 {
+			//if last event has incident id and is not closed, we continue it.
```
mjibson: comments have a space after `//`, and start with caps: https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

cmd/bosun/sched/check.go
```diff
@@ -111,7 +111,20 @@ func (s *Schedule) RunHistory(r *RunHistory) {
...
+			s.incidentLock.Lock()
+			if incident, ok := s.Incidents[prev.IncidentId]; ok && incident.End.IsZero() {
+				event.IncidentId = prev.IncidentId
+			}
+			s.incidentLock.Unlock()
+		} else if event.Status != StNormal {
+			//otherwise create new incident on first non-normal event
```
mjibson: comments have a space after `//`, start with caps, and end with `.`: https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

cmd/bosun/sched/check.go
mjibson: comments have spaces after `//`. If they aren't full sentences then they don't end with `.`

cmd/bosun/sched/check.go
mjibson: space after `//`

cmd/bosun/sched/sched.go
mjibson: This still needs to get moved.

cmd/bosun/sched/sched.go
```diff
@@ -704,6 +718,140 @@ func (a ActionType) MarshalJSON() ([]byte, error) {
...
+	if i[a].Start.Before(i[b].Start) {
+		return true
+	}
+	return string(i[a].AlertKey) < string(i[b].AlertKey)
+}
+func (i incidentList) Swap(a int, b int) { i[a], i[b] = i[b], i[a] }
+
+func (s *Schedule) createHistoricIncidents() {
+	incidents := incidentList{}
+	indexes := map[*Incident]int{}
```
mjibson: use make here and on the previous line

cmd/bosun/sched/sched.go
```diff
@@ -704,6 +718,140 @@ func (a ActionType) MarshalJSON() ([]byte, error) {
...
+}
+
+type incidentList []*Incident
+
+func (i incidentList) Len() int { return len(i) }
+func (i incidentList) Less(a int, b int) bool {
+	if i[a].Start.Before(i[b].Start) {
+		return true
+	}
+	return string(i[a].AlertKey) < string(i[b].AlertKey)
```
mjibson: no need to cast to string here

cmd/bosun/sched/sched.go
```diff
@@ -704,6 +718,140 @@ func (a ActionType) MarshalJSON() ([]byte, error) {
...
+	}
+
+	s.Incidents[id] = incident
+	return incident
+}
+
+type incidentList []*Incident
+
+func (i incidentList) Len() int { return len(i) }
+func (i incidentList) Less(a int, b int) bool {
```
mjibson: `a, b int`

cmd/bosun/sched/sched.go
```diff
@@ -704,6 +718,140 @@ func (a ActionType) MarshalJSON() ([]byte, error) {
...
+			}
+		}
+	}
+	// 2. Sort incidents
+	sort.Sort(incidents)
+	// 3. Assign ids and link events to appropriate ids
+	for _, incident := range incidents {
+		s.maxIncidentId++
+		incident.Id = s.maxIncidentId
+		//find events and mark them
```
mjibson: https://youtu.be/BVn1oQL9sWg?t=30s

cmd/bosun/web/static/js/incident.ts
```diff
@@ -0,0 +1,27 @@
...
+	var search = $location.search();
+	var id = search.id;
+	if (!id) {
+		$scope.error = "must supply incident id as query parameter"
+		return
+	}
+	$http.get('/api/incidents/events?id='+id)
+		.success((data) => {
+			$scope.incident = data.Incident;
+			if (moment($scope.incident.End).year() == 0) {
```
mjibson: The Go should be change so it sends null instead of the zero time.

cmd/bosun/web/web.go
```diff
@@ -346,6 +348,55 @@ func Alerts(t miniprofiler.Timer, w http.ResponseWriter, r *http.Request) (inter
...
+}
+
+func Incidents(t miniprofiler.Timer, w http.ResponseWriter, r *http.Request) (interface{}, error) {
+	alert := r.FormValue("alert")
+	toTime := time.Now().UTC()
+	fromTime := toTime.Add(-14 * 24 * time.Hour) // 2 weeks
+
+	if from := r.FormValue("from"); from != "" {
+		t, err := time.Parse(tsdbFormatSecs, from)
+		if err == nil {
```
mjibson: better error indent flow: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

cmd/bosun/web/web.go
```diff
@@ -346,6 +348,55 @@ func Alerts(t miniprofiler.Timer, w http.ResponseWriter, r *http.Request) (inter
...
+			fromTime = t
+		} else {
+			return nil, err
+		}
+	}
+	if to := r.FormValue("to"); to != "" {
+		t, err := time.Parse(tsdbFormatSecs, to)
+		if err == nil {
```
mjibson: better error indent flow

cmd/bosun/sched/sched.go
```diff
@@ -485,7 +490,7 @@ type State struct {
 	*Result
 
 	// Most recent last.
-	History      []Event  `json:",omitempty"`
+	History      []*Event `json:",omitempty"`
```
mjibson: Why did this change need to happen? The reason isn't obvious to me.

cmd/bosun/web/static/js/incident.ts
captncraig: I have three ideas for how to accomplish this and I hate all of them:

1. Create a new struct type just for web that has a string instead of a time.Time for End. Manually convert in those two calls.

2. Add an EndString or similar field to incident that we manually populate before serving. Serve that in json as `End`.

3. Create JSONTime struct as alias to time.Time that implements Marshaller interface: http://stackoverflow.com/a/23695774/121660 Use that in place of time.

mjibson: What about using a `*time.Time`?

cmd/bosun/sched/sched.go
```diff
@@ -607,18 +621,17 @@ func (s *State) Touch() {
...
-	if len(s.History) == 0 || s.Last().Status != event.Status {
-		event.Time = time.Now().UTC()
+	if len(s.History) == 0 || last.Status != event.Status {
 		s.History = append(s.History, *event)
 	}
 	return last.Status
 }
 
-func (s *State) Last() Event {
+func (s *State) Last() *Event {
```
mjibson: I can't see where we need this to be a `*Event` instead of an Event. I don't mean that there's some caller who requires a *Event, I mean I can't see where any caller needs a *Event for a good reason as opposed to a normal Event.

mjibson:
LGTM

PR #925, opened 2015-04-29
cmd/bosun: Add the top and filter functions

kylebrandt:
Top returns N specified results, and sort sorts based on the Value of a result of type number. This is useful if you want to limit the alert scope in this way, display less things in the gui, and for expressions that get rendered in notification templates.

It can be combined with filter to filter out series that meet the right hand condition. Basically a join that drops non matching groups and can work with series:

image


mjibson:
My initial reaction to both of these functions is that they are syntax solutions for a display problem. Can you come up with a rule that needs these functions? Yes, there are template expressions that might need this feature, but we already have a sort method on results that templates can use, and we could add slice function to perform top. For the graphs, we could add options to sort and limit.

kylebrandt:
I think it would be better to not have `top()` sort, and instead add `sort(typeNumber, "asc|desc")`

cmd/bosun/web/static/js/bosun.js
```diff
@@ -100,12 +100,18 @@ bosunControllers.controller('BosunCtrl', ['$scope', '$route', '$http', '$q', '$r
     $scope.panelClass = function (status, prefix) {
         if (prefix === void 0) { prefix = "panel-"; }
         switch (status) {
-            case "critical": return prefix + "danger";
-            case "unknown": return prefix + "info";
-            case "warning": return prefix + "warning";
-            case "normal": return prefix + "success";
-            case "error": return prefix + "danger";
-            default: return prefix + "default";
+            case "critical":
```
captncraig: I'm thinking you have an older verison of tsc installed maybe.

cmd/bosun/expr/expr.go
```diff
@@ -238,6 +240,10 @@ func (r ResultSlice) Len() int           { return len(r) }
 func (r ResultSlice) Swap(i, j int)      { r[i], r[j] = r[j], r[i] }
 func (r ResultSlice) Less(i, j int) bool { return r[i].Value.(Number) > r[j].Value.(Number) }
 
+func (r ResultSliceByGroup) Len() int           { return len(r) }
```
mjibson: return r.Len()

cmd/bosun/expr/expr.go
```diff
@@ -238,6 +240,10 @@ func (r ResultSlice) Len() int           { return len(r) }
 func (r ResultSlice) Swap(i, j int)      { r[i], r[j] = r[j], r[i] }
 func (r ResultSlice) Less(i, j int) bool { return r[i].Value.(Number) > r[j].Value.(Number) }
```
mjibson: We should probably remove these now.

cmd/bosun/expr/funcs.go
```diff
@@ -299,6 +317,41 @@ func NV(e *State, T miniprofiler.Timer, series *Results, v float64) (results *Re
...
+	case "asc":
+		sort.Stable(sort.Reverse(series.Results))
+	default:
+		return nil, fmt.Errorf("second argument of order() must be asc or desc")
+	}
+	return series, nil
+}
+
+func Limit(e *State, T miniprofiler.Timer, series *Results, v float64) (*Results, error) {
+	if len(series.Results) > int(v) {
```
mjibson: i := int(v), then use i throughout

cmd/bosun/expr/funcs.go
```diff
@@ -299,6 +317,41 @@ func NV(e *State, T miniprofiler.Timer, series *Results, v float64) (results *Re
...
+		series.Results = series.Results[0:int(v)]
```
mjibson: `Results[:i]` is sufficient (no need for the 0)

cmd/bosun/expr/funcs.go
```diff
@@ -299,6 +317,41 @@ func NV(e *State, T miniprofiler.Timer, series *Results, v float64) (results *Re
...
+		return series, nil
```
mjibson: remove this line since it's the same as below

cmd/bosun/expr/expr.go
kylebrandt: So you want me to make the ByValue type, get rid of this, and replace any sort calls on ResultSlice to cast to ByValue and then sort?

mjibson: I would prefer that. It's obvious what's being sorted on instead of you having to read the code or docs.

cmd/bosun/expr/expr.go
```diff
@@ -234,9 +238,13 @@ func (r ResultSlice) Filter(filter opentsdb.TagSet) ResultSlice {
 	return output
 }
 
-func (r ResultSlice) Len() int           { return len(r) }
-func (r ResultSlice) Swap(i, j int)      { r[i], r[j] = r[j], r[i] }
-func (r ResultSlice) Less(i, j int) bool { return r[i].Value.(Number) > r[j].Value.(Number) }
+func (r ResultSliceByValue) Len() int           { return len(r) }
+func (r ResultSliceByValue) Swap(i, j int)      { r[i], r[j] = r[j], r[i] }
+func (r ResultSliceByValue) Less(i, j int) bool { return r[i].Value.(Number) > r[j].Value.(Number) }
```
mjibson: This should be `<`. And DescByValue needs to use sort.Reverse.

mjibson:
LGTM

PR #926, opened 2015-04-29
cmd/bosun: use httpcontrol to retry bad HTTP requests

### LINKED ISSUE #426: Lots of "use of closed network connection errors"

mjibson:
This may fix #426


captncraig:
+1

PR #927, opened 2015-04-29
cmd/bosun: Remove computations from schedule execution

mjibson:
Still present and displayed on expression page.


mjibson:
Screenshot of new alert history table:

ss

kylebrandt:
Do those panels expand to a rendered template?

mjibson:
Nope. But we didn't have that before anyway. We could decide to add those in another PR, but they would only show for new alerts, not old ones.

kylebrandt:
Sounds okay to me

cmd/bosun/expr/expr.go
```diff
@@ -98,6 +99,8 @@ func (e *Expr) ExecuteState(s *State, T miniprofiler.Timer) (r *Results, queries
 	defer errRecover(&err)
 	if T == nil {
 		T = new(miniprofiler.Profile)
+	} else {
```
captncraig: this seems like a really odd condition for enabling computations. What is the logic there?

mjibson: Computations are now only used to show users during testing, never during the scheduled check. Miniprofiler is only present when there's a user sitting at the other side of the request that we will show query and timing stats to. Thus, this is a 100% accurate heuristic for determining if we would like to enable computations.

I initially had it as a separate parameter, but it was always the same as T == nil, so I used that instead so as to not change the calling function signature. Do you have another suggestion? I agree it's a bit weird, but I can't think of anything that would be more useful.

PR #929, opened 2015-04-29
cmd/bosun: Kill on double ctrl-c

mjibson:


kylebrandt:
Why?

mjibson:
While developing, I hit ctrl-c to kill the process. This triggers the save method. Sometimes I just want to kill it right now and abort the save because I have new code to test. When I'm testing against the production state file, it can take 10s of seconds to save. I don't want to wait for that, I just want it to die.

kylebrandt:
+1

PR #936, opened 2015-05-04
add more win.mem metrics and update various other metrics/metadata

gbrayut:
Currently running as a second instance on ny-web01 and ny-service01 pointed at http://devbosun.ds.stackexchange.com/items


cmd/scollector/collectors/mem_windows.go
```diff
@@ -23,21 +21,144 @@ func c_simple_mem_windows() (opentsdb.MultiDataPoint, error) {
...
 type Win32_OperatingSystem struct {
 	FreePhysicalMemory     uint64
 	FreeVirtualMemory      uint64
 	TotalVirtualMemorySize uint64
 	TotalVisibleMemorySize uint64
 }
+
+func c_windows_memory() (opentsdb.MultiDataPoint, error) {
+	var dst []Win32_PerfRawData_PerfOS_Memory
+	var q = wmi.CreateQuery(&dst, "")
```
captncraig: Is the empty query a standard thing with wmi? How does it know what to do?

gbrayut: the "" is for a filter. We only use it if we want to exclude something (like name <> _total). The actual query is created by the wmi package by looking at the dst object.

PR #940, opened 2015-05-07
WIP - Testing tsc in go generate

captncraig:


cmd/bosun/generate.go
```diff
@@ -0,0 +1,4 @@
+package main
+
+//go:generate sh tsc.sh
```
mjibson: Doesn't work on Windows. I'd prefer this to not be the case.

gbrayut:
LGTM

cmd/bosun/main.go
```diff
@@ -155,16 +154,9 @@ func main() {
 	}()
 	if *flagWatch {
 		watch(".", "*.go", quit)
-		watch(filepath.Join("web", "static", "templates"), "*.html", quit)
+		watch(filepath.Join("web", "static", "templates"), "*.html", web.RunEsc)
```
mjibson: Why did this change? It quits when index.html changes, which needs to be done because it gets loaded only once on start.

cmd/bosun/tsc.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
```
mjibson: I think this file can go away now.

cmd/bosun/web/web.go
```diff
@@ -55,23 +55,35 @@ func init() {
...
-	if err != nil {
-		log.Fatal(err)
-	}
-	b, err := ioutil.ReadAll(index)
-	if err != nil {
-		log.Fatal(err)
+	webFS := FS(devMode)
+
+	var parseTemplates = func() *template.Template {
+		index, err := webFS.Open("/templates/index.html")
```
mjibson: Try out the `FSMustString` method instead. Should simplify most of this.

mjibson:
LGTM after my one comment.

PR #941, opened 2015-05-08
cmd/scollector: use new sys windows service package

mjibson:
The winsvc package is going away and has been merged into the official go sys subrepo. The API changed a bit, otherwise things work correctly. I tested service create/remove/start/stop on my machine and it worked correctly. We should perhaps do a test deploy on a real system before merging.


captncraig:
+1

gbrayut:
I will deploy this to a few test systems this afternoon

gbrayut:
Uninstall/Reinstall looks good on a few test systems, and all the event log stuff still works too. I'm going to let it run overnight but feel free to merge whenever. I'll do another test deploy to a larger set when it hits master.

PR #944, opened 2015-05-08
cmd/bosun: bosun -> Bosun

### LINKED ISSUE #906: Need to be consistant with casing of Bosun
On the bosun.org site we use Bosun

image

But the image from the Bosun server is lowercased:

image

I think the correct usage is Bosun with a capital B, as we have used it this way in metadata descriptions.

mjibson:
fixes #906


captncraig:
+1 Probably needs regenerate to merge properly.

PR #947, opened 2015-05-08
cmd/bosun: always use State instead of current time

### LINKED ISSUE #789: Bosun Since() does not work as expected when testing rule or expression on historical data
The Since() operator appears to always use the current time for comparison. When testing an expression or rule on historical data I would expect it to use that date (or better that date + the duration in the query) as the time for comparison.

I talked about this with Kyle, and he said to file a bug. He also said the epoch function may have a similar issue.

mjibson:
Previously epoch, since, and forcastlr would use time.Now.

fixes #789


captncraig:
+1

PR #948, opened 2015-05-12
cmd/bosun: clear body and subject on unknown

mjibson:


cmd/bosun/sched/check.go
```diff
@@ -129,6 +129,11 @@ func (s *Schedule) RunHistory(r *RunHistory) {
 		a := s.Conf.Alerts[ak.Name()]
 		wasOpen := state.Open
 		if event.Status > StNormal {
+			state.Subject = ""
```
captncraig: should this be `state.ClearTemplateData()` or something?

mjibson: I'd do that if you feel strongly. But it's only 5 lines and just needed in one place, so I don't think it needs to be a function.

captncraig: fair enough

captncraig:
LGTM

PR #949, opened 2015-05-12
cmd/bosun: add a default unknownTemplate

### LINKED ISSUE #655: Bosun should have a default unknown Template
I'm not sure where to look to help debug the issue, but from time to time I get "phantom" email notifications. The subject is empty and the body is empty. 

I currently have only one alert setup. See my config below. 

The interesting note is that they seem to be spaced by about one hour. One at 5:46AM, 6:38AM, 7:36AM, 8:38AM. Then again maybe not so interesting as that is my alert interval. =)

```
tsdbHost = localhost:4242
smtpHost = <removed>
smtpUsername = <removed>
smtpPassword = <removed>
emailFrom = <removed>

template test {
	subject = {{.Last.Status}}: {{.Alert.Name}} on {{.Group.host}}
	body = `<p>Alert: {{.Alert.Name}} triggered on {{.Group.host}}
	<hr>
	<p><strong>Computation</strong>
	<table>
		{{range .Computations}}
			<tr><td><a href="{{$.Expr .Text}}">{{.Text}}</a></td><td>{{.Value}}</td></tr>
		{{end}}
	</table>
	<hr>
	{{ .Graph .Alert.Vars.metric }}
	<hr>
	<p><strong>Relevant Tags</strong>
	<table>
		{{range $k, $v := .Group}}
			<tr><td>{{$k}}</td><td>{{$v}}</td></tr>
		{{end}}
	</table>`
}

notification default {
	email = <removed>
	next = default
	timeout = 1h
}

alert api.high.bandwidth {
	template = test
	critNotification = default	
	warnNotification = default		
	$metric = q("sum:rate{counter,,1}:os.net.bytes{host=saddleback-api-*}", "1h", "")
	$max = max($metric)
	crit = $max > 5000000
	warn = $max > 2000000
}
```

mjibson:
fixes #655

### DIFF
--- a/cmd/bosun/sched/check_test.go
+++ b/cmd/bosun/sched/check_test.go
@@ -258,6 +258,68 @@ Loop:
 	}
 }
 
+// TestCheckNotifyUnknownDefault tests the default unknownTemplate.
+func TestCheckNotifyUnknownDefault(t *testing.T) {
+	s := new(Schedule)
+	nc := make(chan string, 1)
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		b, _ := ioutil.ReadAll(r.Body)
+		nc <- string(b)
+	}))
+	defer ts.Close()
+	u, err := url.Parse(ts.URL)
+	if err != nil {
+		t.Fatal(err)
+	}
+	c, err := conf.New("", fmt.Sprintf(`
+		template t {
+			subject = template
+		}
+		notification n {
+			post = http://%s/
+		}
+		alert a {
+			template = t
+			critNotification = n
+			crit = 1
+		}
+	`, u.Host))
+	if err != nil {
+		t.Fatal(err)
+	}
+	c.StateFile = ""
+	err = s.Init(c)
+	if err != nil {
+		t.Fatal(err)
+	}
+	r := &RunHistory{
+		Events: map[expr.AlertKey]*Event{
+			expr.NewAlertKey("a", opentsdb.TagSet{"h": "x"}): {Status: StUnknown},
+			expr.NewAlertKey("a", opentsdb.TagSet{"h": "y"}): {Status: StUnknown},
+		},
+	}
+	s.RunHistory(r)
+	s.CheckNotifications()
+	gotExpected := false
+Loop:
+	for {
+		select {
+		case r := <-nc:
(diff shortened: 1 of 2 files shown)


cmd/bosun/sched/check_test.go
```diff
@@ -258,6 +258,68 @@ Loop:
 	}
 }
 
+// TestCheckNotifyUnknownDefault tests the default unknownTemplate.
+func TestCheckNotifyUnknownDefault(t *testing.T) {
+	s := new(Schedule)
+	nc := make(chan string, 1)
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
captncraig: do we really need to test the full post functionality to make sure this works? What about the approach used at https://github.com/bosun-monitor/bosun/blob/default-unknown-template/cmd/bosun/sched/check_test.go#L45 ?

cmd/bosun/sched/check_test.go
```diff
@@ -258,6 +258,68 @@ Loop:
...
+			expr.NewAlertKey("a", opentsdb.TagSet{"h": "y"}): {Status: StUnknown},
+		},
+	}
+	s.RunHistory(r)
+	s.CheckNotifications()
+	gotExpected := false
+Loop:
+	for {
+		select {
+		case r := <-nc:
```
captncraig: This looks deadlocky to me. Does the server complete the request before it sends to the channel? Otherwise it seems it is waiting for the client to move on to the receive before the send can complete. The whole thing seems a bit fragile, and the timeout feels odd.

cmd/bosun/sched/check_test.go
mjibson: Yes, I think that is sufficient. Will fix.

cmd/bosun/sched/check_test.go
mjibson: This code is duplicated exactly like this in a few places. It is this way because s.CheckNotifications isn't synchronous. So, we start the receiving server above, send the notifications, then try to see which ones we can get in the next 1 second. I think using the non-server based notification check this can all be removed. Will fix.

cmd/bosun/sched/check_test.go
mjibson: Ah, no, that won't work in this case because unknown notifications are computed separately and don't exist in a usable form by tests.

PR #952, opened 2015-05-12
Actually adding historic incidents to global map.

captncraig:
This needs to be added to fix historic incidents after merging previous pr.

### DIFF
--- a/cmd/bosun/sched/check_test.go
+++ b/cmd/bosun/sched/check_test.go
@@ -148,6 +148,55 @@ func TestCheckSilence(t *testing.T) {
 	}
 }
 
+func TestIncidentIds(t *testing.T) {
+	s := new(Schedule)
+	c, err := conf.New("", `
+		alert a {
+			crit = 1
+		}
+	`)
+	if err != nil {
+		t.Fatal(err)
+	}
+	c.StateFile = ""
+	s.Init(c)
+	ak := expr.NewAlertKey("a", nil)
+	r := &RunHistory{
+		Events: map[expr.AlertKey]*Event{
+			ak: {Status: StWarning},
+		},
+	}
+	expect := func(id uint64) {
+		if s.status[ak].Last().IncidentId != id {
+			t.Fatalf("Expeted incident id %d. Got %d.", id, s.status[ak].Last().IncidentId)
+		}
+	}
+	s.RunHistory(r)
+	expect(1)
+
+	r.Events[ak].Status = StNormal
+	r.Events[ak].IncidentId = 0
+	s.RunHistory(r)
+	expect(1)
+
+	r.Events[ak].Status = StWarning
+	r.Events[ak].IncidentId = 0
+	s.RunHistory(r)
+	expect(1)
+
+	r.Events[ak].Status = StNormal
+	r.Events[ak].IncidentId = 0
+	s.RunHistory(r)
+	err = s.Action("", "", ActionClose, ak)
+	if err != nil {
+		t.Fatal(err)
+	}
+	r.Events[ak].Status = StWarning
+	r.Events[ak].IncidentId = 0
+	s.RunHistory(r)
+	expect(2)
+}
+
 func TestCheckNotify(t *testing.T) {
 	s := new(Schedule)
 	nc := make(chan string)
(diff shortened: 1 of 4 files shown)


captncraig:
The situation we have now is that incidents were "created" for 50k or so historic incidents, but not added to the permanent data structure. In order to fix this I propose we 

1. Stop bosun.
2. Pull the state file locally and run the following program to delete all incident data from our state file:
```
package main

import (
	"bosun.org/_third_party/github.com/boltdb/bolt"
	"fmt"
)

func main() {
	db, err := bolt.Open("bosun.state", 0600, nil)
	if err != nil {
		log.Fatal(err)
	}
	defer db.Close()
	err := db.Update(func(tx *bolt.Tx) error {
		bucket := tx.Bucket([]byte("bindata"))
		return bucket.Delete([]byte("incidents"))
	})
	fmt.Println(err)
}
```
3. Restart bosun with the new build.

mjibson:
Another possible fix: add some code that always runs createInitialIncidents. We deploy it once, then can remove it. This would not require running a special program with the bosun state file.

cmd/bosun/sched/check_test.go
```diff
@@ -148,6 +148,67 @@ func TestCheckSilence(t *testing.T) {
...
+	expect(1)
+
+	r.Events[ak].Status = StWarning
+	r.Events[ak].IncidentId = 0
+	s.RunHistory(r)
+	expect(1)
+	r.Events[ak].Status = StNormal
+	r.Events[ak].IncidentId = 0
+	s.RunHistory(r)
+	err = s.Action("joe", "abc", ActionClose, ak)
```
mjibson: these strings can both be `""`

cmd/bosun/sched/check_test.go
```diff
@@ -148,6 +148,67 @@ func TestCheckSilence(t *testing.T) {
...
+
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	r.Events[ak].Status = StWarning
+	r.Events[ak].IncidentId = 0
+	s.RunHistory(r)
+	expect(2)
+
```
mjibson: remove empty line

cmd/bosun/sched/check_test.go
```diff
@@ -148,6 +148,67 @@ func TestCheckSilence(t *testing.T) {
...
+	r.Events[ak].Status = StNormal
+	r.Events[ak].IncidentId = 0
+	s.RunHistory(r)
+	expect(1)
+
+	r.Events[ak].Status = StWarning
+	r.Events[ak].IncidentId = 0
+	s.RunHistory(r)
+	expect(1)
```
mjibson: newline after this to be consistent with other expect calls

cmd/bosun/sched/check_test.go
```diff
@@ -148,6 +148,67 @@ func TestCheckSilence(t *testing.T) {
...
+	r.Events[ak].Status = StNormal
+	r.Events[ak].IncidentId = 0
+	s.RunHistory(r)
+	err = s.Action("joe", "abc", ActionClose, ak)
+
```
mjibson: no newline here since the err was assigned just on the line before

cmd/bosun/sched/check_test.go
```diff
@@ -148,6 +148,67 @@ func TestCheckSilence(t *testing.T) {
 	}
 }
 
+func TestIncidentIdsABC(t *testing.T) {
+	s := new(Schedule)
+	c, err := conf.New("", `
+		template t {
```
mjibson: i don't see a need for the template or notification here

cmd/bosun/sched/check_test.go
```diff
@@ -148,6 +148,67 @@ func TestCheckSilence(t *testing.T) {
...
+			subject = 1
+		}
+		notification n {
+			print = true
+		}
+		alert a {
+			warnNotification = n
+			warn = 1
```
mjibson: the warn can also go

mjibson:
To explain more what I was thinking above: we always run createHistoricIncidents, and have it always reassign all incident IDs. This will get them in the right order. Yes, it'll change some, but that's fine since this is so new.

PR #965, opened 2015-05-14
Providing quick jump between alert and its template.

### LINKED ISSUE #964: Bosun Rule Editor: Ability to quickly switch between template and alert definitions
I end up switching between the template and alert definitions a lot when working on alerts. Previously they were in separate editors, so it was easy, but now it is harder due to having to locate it in the drop down. Two approaches to making this easier:

1. Add a second instance of the editor to the rule page (or potentially more than 2 as needed)
2. Add a toggle button in the "jump to" row that lets you switch between the template and the alert definitions with a single click.

Option 2 shouldn't be that hard, since we already track the "active alert" that will be used when pressing the Test button. Getting the template information should be a simple regex and the toggle button can then let you bounce between the two.

captncraig:
Fixes #964 

screen shot 2015-05-14 at 8 49 45 am


cmd/bosun/web/static/js/config.ts
```diff
@@ -223,6 +225,33 @@ bosunControllers.controller('ConfigCtrl', ['$scope', '$http', '$location', '$rou
...
+						$scope.quickJumpTarget = "template " + found[1];
+					}
+				}
+			}
+		}
+	}
+	
+	$scope.quickJump = () =>{
+		var parts = $scope.quickJumpTarget.split(" ");
+		if (parts.length != 2){return;}
```
mjibson: when well it ever not be 2?

cmd/bosun/web/static/js/config.ts
```diff
@@ -223,6 +225,33 @@ bosunControllers.controller('ConfigCtrl', ['$scope', '$http', '$location', '$rou
...
+		$scope.scrollTo(parts[0],parts[1]);
+		if(parts[0] == "template" && $scope.selected_alert){
```
mjibson: when will parts[0] ever not be "template"?

cmd/bosun/web/static/js/config.ts
captncraig: It will be empty string if there is no template for the alert. But the button should be disabled. It is mostly a protection against this getting called wrong, which should never happen.

cmd/bosun/web/static/js/config.ts
mjibson: is $scope.selected_alert ever not set here?

captncraig: quickJumpTarget will alternate between "template foo" and "alert x".

captncraig: It could potentially be if there are no alerts in the config, or if the thing has just loaded up. I kinda want to leave the check in just in case. 

We should also clear "quickJumpTarget" at the beginning of this though, just in case a new valid target is not set.

mjibson:
LGTM

PR #969, opened 2015-05-15
Providing endpoint to backup state file.

captncraig:
Because bosun maintains a file lock on the state file, a simple solution for backups is for bosun itself to be able to backup its own state file.

A GET request to /api/backup will stream the state file over the web request.

`curl bosunserver/api/backup > mybackup-timestamp.state` is all that is needed.


mjibson:
What's the problem this is solving?

mjibson:
LGTM

PR #971, opened 2015-05-18
Short circuiting || and && when NaNs are present.

captncraig:


gbrayut:
This logic needs to be documented in the http://bosun.org/expressions.html#operators section as well.

PR #972, opened 2015-05-18
Short circuiting || and && when NaN or Inf present

captncraig:


cmd/bosun/expr/expr_test.go
```diff
@@ -33,6 +34,10 @@ func TestExprSimple(t *testing.T) {
 		{"1>=2", 0},
 		{"-1 > 0", 0},
 		{"-1 < 0", 1},
+		{"1 || 1 / 0", 1},
+		{"0 || 1 / 0", Scalar(math.Inf(1))},
```
mjibson: `1 / 0` is Inf, not NaN, so this won't work for the NaN case because NaN has a special case lookup in 2 places.

mjibson:
See my solution to this that handles NaNs: https://github.com/bosun-monitor/bosun/pull/973

PR #973, opened 2015-05-18
Short circuit || and &&

mjibson:
This also corrects the handling of NaN so it can short circuit.

### DIFF
--- a/cmd/bosun/expr/expr.go
+++ b/cmd/bosun/expr/expr.go
@@ -392,65 +392,59 @@ func (e *State) walkBinary(node *parse.BinaryNode, T miniprofiler.Timer) *Result
 				Group:        v.Group,
 				Computations: v.Computations,
 			}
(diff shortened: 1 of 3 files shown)


captncraig:
lgtm

cmd/bosun/expr/expr.go
```diff
@@ -460,6 +454,17 @@ func (e *State) walkBinary(node *parse.BinaryNode, T miniprofiler.Timer) *Result
 }
 
 func operate(op string, a, b float64) (r float64) {
+	// Test short circuit before NaN.
+	switch op {
+	case "||":
+		if a != 0 {
```
captncraig: Will this make `NaN || 1` resolve to 1? Is that what we want?

mjibson: I think it is, because we want `1 || NaN` to also resolve to 1, and `||` should be symmetric.

PR #986, opened 2015-05-22
Function to make incident link available to templates

captncraig:
`{{.Incident}}` should give link to incident on correct host. `{{.State.Last.IncidentId}}` should give just id.


mjibson:
LGTM

PR #988, opened 2015-05-26
Release versions

captncraig:
1. Official release numbers will be `major.minor.count`. Before any official release, the Version in `/version/version.go` will need to be incremented.
2. `release.sh` builds all cross-compiled binaries and uses `githubRelease.go` to create a draft release on github. All merged PR titles since the last release should be the draft body. This can be manually edited before publishing.
3. Code parsing / replacing versions has been replaced in scripts by using `-ldflags -X`appropriately.

If somebody uses `go get bosun.org/cmd/bosun` to fetch they will only get the hard-coded version number of the last release (`0.2.3-dev`).

If you use `build/build.go` you will have a build timestamp and sha baked in, still be a `-dev` build.

Only official builds will have the `-dev` designation removed.

I have not adjusted our internal dev/prod builds to accomidate these changes.


cmd/bosun/main.go
```diff
@@ -66,7 +66,19 @@ var (
...
 	flag.Parse()
 	if *flagVersion {
-		fmt.Printf("bosun version %v (%v)\n", VersionDate, VersionID)
+		if Version == "" {
+			Version = "dev"
+		}
+		if VersionSHA == "" {
+			VersionSHA = "???"
+			// Try to pull git revision if we can. If the working directory is not in the bosun repo, or they have local changes, all bets are off.
+			cmd := exec.Command("git", "rev-parse", "HEAD")
```
mjibson: I don't think we should ever do this. Just leave it empty in this case.

cmd/bosun/main.go
```diff
@@ -66,7 +66,19 @@ var (
 func main() {
 	flag.Parse()
 	if *flagVersion {
-		fmt.Printf("bosun version %v (%v)\n", VersionDate, VersionID)
+		if Version == "" {
+			Version = "dev"
```
mjibson: I thought we discussed having a human hard code the version?

build/release/githubRelease.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	var lastRelease github.RepositoryRelease
+	for _, r := range releases {
+		if r.PublishedAt != nil {
+			lastRelease = r
+			break
+		}
+	}
+
+	fmt.Println("Getting all prs since last release to build release notes...")
```
mjibson: Why not get all commits? It is possible we just committed something to master without a PR.

captncraig: Don't do that.

captncraig: The idea is that pr titles are more likely to be feature descriptions than commit messages

build/release/githubRelease.go
```diff
@@ -0,0 +1,132 @@
...
+	release.Name = &number
+	isDraft := true
+	release.Draft = &isDraft
+	release.Body = &body
+	release, _, err = client.Repositories.CreateRelease("bosun-monitor", "bosun", release)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	fmt.Println("Uploading artifacts...")
```
mjibson: Probably worth printing a link to the draft release here.

version/version.go
```diff
@@ -0,0 +1,31 @@
+package version
```
mjibson: Need doc string

build/build.sh
```diff
@@ -0,0 +1,4 @@
+SHA=`git rev-parse HEAD`
```
mjibson: Can we remove this file since we have build.go?

PR #989, opened 2015-05-26
cmd/scollector: move to TOML conf file

mjibson:
Includes -totoml converter.

### DIFF
--- a/cmd/scollector/main.go
+++ b/cmd/scollector/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bytes"
 	"encoding/json"
 	"flag"
 	"fmt"
@@ -14,6 +15,7 @@ import (
 	"strings"
 	"time"
 
+	"bosun.org/_third_party/github.com/BurntSushi/toml"
 	"bosun.org/cmd/scollector/collectors"
 	"bosun.org/collect"
 	"bosun.org/metadata"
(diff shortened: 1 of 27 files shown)


cmd/scollector/main.go
```diff
@@ -34,39 +36,91 @@ var (
...
 	mains []func()
 )
 
-func readConf() {
+type Conf struct {
+	// Host is the OpenTSDB or Bosun host to send data.
+	Host string
+	// FullHost enables full hostnames: doesn't truncate to first ".".
+	FullHost bool
+	ColDir   string
```
captncraig: Need documentation on everything here. Whatever docs were on the flag should be ok.

cmd/scollector/main.go
```diff
@@ -34,39 +36,91 @@ var (
...
-	flagFreq            = flag.String("freq", "15", "Set the default frequency in seconds for most collectors.")
 	flagConf            = flag.String("conf", "", "Location of configuration file. Defaults to scollector.conf in directory of the scollector executable.")
-	flagAWS             = flag.String("aws", "", `AWS keys and region, format: "access_key:secret_key@region".`)
+	flagToToml          = flag.String("totoml", "", "Location of destination toml file to convert. Reads from value of -conf.")
 
 	mains []func()
 )
 
-func readConf() {
+type Conf struct {
```
captncraig: perhaps we need a seperate file for conf stuff. Maybe even a seperate package so collectors could reference it directly?

cmd/scollector/main.go
```diff
@@ -159,81 +149,63 @@ func main() {
...
-			collectors.SNMPCisco(sp[0], sp[1])
+	for _, h := range conf.HAProxy {
+		for _, i := range h.Instances {
+			collectors.HAProxy(h.User, h.Password, i.Tier, i.URL)
 		}
 	}
-	if *flagICMP != "" {
-		for _, s := range strings.Split(*flagICMP, ",") {
-			collectors.ICMP(s)
+	for _, s := range conf.SNMP {
```
captncraig: Can individual collectors do this kind of configuring on their own on init? Main would be cleaner if fewer collector-specific things where there?

mjibson: Agreed. Will move.

mjibson:
Added doc strings to most stuff of the Conf type. Moved collector init checks into collector logic.

cmd/scollector/main.go
```diff
@@ -358,3 +327,176 @@ func printPut(c chan *opentsdb.DataPoint) {
 		slog.Info(string(b))
 	}
 }
+
+func toToml(fname string) {
```
captncraig: Is there an end of life plan for this? When can we delete this code? I assume new config directives won't need to be updated here?

mjibson: I'd say leave it in for a few weeks or until we add a new config directive, then take it out. Not intended for new configs or long term work.

captncraig:
LGTM

PR #992, opened 2015-05-27
cmd/bosun: Move miniprofiler enable to a header

### LINKED ISSUE #991: Bosun auto-complete for tag values stopped working
When I try graphing linux.cpu.percpu and entering the letter i in the typetag I would expect a list of values but don't get anything. Also no errors are displayed in the console and I don't even see an ajax request (which I do see when changing the metric) 

Not sure when this stopped working, but devbosun works as expected and is running docker image 656f5d37ee36 created 5 weeks ago version 20150415181718 (bbc223c3e55ab554ee4d6aa2fde578af6d5f0d72)

Testing using my personal docker instance it appears to work in a 3 week old bosun version 20150428222236 (8e4440108b9154fa20f485469a31f0c19e7607d2) but is broken in the the latest docker release from 11 days ago version 20150515173229 (9adbcee922cb30c89d7158ba033e05862d47c79e) so the issue must have appeared sometime between those releases.

mjibson:
Caused because some routes treated URL parameters specially, so we can't
use "&miniprofiler=true" everywhere.

fixes #991


captncraig:
Looks like you missed a gofmt in web.go

captncraig:
LGTM

PR #993, opened 2015-05-27
Allow alerts to run on different schedules.

### LINKED ISSUE #976: Allow alerts to run at different frequencies
This proposal is to add a `runEvery` option to alerts, and a global default `runEvery` parameter.

If an alert specified `runEvery = 3`, it will only be evaluated every `3 * checkFrequency`. By default the global value will be `1` unless otherwise specified.

## Possible gotchas: ##

* Unknown checks will need to be aware of the modified checkFrequency for each alert.
* Dependencies will need to be evaluated based on the last known state, not necessarily from the same run.

captncraig:
Adds a few config options:

1. Alert level `runEvery` which determines how many intervals between checks on this alert.
2. Global `defaultRunEvery` to determine the default modulo.

With a `defaultRunEvery` of 5 and a `confFrequency = 1m`, each alert will run every 5m unless explicitely set to `runEvery = 1` to run every minute.

Fixes #976

### DIFF
--- a/cmd/bosun/conf/conf.go
+++ b/cmd/bosun/conf/conf.go
@@ -29,6 +29,7 @@ type Conf struct {
 	Vars
 	Name             string        // Config file name
 	CheckFrequency   time.Duration // Time between alert checks: 5m
+	DefaultRunEvery  int           // Default number of check intervals to run each alert: 1
 	HTTPListen       string        // Web server listen address: :80
 	Hostname         string
 	RelayListen      string // OpenTSDB relay listen address: :4242
@@ -246,6 +247,7 @@ type Alert struct {
 	IgnoreUnknown    bool
 	UnjoinedOK       bool `json:",omitempty"`
 	Log              bool
+	RunEvery         int
 	returnType       eparse.FuncType
 
 	template string
@@ -345,6 +347,7 @@ func New(name, text string) (c *Conf, err error) {
 	c = &Conf{
 		Name:             name,
 		CheckFrequency:   time.Minute * 5,
+		DefaultRunEvery:  1,
 		HTTPListen:       ":8070",
 		StateFile:        "bosun.state",
 		PingDuration:     time.Hour * 24,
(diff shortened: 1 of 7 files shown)


cmd/bosun/conf/conf.go
```diff
@@ -470,6 +473,15 @@ func (c *Conf) loadGlobal(p *parse.PairNode) {
 			c.errorf("responseLimit must be > 0")
 		}
 		c.ResponseLimit = i
+	case "defaultRunEvery":
+		i, err := strconv.ParseInt(v, 10, 32)
```
mjibson: `strconv.Atoi` returns an int if that'd be easier, even though it's different. Whatever.

cmd/bosun/conf/conf.go
```diff
@@ -1326,3 +1344,11 @@ func (c *Conf) MakeLink(path string, v *url.Values) string {
 	}
 	return u.String()
 }
+
+func (c *Conf) RunModulo(alert *Alert) uint64 {
+	modulo := alert.RunEvery
+	if modulo == 0 {
```
mjibson: How about just run this at the end of `loadAlert` so that it only needs to be computed once?

mjibson: Does it make sense to do this calculation at the end of `loadAlert` once?

cmd/bosun/sched/check.go
```diff
@@ -322,22 +324,38 @@ func (s *Schedule) CollectStates() {
...
 			continue
 		}
+		anyFound = true
 		if ev.Status == StUnknown {
 			unknown = append(unknown, ak)
 		} else if ev.Unevaluated {
 			uneval = append(uneval, ak)
 		}
 	}
+	if !anyFound {
```
mjibson: I don't understand this logic. It first checks to see if there were any events in the run history for the given alertkey. If there were not, then it checks the global state. Why is it like this? Hangout is perhaps best to explain.

mjibson:
Why is it harder to start N go routines, with all alerts of the same frequency in the same go routine?

cmd/bosun/sched/check.go
captncraig: That is correct. This allows dependencies to work even if the dependent-upon alert has not been run this interval. If the required alert was not run this interval, we simply look at the current state from the last run. We can discuss tomorrow.

mjibson:
TODO: move the channel locking into web.go

mjibson:
LGTM if this is how we want to do it. But, I do not like the intervals being exposed to users: I think users should be able to specify times instead. If we want to implement it with intervals on the backend, we can compute them.

However I think using goroutines grouped by time makes the most sense. We could also then do things like make the check intervals run at predictable times: a "5m" interval would always run at 12:00, 12:05, etc.

Thoughts @kylebrandt? If you like how this PR works then merge away, though.

PR #994, opened 2015-05-27
metadata: Make time a pointer

mjibson:
This is needed so that go programs using the Metasend type can send a nil
time.


captncraig:
LGTM

PR #1000, opened 2015-05-28
Documentation fetches latest release dynamically

captncraig:
The download pages will now get the latest release info from github via ajax request. 

If the ajax fails, the download links will simply take the user to the github latest release page, so they can still access it.

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -1,3 +1,0 @@
-version:
-  hash: "9adbcee922cb30c89d7158ba033e05862d47c79e"
-  id: 20150515173231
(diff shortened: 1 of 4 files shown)


mjibson:
GitHub Pages says they provide this data right in the jekyll metadata now, so probably no need for an AJAX request:

https://github.com/blog/1996-releases-metadata-for-github-pages

captncraig:
That would probably work if we were hosting from the `gh-pages` branch of the main bosun repo. But we are currently hosting from the seperate repo.

mjibson:
Well let's move to gh-pages then?

mjibson:
Here's a link to a page that uses the github releases data: https://github.com/mjibson/mog/blob/gh-pages/index.html

PR #1003, opened 2015-05-28
cmd/bosun: png/svg graph style changes

kylebrandt:
Before:

warning__os_cpu__on_skippygohungles_-_kyle_stackoverflow_com_-_stack_exchange_mail

After:

warning__os_cpu__on_skippygohungles_-_kyle_stackoverflow_com_-_stack_exchange_mail

The new argument to Graph is a list of units for the axis. It is vargs so this isn't breaking change (you can omit the second argument). The "title" has been moved to notes at the bottom of the graph, and moved outside of the svg/png. This means they can be copied and pasted when it comes in an email, and there is no issue with text being cut off.

### DIFF
--- a/cmd/bosun/sched/chart.go
+++ b/cmd/bosun/sched/chart.go
@@ -1,12 +1,10 @@
 package sched
 
 import (
-	"fmt"
 	"image"
 	"image/color"
 	"image/png"
 	"io"
-	"time"
 
 	"bosun.org/_third_party/github.com/MiniProfiler/go/miniprofiler"
 	"bosun.org/_third_party/github.com/ajstarks/svgo"
@@ -18,7 +16,6 @@ import (
 )
 
 var chartColors = []color.Color{
-	color.NRGBA{0xe4, 0x1a, 0x1c, 0xff},
 	color.NRGBA{0x37, 0x7e, 0xb8, 0xff},
 	color.NRGBA{0x4d, 0xaf, 0x4a, 0xff},
 	color.NRGBA{0x98, 0x4e, 0xa3, 0xff},
@@ -26,24 +23,27 @@ var chartColors = []color.Color{
 	color.NRGBA{0xa6, 0x56, 0x28, 0xff},
 	color.NRGBA{0xf7, 0x81, 0xbf, 0xff},
 	color.NRGBA{0x99, 0x99, 0x99, 0xff},
+	color.NRGBA{0xe4, 0x1a, 0x1c, 0xff},
 }
 
 // Autostyle styles a chart series.
 func Autostyle(i int) chart.Style {
 	c := chartColors[i%len(chartColors)]
+	s := chart.Symbol[i%len(chart.Symbol)]
 	return chart.Style{
-		// 0 uses a default
-		SymbolSize: 0.00001,
-		LineStyle:  chart.SolidLine,
-		LineWidth:  1,
-		LineColor:  c,
+		Symbol:      s,
+		SymbolSize:  .5,
+		SymbolColor: c,
+		LineStyle:   chart.SolidLine,
+		LineWidth:   1,
+		LineColor:   c,
 	}
 }
 
 var white = color.RGBA{0xff, 0xff, 0xff, 0xff}
 
-func (s *Schedule) ExprSVG(t miniprofiler.Timer, w io.Writer, width, height int, res []*expr.Result, q string, now time.Time) error {
-	ch, err := s.ExprGraph(t, res, q, now)
+func (s *Schedule) ExprSVG(t miniprofiler.Timer, w io.Writer, width, height int, unit string, res []*expr.Result) error {
+	ch, err := s.ExprGraph(t, unit, res)
 	if err != nil {
 		return err
 	}
@@ -52,12 +52,11 @@ func (s *Schedule) ExprSVG(t miniprofiler.Timer, w io.Writer, width, height int,
 	g.Rect(0, 0, width, height, "fill: #ffffff")
 	sgr := svgg.AddTo(g, 0, 0, width, height, "", 12, white)
 	ch.Plot(sgr)
-	g.End()
 	return nil
 }
 
-func (s *Schedule) ExprPNG(t miniprofiler.Timer, w io.Writer, width, height int, res []*expr.Result, q string, now time.Time) error {
-	ch, err := s.ExprGraph(t, res, q, now)
+func (s *Schedule) ExprPNG(t miniprofiler.Timer, w io.Writer, width, height int, unit string, res []*expr.Result) error {
+	ch, err := s.ExprGraph(t, unit, res)
 	if err != nil {
 		return err
 	}
@@ -67,10 +66,10 @@ func (s *Schedule) ExprPNG(t miniprofiler.Timer, w io.Writer, width, height int,
 	return png.Encode(w, g)
 }
 
-func (s *Schedule) ExprGraph(t miniprofiler.Timer, res []*expr.Result, q string, now time.Time) (chart.Chart, error) {
+func (s *Schedule) ExprGraph(t miniprofiler.Timer, unit string, res []*expr.Result) (chart.Chart, error) {
 	c := chart.ScatterChart{
-		Title: fmt.Sprintf("%s - %s", q, now.Format(time.RFC1123)),
-		Key:   chart.Key{Pos: "itl"},
+		Key:    chart.Key{Pos: "itl"},
+		YRange: chart.Range{Label: unit},
 	}
 	c.XRange.Time = true
 	for ri, r := range res {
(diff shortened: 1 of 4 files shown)


captncraig:
Do we need to update docs for this?

cmd/bosun/sched/template.go
```diff
@@ -276,16 +276,19 @@ func (c *Context) EvalAll(v interface{}) (interface{}, error) {
...
+	res, exprText, err := c.eval(v, filter, true, 1000)
 	if err != nil {
 		return nil, err
 	}
 	var buf bytes.Buffer
 	const width = 800
 	const height = 600
+	footerHTML := fmt.Sprintf(`<small>Query: %s<br>Time: %s</small>`,
+		template.HTMLEscapeString(fmt.Sprint(exprText)),
+		fmt.Sprint(c.runHistory.Start.Format(time.RFC3339)))
```
captncraig: Why are we `Sprint`ing a string inside of a `Sprintf`?

cmd/bosun/sched/template.go
```diff
@@ -295,26 +298,30 @@ func (c *Context) graph(v interface{}, filter bool) (interface{}, error) {
...
 	}
+	buf.WriteString(footerHTML)
 	return template.HTML(buf.String()), nil
 }
 
 // Graph returns an SVG for the given result (or expression, for which it gets the result)
 // with same tags as the context's tags.
-func (c *Context) Graph(v interface{}) (interface{}, error) {
-	return c.graph(v, true)
+func (c *Context) Graph(v interface{}, unit ...string) (interface{}, error) {
```
captncraig: When would I pass multiple unit args? Is this just vararg to make it a non-breaking change? I would rather break people than use var args as optional parameters.

gbrayut: If you break it you get to up date all the existing templates :-P

Another option is a new Graph2 or GraphMeta function with the new signature

cmd/bosun/sched/template.go
kylebrandt: cause stupid.

kylebrandt:
@captncraig Yes should update docs, but I don't know how the new method works.

cmd/bosun/sched/template.go
kylebrandt: It is an optional argument, so it is nice to not have to specify an empty string if I don't have to.

captncraig:
lgtm

cmd/bosun/sched/chart.go
```diff
@@ -33,17 +31,20 @@ func Autostyle(i int) chart.Style {
...
-		LineStyle:  chart.SolidLine,
-		LineWidth:  1,
-		LineColor:  c,
+		Symbol:      0,
+		SymbolSize:  0.2,
+		SymbolColor: c,
+		LineStyle:   chart.SolidLine,
+		LineWidth:   1,
+		LineColor:   c,
+		Font:        chart.Font{Size: chart.NormalFontSize},
```
mjibson: Are you sure this isn't the default?

cmd/bosun/sched/chart.go
```diff
@@ -33,17 +31,20 @@ func Autostyle(i int) chart.Style {
 	c := chartColors[i%len(chartColors)]
 	return chart.Style{
 		// 0 uses a default
-		SymbolSize: 0.00001,
-		LineStyle:  chart.SolidLine,
-		LineWidth:  1,
-		LineColor:  c,
+		Symbol:      0,
```
mjibson: This is the default. Could we remove it?

cmd/bosun/sched/template.go
```diff
@@ -276,16 +276,19 @@ func (c *Context) EvalAll(v interface{}) (interface{}, error) {
...
+func (c *Context) graph(v interface{}, unit string, filter bool) (interface{}, error) {
+	res, exprText, err := c.eval(v, filter, true, 1000)
 	if err != nil {
 		return nil, err
 	}
 	var buf bytes.Buffer
 	const width = 800
 	const height = 600
+	footerHTML := fmt.Sprintf(`<small>Query: %s<br>Time: %s</small>`,
+		template.HTMLEscapeString(fmt.Sprint(exprText)),
```
mjibson: `exprText` is already a string, no need to convert it

cmd/bosun/sched/template.go
```diff
@@ -295,26 +298,28 @@ func (c *Context) graph(v interface{}, filter bool) (interface{}, error) {
...
 	}
+	buf.WriteString(footerHTML)
 	return template.HTML(buf.String()), nil
 }
 
 // Graph returns an SVG for the given result (or expression, for which it gets the result)
 // with same tags as the context's tags.
-func (c *Context) Graph(v interface{}) (interface{}, error) {
-	return c.graph(v, true)
+func (c *Context) Graph(v interface{}, unit string) (interface{}, error) {
```
mjibson: Having `unit` as the last argument means you can't do some things you could before with the go templates (see the [pipelines section](http://golang.org/pkg/text/template/#hdr-Pipelines) in the godocs). For example, here's a line copied from our prod.conf:

`{{printf "q(\"avg:rate{counter,,1}:os.cpu{host=%s}\", \"1d\", \"\")" (.Group.host) | .Graph}}`

This won't work anymore because you can only pass the last argument of a function via `|`.

One partial fix is to make `Graph` and `GraphAll` take a `...interface{}`. Then we could check the types and counts of things and handle them correctly. This would not break any existing syntax, however you still couldn't pass a y_label string to the above call.

One other possibility is to swap the order of the arguments.

We could also combine both solutions: swap and allow varargs. If the number of arguments is 1, then it's an expression. If it's 2, the first is a unit, the second is the expression.

docs/configuration.md
```diff
@@ -131,8 +131,8 @@ Templates are the message body for emails that are sent when an alert is trigger
 * Eval(string): executes the given expression and returns the first result with identical tags, or `nil` tags if none exists, otherwise `nil`.
 * EvalAll(string): executes the given expression and returns all results.
 * GetMeta(metric, name, tags): Returns metadata data for the given combination of metric, metadata name, and tag. `metric` and `name` are strings. `tags` may be a tag string (`"tagk=tagv,tag2=val2"`) or a tag set (`.Group`). If If `name` is the empty string, a slice of metadata matching the metric and tag is returned. Otherwise, only the metadata value is returned for the given name, or `nil` for no match.
-* Graph(string): returns an SVG graph of the expression with identical tags
-* GraphAll(string): returns an SVG graph of the expression
+* Graph(expression, y_label): returns an SVG graph of the expression with tags identical to the alert instance. `expression` and `y_label` are strings. The second argument was introduced in v0.2.1.
```
mjibson: I'm opposed to putting version notices in until we have a specific strategy for them. AngularJS has a dropdown on the side where you can select exactly the version you are running. Since we are going to soon have numbered releases, we could do that. Where did you get `0.2.1` from?

kylebrandt: @mjibson Craig said that this would be release under that version.

cmd/bosun/sched/template.go
kylebrandt: Is it important that we support pipes for that command? What is the benefit besides possibly better syntax? Also I think that example might just be left over from before Graph was context aware.

cmd/bosun/web/chart.go
```diff
@@ -122,7 +122,7 @@ func Graph(t miniprofiler.Timer, w http.ResponseWriter, r *http.Request) (interf
 	}
 	if _, present := r.Form["png"]; present {
 		c := chart.ScatterChart{
-			Title: fmt.Sprintf("%v - %v", oreq.Start, oreq.End),
+			Title: fmt.Sprintf("%v - %v", oreq.Start, queries),
```
mjibson: Why this change?

docs/configuration.md
```diff
@@ -131,8 +131,8 @@ Templates are the message body for emails that are sent when an alert is trigger
 * Eval(string): executes the given expression and returns the first result with identical tags, or `nil` tags if none exists, otherwise `nil`.
 * EvalAll(string): executes the given expression and returns all results.
 * GetMeta(metric, name, tags): Returns metadata data for the given combination of metric, metadata name, and tag. `metric` and `name` are strings. `tags` may be a tag string (`"tagk=tagv,tag2=val2"`) or a tag set (`.Group`). If If `name` is the empty string, a slice of metadata matching the metric and tag is returned. Otherwise, only the metadata value is returned for the given name, or `nil` for no match.
-* Graph(string): returns an SVG graph of the expression with identical tags
-* GraphAll(string): returns an SVG graph of the expression
+* Graph(expression, y_label): returns an SVG graph of the expression with tags identical to the alert instance. `expression` is a string or an expression and `y_label` is a string. `y_label` is an optional argument.
```
mjibson: Perhaps use `[, y_label]` to also indicate that it's optional?

cmd/bosun/web/chart.go
kylebrandt: End was often nil and it wasn't handled. I'd vote for yanking out this feature all together.

PR #1011, opened 2015-05-29
cmd/bosun: Show bar graph on the expr page if type == number

kylebrandt:
[Review on Reviewable](https://reviewable.io/reviews/bosun-monitor/bosun/1011)


captncraig:
lgtm if it works

PR #1012, opened 2015-05-29
cmd/bosun: window function

mjibson:


kylebrandt:
Query errors do not bubble up.

mjibson:
Syntax errors now bubble.

cmd/bosun/expr/funcs.go
```diff
@@ -528,6 +570,120 @@ func GraphiteBand(e *State, T miniprofiler.Timer, query, duration, period, forma
...
+	fn, ok := e.GetFunction(rfunc)
+	if !ok {
+		return nil, fmt.Errorf("expr: Window: no %v function", rfunc)
+	}
+	f := reflect.ValueOf(fn.F)
+	r = new(Results)
+	r.IgnoreOtherUnjoined = true
+	r.IgnoreUnjoined = true
+	T.Step("window", func(T miniprofiler.Timer) {
+		var d, p opentsdb.Duration
```
mjibson: This whole first part should probably be a function since both Band and Window call it.

cmd/bosun/expr/funcs.go
```diff
@@ -528,6 +570,120 @@ func GraphiteBand(e *State, T miniprofiler.Timer, query, duration, period, forma
...
+		d, err = opentsdb.ParseDuration(duration)
+		if err != nil {
+			return
+		}
+		p, err = opentsdb.ParseDuration(period)
+		if err != nil {
+			return
+		}
+		if num < 1 || num > 100 {
+			err = fmt.Errorf("expr: Band: num out of bounds")
```
mjibson: Band -> Window

cmd/bosun/expr/funcs.go
```diff
@@ -104,24 +107,35 @@ var TSDB = map[string]parse.Func{
...
 		tagQuery,
 		Change,
+		nil,
 	},
 	"count": {
 		[]parse.FuncType{parse.TypeString, parse.TypeString, parse.TypeString},
 		parse.TypeScalar,
 		nil,
 		Count,
+		nil,
```
captncraig: It feels a little wrong to add a new field to every one of these just because one or two functions use it. Are there any good alternatives?

cmd/bosun/expr/funcs.go
```diff
@@ -574,39 +618,122 @@ func Band(e *State, T miniprofiler.Timer, query, duration, period string, num fl
...
+	return
+}
+
+func Window(e *State, T miniprofiler.Timer, query, duration, period string, num float64, rfunc string) (r *Results, err error) {
+	fn, ok := e.GetFunction(rfunc)
+	if !ok {
+		return nil, fmt.Errorf("expr: Window: no %v function", rfunc)
+	}
+	f := reflect.ValueOf(fn.F)
+	r, err = bandTSDB(e, T, query, duration, period, num, func(r *Results, res *opentsdb.Response) error {
```
captncraig: This function is kinda big to be passed in the argument list. Can we not make it anonymous?

cmd/bosun/expr/funcs.go
```diff
@@ -574,39 +618,122 @@ func Band(e *State, T miniprofiler.Timer, query, duration, period string, num fl
...
+			}
+			if min == 0 || i < min {
+				min = i
+			}
+			values[time.Unix(i, 0).UTC()] = float64(v)
+		}
+		if len(values) == 0 {
+			return nil
+		}
+		a.Value = values
```
captncraig: The number of one letter variable names gets real confusing around this point. Perhaps it would be worth investing in some more descriptive names.

cmd/bosun/expr/funcs.go
mjibson: I looked into merging it with the Tags function, but it wasn't a good fit. However I can switch to named parameters, which will eliminate the need for nil if unused.

captncraig: Yeah, that would cause a ton more changes, but may make things more readable throughout. Your call.

mjibson:
updated with improved var names and named params for parse.Func, which I've been wanting to do for a while anyway.

captncraig:
Looks good to me. Can we make a test that exercises it?

mjibson:
Added tests for window function. Also suitable for testing other functions, should we so choose. Other change is to modify cache to accept a nil value that does nothing.

mjibson:
Also the test found a tiiiiiny bug if you happened to have data at the Unix epoch. Worth.

captncraig:
LGTM

Issue #974, opened 2015-05-19
Allow to configure a concrete indexRoot to lscount/lsstat


ayashjorden:
Hi,
Sometimes, one would like (the great) Bosun to query the same Elasticsearch (ES) index for an alert.
Current implementation force the user to use 'prefix-yyyy.MM.dd' index naming convention.

Suggested is to allow the user to configure an indexRoot that will be used as the index name, without Bosun need to suffix it with something or any other changes.

Example use-case is ES indices with a small size of a couple of MB per day. ES indices are an expansive unit in matter of RAM and CPU to maintain it.

Currently I solved this by creating a single index 'my_index-global' and funneling all the relevant logs into it. the I crontab-ed an alias creation  command to add aliases matching Bosun requirements.

Thanks,
Yarden

kylebrandt:
Best idea I have thought of is to have the prefix argument take a special
character at the end that is not allowed in index names....

gbrayut:
Or you could use a verbose postfix. logstashWITHOUT-A-DATE works just as well as logstash@ or whatever obscure character with unknown meaning we can come up with

ayashjorden:
Verbose postfix may create noise in configuration code, and we cant use the
$(regex end of line and bosun var prefix)
Can we do #,<,@       ?

gbrayut:
I think https://github.com/elastic/elasticsearch/issues/9059 talks about the valid characters for index names.

* must be lowercase
* must NOT start with an underscore, a + or a -
* index names starting with '.' are reserved
* . and .. are forbidden, though they can occur in the name
* `\`, `/`, `*`, `?`, `"`, `<`,` >`, `|`,` `,`,` are forbidden

kylebrandt:
I don't have any particular preference, which character if single
character, or if verbose

gbrayut:
I'd say * and ? should be avoided since we may want to support globs on metrics at some point. So any of the other forbidden index characters would work.

ayashjorden:
Is pipe a reserved char in Go Template or somewhere in Bosun?

ayashjorden:
@kylebrandt suggested to short-circut GenIndex function after initializing ES client:
```GO
	if len(r.IndexRoot) > 0 {
		if r.IndexRoot[len(r.IndexRoot)-1] == '/' {
			return r.IndexRoot[:len(r.IndexRoot)-1], nil
		}
	}
```

Tested on my dev-env and it works great!
Can you please add it to next release?

[...]

PR #1013, opened 2015-05-31
Concrete ls index name

ayashjorden:
Tested '/' functionality as concrete ES index name signaling (without Bosun adding date suffix), works great.

[Review on Reviewable](https://reviewable.io/reviews/bosun-monitor/bosun/1013)

### DIFF
--- a/cmd/bosun/expr/logstash.go
+++ b/cmd/bosun/expr/logstash.go
@@ -153,7 +153,7 @@ func (e *LogstashElasticHosts) GenIndices(r *LogstashRequest) (string, error) {
 		}
 	}
 	if len(selectedIndices) == 0 {
-		return "", fmt.Errorf("no elastic indices available during this time range, start/end: [%s|%s]", start, end)
+		return "", fmt.Errorf("no elastic indices available during this time range")
 	}
 	return strings.Join(selectedIndices, ","), nil
 }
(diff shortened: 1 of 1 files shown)


cmd/bosun/expr/logstash.go
```diff
@@ -122,6 +122,12 @@ func (e *LogstashElasticHosts) GenIndices(r *LogstashRequest) (string, error) {
 	if err != nil {
 		return "", err
 	}
+	// Short-circut when using concrete ES index name
+	if len(r.IndexRoot) > 0 {
```
mjibson: This and the line below can be reduced to `if strings.HasSuffix(r.IndexRoot, "/")`.

ayashjorden:
@mjibson, (as I'm new to GO) Should I reduce it?

mjibson:
Yes, you should. But I'm a bit confused about the intended behavior. That will test if just the last character is a "/", not if the entire string has value "/". Which do you want?

ayashjorden:
As you can see in https://github.com/bosun-monitor/bosun/issues/974 , the intended behavior is to enable the user to signal Bosun 'just use this index name and don't add date suffix'.

The context behind this is that ES indices 'costs' quite a lot in terms of memory and cpu. there are use-cases such as the one that initiated this PR that the size of data is <5MB per day so there is no reason to create new indices every day. also there's the fact that olivere/elastic client (that Bosun is using to interact with ES) query's indices by index name, and doesn't include index aliases which you can think of sym-links from 'myindex-2015.05.31' to 'myindex-global'.

Hope that the above is clear enoungh,
Yarden

mjibson:
Strings can't be nil, but they can be empty, and HasSuffix handles that correctly.

mjibson:
@kylebrandt, could you review this?

kylebrandt:
+1

PR #1018, opened 2015-06-01
cmd/bosun: Change diff to be a reduction function

kylebrandt:
This way it can be used with the upcoming window function. It was made a query function when only OpenTSDB existed, therefore diff(q(...)) was not needed, but now that there are different types of query functions it is useful as a generic reduction function.


mjibson:
Also needs fixes for the docs.

captncraig:
LGTM.

PR #1021, opened 2015-06-01
docs: making release notes prettier

captncraig:


build/docs/publish.go
```diff
@@ -22,6 +23,7 @@ var (
 	client   *github.Client
 	msg      = "syncing to gh-pages"
 	gh_pages = "heads/gh-pages"
+	force    = flag.Bool("f", false, "Push an new commit, even if code is identical. Forces gh-pages rebuild.")
```
mjibson: an -> a

mjibson: This should perhaps default to on, and `-f` disables it.

mjibson:
LGTM after style changes. Ignore the `-f` default to on if you want, I don't care, just thinking out loud.

PR #1023, opened 2015-06-02
Add missing docs

mjibson:


docs/expressions.md
```diff
@@ -330,6 +345,20 @@ Remove any values lower than or equal to number from a series. Will error if thi
 
 Remove any NaN or Inf values from a series. Will error if this operation results in an empty series.
 
+## epoch()
+
+Returns the Unix epoch in seconds of the expression start time (scalar).
+
+## filter(series, number)
+
+Returns all results in series that are a subset of anything in number, or
```
captncraig: this is confusing to me.

kylebrandt: Filters the first argument, `series` based on the second argument, `number`. Only the groups in the first argument that are a subset of the second argument are returned. So this allows you to filter out series based on the same or a different series.

Does that help at all @captncraig ?

gbrayut: I agree it is still a bit confusing. Can you add some examples? Is the second argument a series, a single double/int, or a slice of time/value pairs? If it is a series, how can I create them, since we don't have series literals.

docs/expressions.md
```diff
@@ -330,6 +345,20 @@ Remove any values lower than or equal to number from a series. Will error if thi
...
+that have number as a subset. Useful with the limit and sort functions to
+return the top X results of a query.
+
+## limit(number, count)
```
gbrayut: Maybe we need to define what number is... because I'm not sure I've ever seen a definition for number vs series

kylebrandt: Read: http://bosun.org/expressions.html#data-types

gbrayut: Then you shouldn't use number as the type for the second argument, as you can't have a subset of "A number is a single numeric value with a group associated with it.". We should add a new type called NumerSeries or NumberSet.

gbrayut: I often get confused when looking at the documents as we dont make a distinction between single numerical values and multiple numerical values.

mjibson: I've opened https://github.com/bosun-monitor/bosun/issues/1026 to discuss how to better handle documenting function arguments and types.

PR #1032, opened 2015-06-03
bosun: Ability to throttle log notifications

### LINKED ISSUE #975: Add LogDelay parameter to throttle log notifications.
Proposal is to add a `LogDelay = duration` parameter to alerts that will *only* be valid for alerts with `Log = true`.

If this value is set, no two notifications will be sent for an alert key within `duration`. If `LogDelay = 5m` is set, and check frequency is `1m`, we can evaluate the alert every 1m, but only notify every 5m.

captncraig:
Adds `maxLogFrequency` key to alerts that can control how often notification gets run. Only valid on log alerts. 

Fixes #975

### DIFF
--- a/cmd/bosun/sched/check.go
+++ b/cmd/bosun/sched/check.go
@@ -175,11 +175,12 @@ func (s *Schedule) RunHistory(r *RunHistory) {
 		// Do nothing if state did not change.
 		notify := func(ns *conf.Notifications) {
 			if a.Log {
-				lastLogTime, ok := s.lastLogTimes[ak]
-				if ok && lastLogTime.Add(a.MaxLogFrequency).After(time.Now()) {
+				lastLogTime := state.LastLogTime
+				now := time.Now()
+				if now.Before(lastLogTime.Add(a.MaxLogFrequency)) {
 					return
 				}
-				s.lastLogTimes[ak] = time.Now()
+				state.LastLogTime = now
 			}
 			nots := ns.Get(s.Conf, state.Group)
 			for _, n := range nots {
--- a/cmd/bosun/sched/check.go
+++ b/cmd/bosun/sched/check.go
@@ -174,6 +174,13 @@ func (s *Schedule) RunHistory(r *RunHistory) {
 		// If the old alert was not acknowledged, do nothing.
 		// Do nothing if state did not change.
 		notify := func(ns *conf.Notifications) {
+			if a.Log {
+				lastLogTime, ok := s.lastLogTimes[ak]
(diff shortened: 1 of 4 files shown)


cmd/bosun/sched/check.go
```diff
@@ -174,6 +174,13 @@ func (s *Schedule) RunHistory(r *RunHistory) {
 		// If the old alert was not acknowledged, do nothing.
 		// Do nothing if state did not change.
 		notify := func(ns *conf.Notifications) {
+			if a.Log {
+				lastLogTime, ok := s.lastLogTimes[ak]
+				if ok && lastLogTime.Add(a.MaxLogFrequency).After(time.Now()) {
```
mjibson: how about set `now := time.Now()` to avoid calling it twice?

cmd/bosun/sched/sched.go
```diff
@@ -43,6 +43,7 @@ type Schedule struct {
 	LastCheck     time.Time
 	nc            chan interface{}
 	notifications map[*conf.Notification][]*State
+	lastLogTimes  map[expr.AlertKey]time.Time
```
mjibson: Thoughts about putting LastLogTime on the State type instead? Then it could survive restarts and we could avoid the potential need to lock the map.

cmd/bosun/sched/check.go
mjibson: Shouldn't that be `.Before`?

cmd/bosun/sched/sched.go
captncraig: I thought about persisting it and decided it probably isn't worth saving. Perhaps @gbrayut would have input there. Locking is a good point. 

Reasoning for a schedule map is that it will likely consume less memory since it will usually be pretty sparse. I could go either way though.

cmd/bosun/sched/check.go
captncraig: LastTime + 5m is next time we can log. 

If now is before that we should not log. I think mine is correct but I agree it would make more sense to do `now.Before(last + duration)`

cmd/bosun/sched/sched.go
mjibson: Let's move it to State. It's just a few extra bytes, and I don't want to start spreading data around in maps when we already have a place per alertkey.

cmd/bosun/sched/check.go
mjibson: You're right. I was just confused. No need to change unless you feel strongly.

mjibson:
LGTM

PR #1038, opened 2015-06-04
cmd/bosun: validate config via POST

### LINKED ISSUE #1033: Config Validation should use POST
Since it goes via HAProxy in our prod, there is a connection error because HAProxy's default buffers are not big enough. We could change this, but other people will probably have proxy issues so I think we should change this to post.

mjibson:
fixes #1033


captncraig:
needs generate again

mjibson:
updated

captncraig:
LGTM

PR #1039, opened 2015-06-04
docs: Allowing docs to be embedded and hosted in main app.

captncraig:
An interesting idea matt had. The documentation will be hosted inside bosun itself at /docs.

I tried to update links to relative (`./whatever` instead of `/whatever`) but I may have missed some.

`go generate` compiles jekyll stuff and uses esc to make a `cmd/bosun/web/docs` package to serve statically. Dev mode is not currently implemented, and go generate is required on making changes.

[Review on Reviewable](https://reviewable.io/reviews/bosun-monitor/bosun/1039)


docs/_layouts/default.html
```diff
@@ -13,7 +13,7 @@
 		Bosun
 	</title>
 
-	<link rel="stylesheet" href="/styles/bootstrap.min.css">
+	<link rel="stylesheet" href="./styles/bootstrap.min.css">
```
mjibson: `./` should be removed

mjibson:
- Why are all the doc publish scripts being removed? We still need bosun.org.
- If this really needs esc to run, where are the esc files in this PR?
- Do we really want to make jekyll a dependency to get bosun generate to run? I don't. I already don't like that we require typescript just to run go generate.
- How will users of bosun know about this?

Code review comments aside, I feel weird about this idea. I'm not sure it's good. Rather, what problem is it solving?

captncraig:
@mjibson 
- the scripts in /docs should be superceded by scripts in /build.
- cmd/bosun/web/docs/docs.go is the esc generated file. The source (generated jekyll) is not committed.
- I agree about jekyll somewhat. Perhaps we should make it a generate on just the web/docs package so it is run seperately. 
- The main documentation link could potentially go to the local copy instead of bosun.org. I could see it either way. that way bosun.org is more for advertising than for docs for active users.

If it is too cumbersome for the build processes, I am not married to this idea. But it was a fun experiment.

PR #1040, opened 2015-06-05
build: various improvements

mjibson:
- determine git SHA automatically
- more explicit variable to trigger bosun build
- allow no GitHub access token that still builds binaries for all archs


captncraig:
Aren't all of these changes also in the solaris pr?

mjibson:
Yes, but git knows that and will only commit them once. It's a branch on a branch.

PR #1042, opened 2015-06-05
Solaris

mjibson:


build/release.sh
```diff
@@ -16,25 +17,28 @@ build()
...
-go run build/release/githubRelease.go
+build linux arm 5
+build linux arm 6
+build linux arm 7
+build solaris amd64
 
+if [ "$GITHUB_ACCESS_TOKEN" = "" ]; then
+	echo GITHUB_ACCESS_TOKEN not set: not running githubRelease.go
+else
+	export BUILD_NUMBER=`${OUTPUTDIR}bosun-linux-amd64 -version | awk '{print $3}'`
```
captncraig: I think we need to reset GOOS and GOARCH to the native defaults for go run to work here. Don''t want to run as solaris.

mjibson: I've added that to the other PR, where it belongs (since it'd be running ARM). I'll re-fork this one after that one gets merged.

PR #1044, opened 2015-06-05
scollector: httpunit collector

mjibson:


cmd/scollector/collectors/httpunit.go
```diff
@@ -0,0 +1,79 @@
...
+	collectors = append(collectors, &IntervalCollector{
+		F: func() (opentsdb.MultiDataPoint, error) {
+			return cHTTPUnit(plans)
+		},
+		name: fmt.Sprintf("c_httpunit_%s", name),
+	})
+}
+
+func cHTTPUnit(plans *httpunit.Plans) (opentsdb.MultiDataPoint, error) {
+	ch, _, err := plans.Test("", false)
```
captncraig: general api design question: What happens with errors that happen later in the check process? Do they get propagated through the result channel, or just swallowed?

cmd/scollector/collectors/httpunit.go
```diff
@@ -0,0 +1,79 @@
...
+			er = 1
+		}
+		Add(&md, "hu.error", er, tags, metadata.Gauge, metadata.Bool, "")
+		var sc = 0
+		if r.Result.Connected {
+			sc = 1
+		}
+		Add(&md, "hu.socket_connected", sc, tags, metadata.Gauge, metadata.Bool, descHTTPUnitSocketConnected)
+		switch r.Case.URL.Scheme {
+		case "http", "https":
```
captncraig: what other possibilities are there?

cmd/scollector/collectors/httpunit.go
```diff
@@ -0,0 +1,79 @@
...
+				Add(&md, "hu.cert.valid", cv, tags, metadata.Gauge, metadata.Bool, "")
+				Add(&md, "hu.cert.expires", r.Result.Resp.TLS.PeerCertificates[0].NotAfter.Unix(), tags, metadata.Gauge, metadata.Timestamp, "")
+			}
+		}
+	}
+	return md, nil
+}
+
+const (
+	descHTTPUnitError           = "1 if any error (no connection, unexpected code or text) occurred, else 0."
```
captncraig: Should we invert this condition? That way 1 is always good and 0 is always bad for hu metrics. Just a thought.

metadata/metadata.go
```diff
@@ -94,6 +94,7 @@ const (
 	StatusCode          = "status code"
 	Syscall             = "system calls"
 	Thread              = "threads"
+	Timestamp           = "timestamp"
```
captncraig: Does bosun have capabilities to alert properly on timestamps? Would we be better of reporting time remaining till expiration?

cmd/scollector/collectors/httpunit.go
mjibson: Errors during tests get sent to the results channel and the `hu.error` metric gets set to 1.

cmd/scollector/collectors/httpunit.go
mjibson: tcp, udp, ip, 4 and 6 versions of those three. But we don't yet have special things for those, hence the case.

cmd/scollector/collectors/httpunit.go
mjibson: We discussed this on a call and it was decided that (unlike status exit codes), these were closer to booleans, and so 1 means "present", hence, 1 means there was an error.

captncraig:
lgtm

metadata/metadata.go
mjibson: I think leaving as a timestamp is correct because:

1. bosun has epoch and since functions that work with timestamps
2. the thing being measured here is the expiration date of a ssl certificate, and i think the thing we want is to record exactly what's in the certificate. that is, we can record the facts, and compute what we want from that.

PR #1050, opened 2015-06-08
docker: cd to correct directory

mjibson:


captncraig:
Nitpicking, but I would prefer to do this with `WORKDIR`

mjibson:
Changed to use WORKDIR.

Also changed to use the bosun directory already cloned from the local git repo. (i.e., the `go get -d` was unneeded.)

captncraig:
LGTM

PR #1052, opened 2015-06-09
cmd/scollector: convert bools to numbers in httpunit

mjibson:


mjibson:
Other options instead of this one:

- change Add to do this itself to prevent this kind of error in the future
- have Add detect and log these problems, but not modify the data

Thoughts? I sort of favor the first one, but just a little.

elementai:
Second option is the least surprising one.

captncraig:
I would support having Add manage this for you. true/false to 1/0 seems pretty straightforward.

PR #1055, opened 2015-06-09
build: Using correct env vars

captncraig:


build/dockerRelease.sh
```diff
@@ -1,10 +1,10 @@
 #!/bin/sh
 # Must be run from $GOPATH/src/bosun.org, not the build directory.
-set -e
```
mjibson: I don't understand why this file changed.

build/dockerRelease.sh
```diff
@@ -4,7 +4,7 @@ set -e
 
 rm -rf buildoutput
 mkdir buildoutput
-set -e
```
mjibson: What's wrong with `set -e`?

PR #1061, opened 2015-06-11
bosun: denormalizing tsdb metrics in relay

captncraig:
opentsdb metrics with high tag cardinality have poor query performance. This allows the tsdb relay to denormalize and duplicate data based on tags.

It can transform `os.mem.used{host=web01}` to `__web01.os.mem.used{host=web01}` so that queries can use the much lower cardinality metric. Tags, Timestamps and Values will be unchanged in denormalized metrics.

Current configSyntax = `-denormalize=metric1_someTag,metric2_host_cpu`. Allow multiple metrics and multiple tag potentially per metric.

[Review on Reviewable](https://reviewable.io/reviews/bosun-monitor/bosun/1061)


cmd/tsdbrelay/main.go
```diff
@@ -104,4 +115,65 @@ func (rp *relayProxy) ServeHTTP(responseWriter http.ResponseWriter, r *http.Requ
 		resp.Body.Close()
 		verbose("bosun relay success")
 	}()
+	if parse && denormalizationRules != nil {
+		go rp.denormalize(&reader.buf)
```
mjibson: Use `bytes.NewReader(reader.buf.Bytes())` here, as above, because otherwise the reader position of the buffer could be affected by or affect other go routines.

cmd/tsdbrelay/main.go
```diff
@@ -104,4 +115,65 @@ func (rp *relayProxy) ServeHTTP(responseWriter http.ResponseWriter, r *http.Requ
...
+	err = encoder.Encode(relayDps)
+	if err != nil {
+		verbose("error encoding denormalized data points: %v", err)
+		return
+	}
+	if err = gWriter.Close(); err != nil {
+		verbose("error zipping denormalized data points: %v", err)
+		return
+	}
+	u := &url.URL{
```
mjibson: Make this a global var or something since we do it twice.

cmd/tsdbrelay/denormalization.go
```diff
@@ -0,0 +1,69 @@
...
+func parseDenormalizationRules() {
+	if *denormalize == "" {
+		return
+	}
+	denormalizationRules = make(map[string]*DenormalizationRule)
+	rules := strings.Split(*denormalize, ",")
+	for _, r := range rules {
+		parts := strings.Split(r, "__")
+		if len(parts) < 2 {
+			log.Fatal("Denormalization rules must have at least one tag name specified.")
```
mjibson: return err

captncraig: This is on config parsing. I thought we failed fast in that case.

mjibson: yes, but let's try to contain log.Fatal to main(). this should probably take a string, and return the rules, so that it's testable.

captncraig: that's fair.

cmd/tsdbrelay/denormalization.go
```diff
@@ -0,0 +1,69 @@
...
+		}
+		log.Println("Denormalizing", rule)
+		denormalizationRules[rule.Metric] = rule
+	}
+}
+
+func (d *DenormalizationRule) Translate(dp *opentsdb.DataPoint) (*opentsdb.DataPoint, error) {
+	newDp := &opentsdb.DataPoint{Timestamp: dp.Timestamp, Value: dp.Value}
+	tagString := "__"
+	newDp.Tags = dp.Tags
```
mjibson: can this be in the init 2 lines above?

cmd/tsdbrelay/denormalization.go
```diff
@@ -0,0 +1,69 @@
...
+		rule := &DenormalizationRule{Metric: parts[0]}
+		for i := 1; i < len(parts); i++ {
+			rule.TagNames = append(rule.TagNames, parts[i])
+		}
+		log.Println("Denormalizing", rule)
+		denormalizationRules[rule.Metric] = rule
+	}
+}
+
+func (d *DenormalizationRule) Translate(dp *opentsdb.DataPoint) (*opentsdb.DataPoint, error) {
```
mjibson: does it make sense to just do this in place instead of return a new dp?

cmd/tsdbrelay/denormalization.go
```diff
@@ -0,0 +1,69 @@
...
+		if !ok {
+			return nil, fmt.Errorf("tag %s not present in data point for %s.", tagName, dp.Metric)
+		}
+		if i > 0 {
+			tagString += "."
+		}
+		tagString += val
+	}
+	newDp.Metric = tagString + "." + dp.Metric
+	return newDp, nil
```
mjibson: this function doesn't remove any tags from the data point, it just makes the metric name longer. is that the intention?

captncraig: That is correct. If you remove the tags, you risk sending tagless data which tsdb does not like.

cmd/tsdbrelay/denormalization.go
```diff
@@ -0,0 +1,65 @@
...
+		inputTags += fmt.Sprintf("%s=%s", tagk, string(val))
+		outputTags += fmt.Sprintf("%s", string(val))
+		val++
+	}
+	return fmt.Sprintf("%s{%s} -> __%s.%s", d.Metric, inputTags, outputTags, d.Metric)
+}
+
+var denormalizationRules map[string]*DenormalizationRule
+
+func parseDenormalizationRules(config string) (map[string]*DenormalizationRule, error) {
```
mjibson: Isn't this a `denormalizationRules` type?

cmd/tsdbrelay/denormalization.go
```diff
@@ -0,0 +1,65 @@
...
+	m := make(map[string]*DenormalizationRule)
+	rules := strings.Split(config, ",")
+	for _, r := range rules {
+		parts := strings.Split(r, "__")
+		if len(parts) < 2 {
+			return nil, fmt.Errorf("Denormalization rules must have at least one tag name specified.")
+		}
+		rule := &DenormalizationRule{Metric: parts[0]}
+		for i := 1; i < len(parts); i++ {
```
mjibson: `for _, part := range parts[1:]`

cmd/tsdbrelay/main.go
```diff
@@ -15,6 +20,12 @@ var (
 	bosunServer = flag.String("b", "bosun", "Target Bosun server. Can specify port with host:port.")
 	tsdbServer  = flag.String("t", "", "Target OpenTSDB server. Can specify port with host:port.")
 	logVerbose  = flag.Bool("v", false, "enable verbose logging")
+	denormalize = flag.String("denormalize", "os.mem.used__host", "List of metrics to denormalize. Comma seperated list of `metric__tagname__tagname` rules. Will be translated to `___metric__tagvalue__tagvalue`")
```
mjibson: Why is the default `os.mem.used__host`? Should it instead be the empty string?

cmd/tsdbrelay/main.go
```diff
@@ -15,6 +20,12 @@ var (
...
+)
+
+var (
+	tsdbPutURL    *url.URL
```
mjibson: Can we make these strings? They never change.

cmd/tsdbrelay/denormalization.go
```diff
@@ -0,0 +1,65 @@
...
+			rule.TagNames = append(rule.TagNames, parts[i])
+		}
+		log.Println("Denormalizing", rule)
+		m[rule.Metric] = rule
+	}
+	return m, nil
+}
+
+func (d *DenormalizationRule) Translate(dp *opentsdb.DataPoint) (*opentsdb.DataPoint, error) {
```
mjibson: Since this now modifies dp in place, I think the return signature should just be `error`. In our code we have `newDp := d.Translate(...)`, which implies a new dp is created, which is dangerous.

captncraig:
Review status: 0 of 7 files reviewed, 7 unresolved discussions, some commit checks failed.

---

<sup>**[cmd/tsdbrelay/denormalization.go, line 33 \[r3\]](https://reviewable.io:443/reviews/bosun-monitor/bosun/1061#-Jrd6JazBSbmFkpoVcBR-r3-33)** ([raw file](https://github.com/bosun-monitor/bosun/blob/d7cc729d9e107381ba0d70ac00075f4d4622e5aa/cmd/tsdbrelay/denormalization.go#L33)):</sup>
denormalizationRules is a var, not a type.

---

<sup>**[cmd/tsdbrelay/denormalization.go, line 52 \[r2\]](https://reviewable.io:443/reviews/bosun-monitor/bosun/1061#-Jrd6JazBSbmFkpoVcBR-r2-56)** ([raw file](https://github.com/bosun-monitor/bosun/blob/cde66b11ecae8b67622c3070b320e793d731d4e7/cmd/tsdbrelay/denormalization.go#L52)):</sup>
not really, the tag string is built from the tag values in the particular data point.

---

<sup>**[cmd/tsdbrelay/main.go, line 23 \[r3\]](https://reviewable.io:443/reviews/bosun-monitor/bosun/1061#-Jrd6JazBSbmFkpoVcBQ-r3-23)** ([raw file](https://github.com/bosun-monitor/bosun/blob/d7cc729d9e107381ba0d70ac00075f4d4622e5aa/cmd/tsdbrelay/main.go#L23)):</sup>
testing purposed. will change.

---

<sup>**[cmd/tsdbrelay/main.go, line 27 \[r3\]](https://reviewable.io:443/reviews/bosun-monitor/bosun/1061#-Jrd6JazBSbmFkpoVcBQ-r3-27)** ([raw file](https://github.com/bosun-monitor/bosun/blob/d7cc729d9e107381ba0d70ac00075f4d4622e5aa/cmd/tsdbrelay/main.go#L27)):</sup>
Done.

---

<sup>**[cmd/tsdbrelay/main.go, line 136 \[r1\]](https://reviewable.io:443/reviews/bosun-monitor/bosun/1061#-Jrd6JazBSbmFkpoVcBQ-r1-119)** ([raw file](https://github.com/bosun-monitor/bosun/blob/317667dab9be3bf84688b5c8da25502e18ccde65/cmd/tsdbrelay/main.go#L136)):</sup>
Done.

---

<sup>**[cmd/tsdbrelay/main.go, line 178 \[r1\]](https://reviewable.io:443/reviews/bosun-monitor/bosun/1061#-Jrd6JazBSbmFkpoVcBQ-r1-162)** ([raw file](https://github.com/bosun-monitor/bosun/blob/317667dab9be3bf84688b5c8da25502e18ccde65/cmd/tsdbrelay/main.go#L178)):</sup>
Done.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/bosun-monitor/bosun/1061)

mjibson:
LGTM

skygragon:
> opentsdb metrics with high tag cardinality have poor query performance.

@captncraig is there any detailed perf data or suffering sightings for this?

PR #1067, opened 2015-06-12
tsdbrelay: build improvements and versioning

captncraig:
[Review on Reviewable](https://reviewable.io/reviews/bosun-monitor/bosun/1067)

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -31,7 +31,7 @@ func main() {
 
 	ldFlags := fmt.Sprintf("-X bosun.org/version.VersionSHA %s -X bosun.org/version.VersionDate %s", sha, timeStr)
 
-	for _, app := range []string{"bosun", "scollector"} {
+	for _, app := range []string{"bosun", "scollector", "tsdbrelay"} {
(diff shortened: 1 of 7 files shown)


mjibson:
I don't think we should change the docker image. Putting tsdbrelay in there doesn't solve any problems.

If we are going to build the binaries in build.go, then they should also be added in the release script and a page similar to the scollector page needs to be added.

---

Review status: 0 of 7 files reviewed, 1 unresolved discussion, all commit checks successful.

---

<sup>**[cmd/tsdbrelay/main.go, line 38 \[r1\]](https://reviewable.io:443/reviews/bosun-monitor/bosun/1067#-JrdVq-VF5Tw3S2dTZax)** ([raw file](https://github.com/bosun-monitor/bosun/blob/11b5e01d0dc0ec6079007559b46ad2737ee646fc/cmd/tsdbrelay/main.go#L38)):</sup>
`bosun` -> `tsdbrelay`

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/bosun-monitor/bosun/1067)

Issue #1026, opened 2015-06-02
Documenting expression function arguments


mjibson:
The docs for our expression arguments are totally unclear, and only usable with examples. The problems are:

1. The difference between number and scalar is confusing. It should be obvious (somehow) that scalars are just normal numbers, and a "number" is the result of an expression that is a series. The tagged vs untagged thing here is hard to explain.
2. Arguments are sometimes named and sometimes typed. That is, sometimes we describe the type of the argument, and sometimes we put a human-readable name as the argument. We need to do both.
3. Return types are also poorly documented.

We need to come up with a standard format to describe these. Here's a proposal:

`## func_name(var_name var_type[, name2 type2]) return_type`

Example:

```
## forecastlr(s series, y_val scalar) number

Returns the number of seconds a linear regression predicts s will take to reach y_val.
```

kylebrandt:
All types should href link to the type def in document

kylebrandt:
Matt and I agreed that changing these to seriesSet, numberSet, and scalar will help make things clearer. Thanks to Greg for highlighting and clarifying this point of confusion!

gbrayut:
Some other examples to keep in mind when trying to figure out the new format:

**lookup(table, key)** Kyle and I were not even sure what the return type was for this function

**sort(number, asc_desc)** A better syntax for restricted arguments would be nice, like <"asc" | "desc">

**filter(series, number)** If I remember correctly this has a different return type than most other functions. When using in an expression you may not notice, but if you use these results in a template it may behave differently than other functions.

Speaking of which, whatever format we use to document expression functions we should also plan on using to document template functions. These have a different invocation due to how go templates work, but are much more particular about input types ( an int literal like 1 does not work where a double is required). Not saying it needs to exactly match how we document expressions, but whatever we do now should eventually line up with what we use to document template functions.

Example of a template/functions we may want to document using a similar same syntax:

```
{{$sources := (.EvalAll .Alert.Vars.source).DescByValue}}
{{$top_source := index $sources 0}}
{{$source_display_at := .Eval .Alert.Vars.source_display_at}}
...
{{ if not .IsEmail }}
            {{if gt (.Eval .Alert.Vars.sources_display_count) 1.0}}
                Sources: {{range $r := $sources}}
                    {{ if gt $r.Value $source_display_at }}
                        https://{{$.LookupAll "exception_source" "hostname" $r.Group}} = {{$r.Value | printf "%.f"}}
                    {{end}}
[...]

PR #1071, opened 2015-06-15
Documenting function parameter types and returns

captncraig:
This also includes formally renaming Number to NumberSet to cause less confusion.

[Review on Reviewable](https://reviewable.io/reviews/bosun-monitor/bosun/1071)

### DIFF
--- a/docs/expressions.md
+++ b/docs/expressions.md
@@ -25,12 +25,12 @@ This section documents Bosun's expression Language. Bosun's expression language 
 There are three data types in Bosun's expression language:
 
  1. **Scalar**: This is the simplest type, it is a single numeric value with no group associated with it. Keep in mind that an empty group, `{}` is still a group.
- 2. **Number**: A number is a single numeric value with a group associated with it.
- 3. **Series**: A series is an array of timestamp-value pairs and an associated group.
+ 2. **NumberSet**: A number set is a group of tagged numeric values with one value per unique grouping.
+ 3. **SeriesSet**: A series is an array of timestamp-value pairs and an associated group.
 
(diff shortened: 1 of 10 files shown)


kylebrandt:
This addresses https://github.com/bosun-monitor/bosun/issues/1026

docs/expressions.md
```diff
@@ -331,55 +331,55 @@ is also returned with a value of `1`. Primarily for use with `depends`.
...
 
 Returns the first key from the given lookup table with matching tags.
 
-## nv(number, scalar)
+## nv(numberSet, scalar) numberSet
 
 Change the NaN value during binary operations (when joining two queries) of unknown groups to the scalar. This is useful to prevent unknown group and other errors from bubbling up.
 
-## sort(number, asc_desc)
+## sort(numberSet, asc_desc string) numberSet
```
kylebrandt: Should we have a format when strings have a predefined set of values. So for sort the asc desc, lsstat is has a set as well.

captncraig: I'm not sure how to define it. The description makes it pretty clear what to do.

gbrayut: something like regex syntax: \"(asc|desc)"\

or Powershell validation sets: [ValidateSet("Steve","Mary")]

gbrayut: And I disagree that the description is clear. You should be able to call a function knowing only it's signature, and asc_desc is not a defined type. If you want to do that you should do what .NET and Go do and use enums or consts instead of strings

kylebrandt: I like the Piped version (Minus the escapes and quotes, so

`(asc|desc) string`

kylebrandt:
@captncraig Looking good - much appreciated. Ship at will

gbrayut:
Agreed... looks much better! Great work!

PR #1078, opened 2015-06-18
cmd/scollector: don't use time var just for an error return

mjibson:
[Review on Reviewable](https://reviewable.io/reviews/bosun-monitor/bosun/1078)


gbrayut:
Much nicer... I new there had to be a way to get a default instance but just wasn't sure how. LGTM

PR #1081, opened 2015-06-18
Build improvements

mjibson:
[Review on Reviewable](https://reviewable.io/reviews/bosun-monitor/bosun/1081)

### DIFF
--- a/build/build.go
+++ b/build/build.go
@@ -12,8 +12,15 @@ import (
 	"time"
 )
 
+var (
+	shaFlag         = flag.String("sha", "", "SHA to embed.")
+	buildBosun      = flag.Bool("bosun", false, "Only build Bosun.")
+	buildScollector = flag.Bool("scollector", false, "Only build scollector.")
+
+	allProgs = []string{"bosun", "scollector"}
+)
+
 func main() {
-	shaFlag := flag.String("sha", "", "Sha to embed")
 	flag.Parse()
 	// Get current commit SHA
 	sha := *shaFlag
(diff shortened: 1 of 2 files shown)


captncraig:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[version/version.go, line 29 \[r1\]](https://reviewable.io:443/reviews/bosun-monitor/bosun/1081#-Js6reHrB4mTATtWfIf9)** ([raw file](https://github.com/bosun-monitor/bosun/blob/d593714622e88aaa9dd9e61adb6bf2e4a8a50eab/version/version.go#L29)):</sup>
One issue here is that it get's `-dev` added each time you call this. It is ok if we are just running -version, but if we wanted to show it on the web app or something it would be bad.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/bosun-monitor/bosun/1081)

mjibson:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[version/version.go, line 29 \[r1\]](https://reviewable.io:443/reviews/bosun-monitor/bosun/1081#-Js6reHrB4mTATtWfIf9)** ([raw file](https://github.com/bosun-monitor/bosun/blob/d593714622e88aaa9dd9e61adb6bf2e4a8a50eab/version/version.go#L29)):</sup>
If the webapp were called this function twice it would add `-dev` each time, which is wrong. This function had a side effect that changed a global variable. At least it's predictable now.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/bosun-monitor/bosun/1081)

gbrayut:
<sup>**[version/version.go, line 29 \[r1\]](https://reviewable.io:443/reviews/bosun-monitor/bosun/1081#-Js6reHrB4mTATtWfIf9)** ([raw file](https://github.com/bosun-monitor/bosun/blob/d593714622e88aaa9dd9e61adb6bf2e4a8a50eab/version/version.go#L29)):</sup>
version is a copy of Version, so you will only get -dev once. Confirmed by calling fmt.Println(version.GetVersionInfo("scollector")) twice and the output was correct:

scollector version 0.2.0-dev (deef5c5ead4a93fc79203142f47ce2694b82bf59) built 2015-06-18T18:22:27Z
scollector version 0.2.0-dev (deef5c5ead4a93fc79203142f47ce2694b82bf59) built 2015-06-18T18:22:27Z

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/bosun-monitor/bosun/1081)

gbrayut:
LGTM.  C:\Go\bin\go.exe run .\build\build.go -scollector works and the scollector.version metric works as expected

PR #1082, opened 2015-06-18
Build package sh

mjibson:
[Review on Reviewable](https://reviewable.io/reviews/bosun-monitor/bosun/1082)

### DIFF
--- a/build.scollector.sh
+++ b/build.scollector.sh
@@ -1,0 +1,3 @@
+#!/bin/sh
+
+go run build/build.go -scollector -output cmd/scollector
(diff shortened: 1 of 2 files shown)


build.scollector.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
```
captncraig: would it be preferable to have this in the build directory?

mjibson: Yes, but this is triggered by our (StackExchange's) build scripts, which require it to be at the project root.

PR #1087, opened 2015-06-19
cmd/scollector: process_linux process start_time and uptime

dimamedvedev:
Calculate process start time and uptime from file stat data of /proc/<pid> directory.

P.S. Chose this because it's way simpler than querying uptime, `getconf CLK_TCK` , parsing stat and jiffies. There were times when `/proc/<pid>` stat data was modifiable but that's fixed now.

[Review on Reviewable](https://reviewable.io/reviews/bosun-monitor/bosun/1087)

### DIFF
--- a/cmd/scollector/collectors/processes_linux.go
+++ b/cmd/scollector/collectors/processes_linux.go
@@ -39,6 +39,11 @@ func WatchProcesses() {
 func linuxProcMonitor(w *WatchedProc, md *opentsdb.MultiDataPoint) error {
 	var err error
 	for pid, id := range w.Processes {
+		file_status, e := os.Stat("/proc/" + pid)
+		if e != nil {
+			w.Remove(pid)
+			continue
+		}
 		stats_file, e := ioutil.ReadFile("/proc/" + pid + "/stat")
 		if e != nil {
 			w.Remove(pid)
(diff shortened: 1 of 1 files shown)


cmd/scollector/collectors/processes_linux.go
```diff
@@ -104,6 +110,8 @@ func linuxProcMonitor(w *WatchedProc, md *opentsdb.MultiDataPoint) error {
 		Add(md, "linux.proc.io_bytes", io[4], opentsdb.TagSet{"type": "read"}.Merge(tags), metadata.Counter, metadata.Bytes, descLinuxProcIoBytesRead)
 		Add(md, "linux.proc.io_bytes", io[5], opentsdb.TagSet{"type": "write"}.Merge(tags), metadata.Counter, metadata.Bytes, descLinuxProcIoBytesWrite)
 		Add(md, "linux.proc.num_fds", len(fds), tags, metadata.Gauge, metadata.Files, descLinuxProcFd)
+		Add(md, "linux.proc.start_time", start_ts, tags, metadata.Gauge, metadata.Timestamp, descLinuxProcStartTS)
+		Add(md, "linux.proc.uptime", timestamp-start_ts, tags, metadata.Gauge, metadata.Second, descLinuxProcUptime)
```
mjibson: This is not the uptime of the program. It is the uptime of the program since scollector has been running. I don't see the value of this metric, since one can always just subtract the start time and a timestamp to figure out uptime.

mjibson: Nope, I'm wrong. timestamp is updated to be now. But use `now()` to get it because it's race-free.

cmd/scollector/collectors/processes_linux.go
```diff
@@ -91,6 +97,8 @@ func linuxProcMonitor(w *WatchedProc, md *opentsdb.MultiDataPoint) error {
 				}
 			}
 		}
+		start_ts := file_status.ModTime().Unix()
+		now := time.Now().Unix()
```
mjibson: Just use `now()` for this: https://github.com/bosun-monitor/bosun/blob/3e9002eba1699fe4db468ae4c7cc0aee06e96842/cmd/scollector/collectors/collectors.go#L107

PR #1097, opened 2015-06-23
Add User-Agent to HTTP notifications

inkel:
Use a custom `http.Client` for sending HTTP notifications, adding a proper `User-Agent` header with the `Bosun/{{version}}` format. The version only includes the version number and release tag, not the most recent commit hash nor build time.

[Review on Reviewable](https://reviewable.io/reviews/bosun-monitor/bosun/1097)

### DIFF
--- a/cmd/bosun/main.go
+++ b/cmd/bosun/main.go
@@ -30,12 +30,25 @@ import (
 	"bosun.org/version"
 )
 
+type bosunHttpTransport struct {
+	UserAgent string
+	http.RoundTripper
+}
+
+func (t *bosunHttpTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	req.Header.Add("User-Agent", t.UserAgent)
+	return t.RoundTripper.RoundTrip(req)
+}
+
 func init() {
 	client := &http.Client{
-		Transport: &httpcontrol.Transport{
-			Proxy:          http.ProxyFromEnvironment,
-			RequestTimeout: time.Minute,
-			MaxTries:       3,
+		Transport: &bosunHttpTransport{
+			"Bosun/" + version.ShortVersion(),
+			&httpcontrol.Transport{
+				Proxy:          http.ProxyFromEnvironment,
+				RequestTimeout: time.Minute,
+				MaxTries:       3,
+			},
 		},
 	}
 	http.DefaultClient = client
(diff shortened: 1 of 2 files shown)


inkel:
This change only covers adding the `User-Agent` header. The HTTP notification code could use a refactor to removing some duplication in the code.

The `version` package could also be refactored to include either a method returning the value for the `User-Agent` instead of having to build the `userAgent` variable in the `init` function.

captncraig:
We already set http.DefaultClient to a custom client here: https://github.com/bosun-monitor/bosun/blob/080bbf1f1d27077306e3ad84603fd1aac48f37d2/cmd/bosun/main.go#L34 

Any changes should be made there. Adding a "ShortVersion()" function to the version package would be ok by me as well.

inkel:
@captncraig ah, great, I'll use `http.DefaultClient` then.  Sadly, custom headers cannot be set in the default client but only in `Request.Header`, unless I'm missing something. I'll double check.

I'll work on the `ShortVersion` function too.

captncraig:
I would think we should set up out main client to set the user-agent on every request. That can be done with a custom RoundTripper. Something like this: https://play.golang.org/p/_E6fRYPvrN

inkel:
@captncraig I've applied your suggestions. I think the initialization of the client could probably be done in a nicer way than this. I'd like to know your thoughts.

captncraig:
LGTM. How do you feel about this implementation @mjibson ?

cmd/bosun/main.go
```diff
@@ -30,12 +30,25 @@ import (
 	"bosun.org/version"
 )
 
+type bosunHttpTransport struct {
+	UserAgent string
+	http.RoundTripper
+}
+
+func (t *bosunHttpTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	req.Header.Add("User-Agent", t.UserAgent)
```
mjibson: Do not set this if it already exists.

mjibson:
Implementation looks good after my one comment. What problem does this solve, or why is it needed?

inkel:
> What problem does this solve, or why is it needed?

It isn't really a problem what is solving, maybe a best practice. At work we're working on a custom notifications aggregator, which several other services sending HTTP notifications. We record all the information received and use the `User-Agent` header to identify which service was the one sending the notification. I'm aware HTTP headers are easily spoofed, but this will mostly internal, so it's not a concern. And in our case, using the `User-Agent` for identification is easier than checking the remote IP, for instance.

I'll see to fix that line you pointed me, sorry for the delay. Thanks!

captncraig:
Closing. Will merge commit in #1265

inkel:
Thanks!

Issue #1096, opened 2015-06-23
Use templates body as payload for notifications and subject for other HTML related stuff


ayashjorden:
Hi all, as described in the docs, I'm using the templates subject as body for POSTing stuff to our hipchat bot. the problem I encounter is in Bosun main view (list of alerts) where the template subject is presented when clicking an alert for details.

![image](https://cloud.githubusercontent.com/assets/4204371/8315624/aae096e2-19fa-11e5-9708-49dbb6dfe8d4.png)

Suggested is to use templates' body as payload for notification (POST notifications mainly). a flag can be also added to let the user which template will use the subject as payload and which will use the body.

Thanks,
Yarden

captncraig:
So in notification we could specify 

    post=http://hipchat
    useBody = true

I kinda like it.

ayashjorden:
Yup, thats what I thought.
Is it still large refactor as @mjibson said ?

captncraig:
no, that alone should be pretty easy. A small addition to conf, and a single check when we send the post. You could probably handle that.

ayashjorden:
Were would it be in the code?

captncraig:
The posts actually get dispatched around https://github.com/bosun-monitor/bosun/blob/2454db52018ff3d34c84b5f0e9544c77dc96b096/cmd/bosun/conf/notify.go#L31

ayashjorden:
Added UseBody boolean, it works.
Now we're left with two things:
1. Action page, 'Send Notification' message payload is also using template body, I'd like to override it so it will use templates' subject anyway (when taking action on multiple alerts, we need to list only the template subject and not the json message in the body).
2. Bosun main page, let the alert list use the template's subject.

Please point me in the direction for solving the above.
Thanks,
Yarden

mjibson:
We should also consider changing `.` to instead have `.Body` and `.Subject`. I prefer that method instead of `useBody` because it allows a user to use both if they wish, which is certainly possible.

captncraig:
+1 to matt's idea

ayashjorden:
+1 to @mjibson suggestion

ayashjorden:
@mjibson take a look on this one:
![image](https://cloud.githubusercontent.com/assets/4204371/8347997/0be3cb1e-1b16-11e5-9026-062963848f9a.png)

[...]

PR #1101, opened 2015-06-24
1096 notification use body switch

ayashjorden:
Hi,
As discussed with @captncraig, and in [issue 1096](https://github.com/bosun-monitor/bosun/issues/1096):
1. When taking action on several alerts, list alert names in bulk message passed to notifiaction.
2. let ```.``` have the ```.Subject``` and ```.Body``` (as per @mjibson idea)
3. Bosun main page, let the alert list use the template's subject.

[Review on Reviewable](https://reviewable.io/reviews/bosun-monitor/bosun/1101)


cmd/bosun/conf/conf.go
```diff
@@ -297,11 +297,14 @@ type Notification struct {
...
 	ContentType  string
 	RunOnActions bool
+	UseBody      bool
 
 	next      string
 	email     string
 	post, get string
 	body      string
+	// Maybe not used, need to learn more GO :(
+	useBody bool
```
captncraig: This is indeed not used.

cmd/bosun/conf/notify.go
```diff
@@ -29,30 +29,42 @@ func (n *Notification) Notify(subject, body string, emailsubject, emailbody []by
...
-		go n.DoPrint(subject)
+		if n.UseBody {
+			go n.DoPrint("Subject: " + subject + ", Body: " + body)
+		} else {
+			go n.DoPrint(subject)
+		}
+	}
+}
+
+func (n *Notification) GetPayload(subject, body string) (payload string) {
```
captncraig: I'd have this function return `[]byte` instead of `string` so you don't need to convert it at the call.

cmd/bosun/conf/conf.go
```diff
@@ -297,11 +297,14 @@ type Notification struct {
 	Timeout      time.Duration
 	ContentType  string
 	RunOnActions bool
+	UseBody      bool
```
captncraig: I think @mjibson means to remove this altogether, and make no conf changes. Just change the data object passed into the body template here:https://github.com/bosun-monitor/bosun/blob/master/cmd/bosun/conf/notify.go#L49 to have both the subject and body available. 

This will be a breaking change, but I think it will be more flexible that way.

ayashjorden:
Hi @captncraig,
If we're going to make template's body available for notification body, will the template's body still be rendered as HTML? I once tried using the template's body to render json and got the json object wrapped with HTML tags.

My pure intention with this PR and issue 1096 was to have a clean template subject and use the template's body to render a json object to be sent to our Hipchat bot.

I've read the docs about 'json' function available in notification body template,(assuming we'll add the template's body to be available in notification body) it is unclear if the template's body will be rendered as HTML and then passed to notification body or notification body will take the template's body as plain text and render it.

ayashjorden:
Running validate locally exited with code 0.

captncraig:
@ayashjorden sorry this has sat so long. If you can rebase to master and resolve conflicts I can probably merge it. Something about it feels a bit off, but I think it is an ok solution for now.

mjibson:
Reviewed 4 of 4 files at r1, 3 of 3 files at r3.
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/bosun-monitor/bosun/1101)*

ayashjorden:
@captncraig,
I've updated my branch with latest from bosun.org/bosun:master.

I'd be happy to understand why this solution is a 'bit off' from your point of view...

Thank you,
Yarden

ayashjorden:
@captncraig @kylebrandt ,
*defrati* referred me to an issue I mentioned earlier(> month), but we didn't conclude on.
The template's body is rendered as HTML, thus that the JSON string is wrapped `<html><head></head><body>JSON_STRING_HERE</body></html>` .

To overcome this, I've been using a regex (`"<html><head><\/head><body>(.*)<\/body><\/html>`) to extract the JSON string and parse in the consumer(e.g. Will bot).

What do you think?
Yarden

deanefrati:
i would really love to see an option to encode the body as JSON natively. I'm using the same workaround @ayashjorden is using right now but it's not the prettiest way to handle this... Also the ack and close messages are HTML encoded, it would be great if those can also have an option to JSON encode. This will give us a lot of flexibility for alert processing in external tools, for example we are looking at using stackstorm to receive the alerts from Bosun and run some automations based on the alert fields.

PR #1105, opened 2015-06-26
cmd/bosun: convert absolute to relative times for graph exprs

### LINKED ISSUE #127: Expression links from graph page should always be relative time
Even though the time fields in the graph page might be absolute time. The expression link should convert that to relative time since our query functions only support relative time. Or we could let our query functions support absolute time. I think I like the latter

mjibson:
fixes #127


captncraig:
LGTM

PR #1107, opened 2015-06-26
cmd/scollector: scollector.* metrics disabled properly with DisableSelf

### LINKED ISSUE #1106: scollector: `scollector.collector.*` metrics not properly disabled.
Even if DisableSelf is true these metrics still get sent:

https://github.com/bosun-monitor/bosun/blob/master/cmd/scollector/collectors/interval.go#L64

captncraig:
Fixes #1106

### DIFF
--- a/cmd/scollector/collectors/interval.go
+++ b/cmd/scollector/collectors/interval.go
@@ -10,6 +10,7 @@ import (
 	"sync"
 	"time"
 
+	"bosun.org/collect"
 	"bosun.org/metadata"
 	"bosun.org/opentsdb"
 	"bosun.org/slog"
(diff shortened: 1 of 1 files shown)


cmd/scollector/collectors/interval.go
```diff
@@ -61,8 +63,11 @@ func (c *IntervalCollector) Run(dpchan chan<- *opentsdb.DataPoint) {
 				slog.Errorf("%v: %v", c.Name(), err)
 				result = 1
 			}
-			Add(&md, "scollector.collector.duration", timeFinish.Seconds(), tags, metadata.Gauge, metadata.Second, "Duration in seconds for each collector run.")
-			Add(&md, "scollector.collector.error", result, tags, metadata.Gauge, metadata.Ok, "Status of collector run. 1=Error, 0=Success.")
+			if !DisableSelf {
```
mjibson: What about using `collect.DisableDefaultCollectors` instead of adding a new variable? Your choice, just a thought. Otherwise LGTM.

PR #1110, opened 2015-06-28
cmd/scollector: Order process identifiers by process creation timestamp.

dimamedvedev:
Sort processes by creation time so lower ids are assigned to older processes. Also helps preserve process ids in case scollector restarts.

P.S. Could be made a bit more effective, but complicates code more than necessary.


mjibson:
LGTM. Travis CI build failure is not related to this change.

PR #1111, opened 2015-06-29
Improve silence page

mjibson:


mjibson:
Could you achieve the same goal by setting the user field to "bot"?

kylebrandt:
We will likely have multiple bots, and we will probably want to distinguish
between them.

mjibson:
Well then each can set their own username. My point is that I'm not completely convinced this solution (an extra isBot boolean) is the best solution for the problem. If we can use an existing field to achieve the same goal, will that work?

gbrayut:
+1 on using an existing field. We can use a naming prefix for bots, and add search with globs if we want filtering

kylebrandt:
The main reason I don't like this idea is that is a convention that needs
to be enforced by humans, and as someone programming or querying the API I
have to no reason to think I should be doing anything like that.

gbrayut:
We've been down this road before though with isEmail/isPost/isDashboard/isWhatever in rendering alert templates. This method has a lot more flexibility and no additional coding/maintenance. We already have a lot of conventions, and if a human does something wrong we can address that. Or if someone else wants to change how it works they can come up with their own convention.

If you really want an isBot, that should be an attribute on an actual user profile/record, which is something we can address when adding authentication and user management. In my mind it doesn't make much sense to tack it on to the silence item/event.

mjibson:
Anyone object to merging this now? We can do further improvements in another PR.

gbrayut:
No objections here. I'm sure we'll have more improvements to that page soon.

PR #1112, opened 2015-06-29
cmd/bosun: download entire alerts dashboard at once

### LINKED ISSUE #1094: Improve dashboard performance
Don't make N requests per alertkey now. Needed before because of computations. Now it's probably fine to do them all at once.

mjibson:
Previously, only a summary of each alert was downloaded. This was because
the entire struct was large (hundreds of megs) due to computations. Since
we have removed computations from the dashboard, we can now again download
the entire struct. The bug this new method prevents is that if an alert has
changed since the dashboard summary was downloaded, it will no longer show
up in the wrong place. The dashboard is now faster because it does not
download each alert when a group is expanded.

Testing with our production state file, this route went from 6K to 900K. A
substantial increase, but still small and fast enough to be fine.

fixes #1094


captncraig:
LGTM. What is wrong with the travis build?

PR #1113, opened 2015-06-30
bosun: Adding instrumentation around schedule lock.

captncraig:
Here we emit metrics about how long a method spent waiting for a lock, as well as how long it held it.

I suspect the slowness on the dashboard is related to lock contention, and this should help diagnose it.

### DIFF
--- a/cmd/bosun/sched/sched.go
+++ b/cmd/bosun/sched/sched.go
@@ -29,7 +29,10 @@ func init() {
 }
 
 type Schedule struct {
-	sync.Mutex
+	mutex         sync.Mutex
+	mutexHolder   string
+	mutexAquired  time.Time
+	mutexWaitTime int64
 
 	Conf          *conf.Conf
 	status        States
@@ -49,10 +52,30 @@ type Schedule struct {
 	db            *bolt.DB
 }
 
-func (s *Schedule) TimeLock(t miniprofiler.Timer) {
-	t.Step("lock", func(t miniprofiler.Timer) {
-		s.Lock()
-	})
+func init() {
+	metadata.AddMetricMeta(
+		"bosun.schedule.lock_time", metadata.Counter, metadata.MilliSecond,
+		"Length of time spent waiting for or holding the schedule lock.")
+	metadata.AddMetricMeta(
+		"bosun.schedule.lock_count", metadata.Counter, metadata.Count,
+		"Number of times the given caller acquired the lock.")
+}
+
+func (s *Schedule) Lock(method string) {
+	start := time.Now()
+	s.mutex.Lock()
+	s.mutexAquired = time.Now()
+	s.mutexWaitTime = int64(s.mutexAquired.Sub(start) / time.Millisecond) // remember this so we don't have to call put until we leave the critical section.
+}
+
+func (s *Schedule) Unlock() {
+	holder := s.mutexHolder
+	start := s.mutexAquired
+	waitTime := s.mutexWaitTime
+	s.mutex.Unlock()
+	collect.Add("bosun.schedule.lock_time", opentsdb.TagSet{"caller": holder, op: "wait"}, waitTime)
+	collect.Add("bosun.schedule.lock_time", opentsdb.TagSet{"caller": holder, op: "hold"}, int64(time.Since(start)/time.Millisecond))
+	collect.Add("bosun.schedule.lock_count", opentsdb.TagSet{"caller": holder}, 1)
 }
 
 type Metavalue struct {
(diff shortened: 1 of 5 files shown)


cmd/bosun/sched/sched.go
```diff
@@ -49,10 +52,29 @@ type Schedule struct {
...
+	s.mutexWaitTime = s.mutexAquired.Sub(start) / time.Millisecond // remember this so we don't have to call put until we leave the critical section.
+}
+
+func (s *Schedule) Unlock() {
+	holder := s.mutexHolder
+	start := s.mutexAquired
+	waitTime := s.mutexWaitTime
+	s.mutex.Unlock()
+	collect.Put("bosun.schedule.lock.wait", opentsdb.TagSet{"caller": holder}, waitTime)
+	collect.Put("bosun.schedule.lock.held", opentsdb.TagSet{"caller": holder}, time.Now().Sub(start)/time.Millisecond)
```
mjibson: `time.Now().Sub(start)` -> `time.Since(start)`

cmd/bosun/sched/sched.go
```diff
@@ -49,10 +52,29 @@ type Schedule struct {
...
+	s.mutexAquired = time.Now()
+	s.mutexWaitTime = s.mutexAquired.Sub(start) / time.Millisecond // remember this so we don't have to call put until we leave the critical section.
+}
+
+func (s *Schedule) Unlock() {
+	holder := s.mutexHolder
+	start := s.mutexAquired
+	waitTime := s.mutexWaitTime
+	s.mutex.Unlock()
+	collect.Put("bosun.schedule.lock.wait", opentsdb.TagSet{"caller": holder}, waitTime)
```
mjibson: Instead of only recording the last event, how about making this a counter, so we just increase the total time. We would then also add another metric that is the count of locks acquired by a caller. This would allow us to collect data about the average lock hold time for each caller and find calls that are outside the standard deviation. It would also prevent missing any of the slow calls, which Put may certainly do.

cmd/bosun/sched/sched.go
```diff
@@ -49,10 +52,31 @@ type Schedule struct {
...
-		s.Lock()
-	})
+func init() {
+	metadata.AddMetricMeta(
+		"bosun.schedule.lock_time", metadata.Counter, metadata.MilliSecond,
+		"The length of time spent waiting for, or holding the schedule lock.")
+	metadata.AddMetricMeta(
+		"bosun.schedule.lock_count", metadata.Counter, metadata.Count,
+		"The number of times the given caller acquired the lock")
+
```
mjibson: no extra line

cmd/bosun/sched/sched.go
```diff
@@ -259,11 +282,16 @@ type StateGroups struct {
 }
 
 func (s *Schedule) MarshalGroups(T miniprofiler.Timer, filter string) (*StateGroups, error) {
-	silenced := s.Silenced()
```
mjibson: do these changes need to be part of this PR?

cmd/bosun/sched/sched.go
```diff
@@ -49,10 +52,30 @@ type Schedule struct {
...
-	t.Step("lock", func(t miniprofiler.Timer) {
-		s.Lock()
-	})
+func init() {
+	metadata.AddMetricMeta(
+		"bosun.schedule.lock_time", metadata.Counter, metadata.MilliSecond,
+		"The length of time spent waiting for, or holding the schedule lock.")
+	metadata.AddMetricMeta(
+		"bosun.schedule.lock_count", metadata.Counter, metadata.Count,
+		"The number of times the given caller acquired the lock")
```
mjibson: period at EOL

cmd/bosun/sched/sched.go
captncraig: Not strictly speaking. Removing timelock made sense since every lock is timelock now. Adding a step around silence was useful because I suspect that is where most of the time is spent. This whole thing is really to diagnose why this endpoint is getting locked.

cmd/bosun/sched/sched.go
```diff
@@ -49,10 +52,30 @@ type Schedule struct {
...
 }
 
-func (s *Schedule) TimeLock(t miniprofiler.Timer) {
-	t.Step("lock", func(t miniprofiler.Timer) {
-		s.Lock()
-	})
+func init() {
+	metadata.AddMetricMeta(
+		"bosun.schedule.lock_time", metadata.Counter, metadata.MilliSecond,
+		"The length of time spent waiting for, or holding the schedule lock.")
```
mjibson: Can we omit "The" at the beginning of these sentences? Most of the other descs I see don't have it? And remove the comma. Or add another one after "holding".

cmd/bosun/sched/sched.go
mjibson: Ok. Keep it. Just making sure.

mjibson:
LGTM if you squish it.

PR #1126, opened 2015-07-01
cmd/bosun: shorten the action page URLs

### LINKED ISSUE #1053: Actions will fail if you are performing them on many alerts
Just like the https://github.com/bosun-monitor/bosun/issues/1033 , actions will fail if you are performing them on X alerts.

mjibson:
This prevents proxies from refusing to relay because the HTTP header is
too large. This is done by reducing the length of the requesting page's
URL. But since bosun communicates data between controllers via the URL, we
now introduce a new key/value store to perform that exchange.

The other implementations for this communication aren't sibling-sibling,
they are parent-child. I didn't find a good way to do sibling-sibling
communication, and this is pretty clear, so I think it's a good choice.

fixes #1053


cmd/bosun/web/static/js/action.ts
```diff
@@ -12,10 +12,15 @@ bosunControllers.controller('ActionCtrl', ['$scope', '$http', '$location', '$rou
...
 	$scope.notify = true;
-	if (!angular.isArray(search.key)) {
-		$scope.keys = [search.key];
+	if (search.key) {
+		var keys = search.key;
+		if (!angular.isArray(search.key)) {
+			keys = [search.key];
+		}
+		$location.search('key', null);
+		$scope.setKey('action-keys', keys);
```
captncraig: Do we need to set $scope.keys in this block?

mjibson: Nope. The changing of $location.search above changes the URL which re-renders this controller.

captncraig: gotcha. lgtm

PR #1129, opened 2015-07-01
cmd/scollector: Increase default batch size to 500

gbrayut:
and add BatchSize option to toml file


mjibson:
LGTM.

I have also noticed we need to document the defaults of the options. That can happen in another PR.

Issue #1064, opened 2015-06-11
It would be useful it forecastlr could also take dynamic (type number) y_val argument


kylebrandt:
Currently we use it on percentage based data. So it is simple to specify (100%) or some other hard coded value.

However there are situations where current values and limits are both pieces of data provided in the time series (HAProxy session limits, File descriptor limits, etc). If they y_val argument could be of type Number and not just Scalar we could create this sort of instantiated dynamic forecasting without having to go back and send the data itself as a percentage when creating the alert.

kylebrandt:
This has been fixed

PR #1137, opened 2015-07-02
cmd/bosun: promote scalars to numberSets

mjibson:


mjibson:
Don't merge yet: a few more functions will get promotion. But ready to review.

mjibson:
Other functions (des, dropge, drople, limit, nv) can also benefit from scalar promotion. But since they are not reduction functions we'll need another way to do the tag matching. They may come in a later PR. At least we get forecastlr and percentile now.

kylebrandt:
drop__ could add a fair amount of power I think, something like:

```
$series = q("...")
$limit = last(q("..."))
$filteredSeries = dropge($series, percentille($series, .95))
$daysLeft = forecastlr($filteredSeries, $limit) / d("1d")  
``` 

Unless there is another way to do this...

mjibson:
Yes, they will happen. There's just no obvious common way to do them all now, since they don't have the same function signature.

mjibson:
Ok, now does the drop functions.

kylebrandt:
docs need to be updated to reflect the drop funcs

mjibson:
Updated with dropg, dropl, and docs.

kylebrandt:
Will fix https://github.com/bosun-monitor/bosun/issues/1064

cmd/bosun/expr/expr_test.go
```diff
@@ -224,3 +224,31 @@ func TestQueryExpr(t *testing.T) {
 		}
 	}
 }
+
+func TestScalarPromotion(t *testing.T) {
+	tests := map[string]map[string]Value{
```
captncraig: this is a really confusing data structure to only hold one test. Either add more or inline the fields (or make an anonymous struct for test cases).

cmd/bosun/expr/expr_test.go
```diff
@@ -224,3 +224,31 @@ func TestQueryExpr(t *testing.T) {
...
+		`abs(-1)`: {"": Number(1)},
+	}
+
+	for exprText, expected := range tests {
+		e, err := New(exprText)
+		if err != nil {
+			t.Fatal(err)
+		}
+		results, _, err := e.Execute(nil, nil, nil, nil, nil, queryTime, 0, false, nil, nil, nil)
```
captncraig: can this exercise walkfunc directly?

cmd/bosun/expr/expr.go
```diff
@@ -594,6 +594,15 @@ func (e *State) walkFunc(node *parse.FuncNode, T miniprofiler.Timer) *Results {
 			default:
 				panic(fmt.Errorf("expr: unknown func arg type"))
 			}
+			if f, ok := v.(float64); ok && node.F.Args[i] == parse.TypeNumberSet {
+				v = &Results{
```
captncraig: should this use the fromScalar helper?

cmd/bosun/expr/expr_test.go
mjibson: I'm not convinced that's a great test. The scalar promotion stuff does things with groups, and I want to make sure it works at the top level.

mjibson:
Updated with better tests - now combined with the existing expr tests, it just knows about types now.

Issue #408, opened 2014-11-19
Combine two metrics from linux.cpu


lukaspustina:
Bosun is an awesome tool. Thanks for building it.

I'm trying to combine two metrics similar to the HAproxy example, but I'm failing miserably:
`avg(q("sum:5m-sum:rate{counter,,1}:linux.cpu{type=idle}", "1d", "")) + avg(q("sum:5m-sum:rate{counter,,1}:linux.cpu{type=user}", "1d", ""))`

Trying to do so and evaluating the expression, I receive:
`unjoined group (NaN)`

Any ideas how I can fix that?

PR #1138, opened 2015-07-03
cmd/scollector: Add linux.percent_cpu for easier reporting

pdf:
We've found iowait and steal percentages to be useful alert metrics

### DIFF
--- a/cmd/scollector/collectors/procstats_linux.go
+++ b/cmd/scollector/collectors/procstats_linux.go
@@ -114,6 +114,7 @@ func c_procstats_linux() (opentsdb.MultiDataPoint, error) {
 			return nil
 		}
 		if strings.HasPrefix(m[1], "cpu") {
+			total_cpu := 0.0
 			metric_percpu := ""
 			tag_cpu := ""
 			cpu_m := statCPURE.FindStringSubmatch(m[1])
@@ -123,6 +124,13 @@ func c_procstats_linux() (opentsdb.MultiDataPoint, error) {
 				tag_cpu = cpu_m[1]
 			}
 			fields := strings.Fields(m[2])
+			for _, value := range fields {
+				vf, err := strconv.ParseFloat(value, 64)
+				if err != nil {
+					return nil
+				}
+				total_cpu += vf
+			}
 			for i, value := range fields {
 				if i >= len(CPU_FIELDS) {
 					break
@@ -134,6 +142,12 @@ func c_procstats_linux() (opentsdb.MultiDataPoint, error) {
 					tags["cpu"] = tag_cpu
 				}
 				Add(&md, "linux.cpu"+metric_percpu, value, tags, metadata.Counter, metadata.CHz, cpu_stat_desc[CPU_FIELDS[i]])
+
+				vf, err := strconv.ParseFloat(value, 64)
+				if err != nil {
+					return nil
+				}
+				Add(&md, "linux.percent_cpu"+metric_percpu, vf/total_cpu*100.0, tags, metadata.Counter, metadata.Pct, cpu_stat_desc[CPU_FIELDS[i]])
 			}
 			if metric_percpu == "" {
 				if len(fields) < 3 {


cmd/scollector/collectors/procstats_linux.go
```diff
@@ -134,6 +142,12 @@ func c_procstats_linux() (opentsdb.MultiDataPoint, error) {
 					tags["cpu"] = tag_cpu
 				}
 				Add(&md, "linux.cpu"+metric_percpu, value, tags, metadata.Counter, metadata.CHz, cpu_stat_desc[CPU_FIELDS[i]])
+
```
kylebrandt: Remove blank line.

cmd/scollector/collectors/procstats_linux.go
```diff
@@ -134,6 +142,12 @@ func c_procstats_linux() (opentsdb.MultiDataPoint, error) {
...
+				vf, err := strconv.ParseFloat(value, 64)
+				if err != nil {
+					return nil
+				}
+				Add(&md, "linux.percent_cpu"+metric_percpu, vf/total_cpu*100.0, tags, metadata.Counter, metadata.Pct, cpu_stat_desc[CPU_FIELDS[i]])
```
kylebrandt: I think the description of these should explain the difference between The metric on 144 and this one linux.cpu"+metric_percpu

mjibson: @dimamedvedev points out that this should be a gauge

kylebrandt: Is this just like "linux.cpu"+metric_percpu but a gauge instead? If so how is that useful?

pdf: Making the counter values useful would mean performing these calculations on client side, and I'm struggling to see how you would actually get there, but it's entirely possible I'm missing something?

kylebrandt: OpenTSDB does this confession at query time, what time series database are
you using?

kylebrandt: *conversion

pdf: So, with OpenTSDB, if I was to query a single tag value, it would convert that to a percentage of the sum of all tag values somehow?  I'm using InfluxDB.

pdf: Actually, I realise now that this is not going to provide what I'm looking for, since stats are since boot, sorry for the noise.

pdf:
@kylebrandt, thanks for pointing out I was being stupid here, I really failed to grasp what I was looking at.

I'm still struggling to find a way to interpret the current data usefully though.  I can't work with it as a time value, because the rate of change would need to be scaled by the number of CPUs in the host (so, different for each host, difficult to generalize) before it could be compared to real time, and I don't have access to functions that can produce the rate of change a percentage relative to sum the other values in a single query.  I don't believe this can be achieved with TSDB either unless I'm mistaken?  If you're using these values, are you only doing so by processing two series programmatically?

I see in #408 you'd mentioned a desire to convert this to a percentage, did you have a concept of how this might be achieved that I might attempt to implement?  I can't think of a way that doesn't require access to the previous measurement's value.

kylebrandt:
@pdf I'm having a little trouble following you. But if you want these breakdowns as a 0-100% range than with OpenTSDB you could do the following: 

image

The "Series Type" is going to be a Counter on this one (It is set to auto because scollector tells bosun via metadata it is, and this graphing page knows what to do with that.) What is of particular interest if I understand you is that "Aggregator". That says "Average all the values for any tag that didn't have a value or `*` specified. In other words, average each type of utilization over all cores per host. I would guess Influx can do the same, but I am not familiar with it yet.

If this doesn't help, maybe write out the alert you are trying to create in english as best you can.

pdf:
Hrm, I'm not entirely certain I understand how that query returns values from 0 to 100.  I understand why you'd take the average of the per-cpu stats... I suppose this relies on the kernel having HZ == 100, and on the rate being calculating per second?  Thanks for taking the time btw @kylebrandt, I appreciate it.

I just want to graph/alert on %iowait/%steal/etc above a specified percentage per host per interval.  If I could get the same values out of InfluxDB that you're getting there, that'd probably do the job, ignoring the possibility of odd/old kernels configured for different HZ.

Unfortunately, I'm not having much luck there - I'm not sure there's a combination of functions at the moment that will produce the same output you're able to get from OpenTSDB - derivative(rate) and mean don't appear to interact in the same way.  Perhaps I'll switch since I'm only in the early stages, but the deployment story for OpenTSDB kind of sucks by comparison :confused:

pdf:
Just FYI, (one of) the problem(s) with InfluxDB is that it doesn't do interpolation by default, so you can't group by an interval smaller than your sampling interval.  It does support manual interpolation, but does not support interpolating mean values, only fixed numbers, previous value, or null, which makes it pretty useless for rate calculation.

lukaspustina:
@kylebrandt can you comment on why calculating an average produces values in the range of 0 ... 100? 

Performing the same query on `linux.cpu` produces the "expected" average in CentiHz.

pdf:
@lukaspustina I'm pretty sure I understand this now, and it's basically luck - a confluence of defaults.  The CPU stats count jiffies spent per task type, the duration of a jiffy depends on the value of the kernel HZ constant.  Many kernels happen to default to a HZ value of 100, so in that case, there are 100 jiffies to a second.  OpenTSDB calculates rate of change per second, and interpolates means between measurements to arrive at this figure, so we get 0-100 by virtue of all these things lining up.  In the case of any of these variables having a different value, you would not get this result.

@kylebrandt please correct me if I'm wrong.

lukaspustina:
@pdf Thanks for answering, but I'm afraid your wrong. Please take this with some salt, because I stopped reading kernel sources in 2011/2012:

1. The kernel "constant" `HZ` is architecture depended. 100 is the value often used for embedded platforms like ARM. But for more powerful platforms like x86 it is set by default to 1000; alpha uses 1024. I just checked the kernel sources for 3.14.48; also cf. "Understanding the Linux Kernel", 3rd edtion, page 230.
1. On one of our production systems running Ubuntu 14.04 and 3.13.0-27-generic has `CONFIG_HZ` set to 250. I have no clue why, but the calculation you're describing cannot be correct for this system. 
1. Even if you were right, it does not explain why it works for `linux.cpu.percpu` but not for `linux.cpu`.

Unfortunately, I have no better explanation.

lukaspustina:
Just an amendment: I just measured the timer tick frequency on the above mentioned Ubuntu system and it is `4096 Hz` which better fits my memory. I still don't know how the Kernel is configured in Ubuntu.

pdf:
@lukaspustina I believe jiffies are now calculated against USER_HZ, which for this very reason is defined as 100.  As for why `linux.cpu` and `linux.cpu.percpu` are different - each CPU has 100 jiffies per second to divide between tasks, so on a 4 CPU (or core) system, the total at `linux.cpu` will be 4 times the respective real time, whereas by taking the average over the 4 cores in `linux.cpu.percpu` you get reduce this down to a single CPU worth of time.

lukaspustina:
@pdf Aha! Perfect.

1. You're right -- kernel source for 3.14.48
   1. `x86/include/asm/compat.h:13:#define COMPAT_USER_HZ		100`
   1.  From `include/linux/jiffies.h`

        ```
        62:/* TICK_USEC is the time between ticks in usec assuming fake USER_HZ */
        63:#define TICK_USEC ((1000000UL + USER_HZ/2) / USER_HZ)
        ```

2. Your hypothesis seems to be correct. Dividing the values of `linux.cpu` by the number of CPUs yield values in the expected range of 1 ... 100.

pdf:
Got there in the end :wink:

lukaspustina:
Thanks for your explanations :)

PR #1144, opened 2015-07-07
Snmp tester

captncraig:
simple little web app to test snmp configs and oids and stuff outside of scollector. Not sure best way to deploy it. I can think of a few options:

- standalond app in cmd.
- hidden in bosun, since we already have web stuff there.
- Run from scollector as a special mode. 

This should be useful for anyone working on adding new snmp interfaces to scollector.


captncraig:
Ugly yet effective
screen shot 2015-07-07 at 9 05 55 am

screen shot 2015-07-07 at 9 06 47 am

mjibson:
I recommend a stand alone app in cmd.

PR #1148, opened 2015-07-08
collect: Add aggregation support to collect

kylebrandt:


collect/collect.go
```diff
@@ -179,6 +182,104 @@ func setHostName() error {
...
+	tchan <- &opentsdb.DataPoint{
+		Metric:    extRoot + "_99",
+		Timestamp: now,
+		Value:     percentile(am.values, .99),
+		Tags:      am.ts,
+	}
+}
+
+func percentile(dps []float64, p float64) (a float64) {
+	var x []float64
```
captncraig: pre-size this slice.

collect/collect.go
```diff
@@ -179,6 +182,104 @@ func setHostName() error {
...
+	for _, v := range dps {
+		x = append(x, v)
```
captncraig: x[i] = v

collect/collect.go
```diff
@@ -179,6 +182,104 @@ func setHostName() error {
...
+	}
+	sort.Float64s(x)
+	if p <= 0 {
+		return x[0]
```
captncraig: this could out of bounds if len == 0

collect/collect.go
```diff
@@ -179,6 +182,104 @@ func setHostName() error {
...
+	}
+	if p >= 1 {
+		return x[len(x)-1]
+	}
+	i := p * float64(len(x)-1)
+	i = math.Ceil(i)
```
captncraig: I feel like there is an overflow case here if p = .99 or something.

collect/collect.go
```diff
@@ -179,6 +182,104 @@ func setHostName() error {
...
+		Value:     percentile(am.values, .95),
+		Tags:      am.ts,
+	}
+	tchan <- &opentsdb.DataPoint{
+		Metric:    extRoot + "_99",
+		Timestamp: now,
+		Value:     percentile(am.values, .99),
+		Tags:      am.ts,
+	}
+}
```
captncraig: Do we want to clear the points after we process? Roll over n points or a fixed time window?

PR #1155, opened 2015-07-13
removing solaris stuff

captncraig:


mjibson:
Also need removal from the bosun.org/scollector page

Issue #987, opened 2015-05-26
Feature: monitor services automatically


dimamedvedev:
Current implementation can be improved as it's a bit difficult to debug regexes given that processes may have long argument list.

What I propose here is:
1. Detect init system (sysv, runit, upstart, systemd) and currently running services via specific init-commands (e.g. `sv status /etc/sv` for runit).
2. Find all processes belonging to service based on 
3. Assign an id to process based on process start time (i.e. service nginx will have id '1' assigned for master process and subsequent ids for workers).
4. Gather and store process start timestamp in addition to metrics in `process_linux` collector. That way user can understand whether process was restarted or not if scollector dies.

elementai:
There can be rather large amount of code, I'm thinking about coding a library for it first, or even contribute to something like https://github.com/shirou/gopsutil .

kylebrandt:
Regarding the first five items:
 Item 3: Isn't this what we already do?
 Item 4: Rather do the number of elapsed seconds since that is what we already do with the windows process collector. This is likely very easy

Items 1 and 2: How do you select which processes to monitor in this model? Why is this model better?

elementai:
Item 3: I don't think so, there're no `os.Stat` calls in `process_linux` and no sorting based on any timestamps
Item 4: Agreed

Items 1 and 2: First and foremost, it's (hopefully) will be automatic, in `processes_windows` collector there's already a hardcoded regex https://github.com/bosun-monitor/bosun/blob/master/cmd/scollector/collectors/processes_windows.go#L27

Second, there's an option to monitor services (up/down/etc) too.

Although it all can be rather spammy, we can write a simple whitelist/blacklist filter in scollector.

tdinucci:
Closing - see linked PR

PR #1161, opened 2015-07-14
cmd/scollector: collect process metrics from linux initd, systemd and runit

dimamedvedev:
This is for #987 . Posted for discussion of feature hence WIP
What it does:
* read pidfiles from `/etc/service/*/supervise/pid` & `/var/run/**/*.pid`
* gather those processes children via /proc `PPID` using small `pstree` lib
* create `WatchedProc` structs and use `process_linux.go` to collect metrics

It probably needs:
* blacklist/whitelist feature
* systemd support
* gather service-specific metrics (e.g. enabled/disabled)
* some love for `pstree.go`

Pros:
* easier to config than regexps

Cons:
* a bit spammy by default without blacklisting
* initd pid files names/locations in `/var/run` are not 100% uniform across distributions and daemons


kylebrandt:
I'm not sure this is really better to be honest. Additional con is that monitoring will brake with any issues involving stale pids files. Also I think the variation between distributions and daemons adds a long tail of work to the problem that I outweighs something that might be easier to config.

gbrayut:
If this is added it should still use a regex based whitelist. If you want to monitor everything you could use .*, and if you don't want to enable it you could omit that section from the TOML file. Adding a black list regex match would help cut down the spammy ness, but we've had issues with dead pid files so we probably wouldn't enable this feature.

captncraig:
I agree with @gbrayut. I think we have gotten to the point where new collectors need to be disabled unless specifically enabled by config unless they are deemed universal enough to be desired everywhere.

dimamedvedev:
I propose that only as an addition to existing feature. It shall work without problems with runit and systemd and probably other systems that enforce pid file consistency. And, yeah, with required white list regex it wont hurt even with sysv-init.

Also to me stale pids files are bad on their own because without them one can't even restart a daemon in most cases.

dimamedvedev:
Added whitelist support from config, so collector wont run automatically. Also systemd support via running `systemctl list-units` and `systemctl show Name -p MainPID`.

dimamedvedev:
Feels complete. 
Also gotta find out why tests fail.

IzakMarais:
For the systemd based collector, we would really appreciate if the collector could monitor services and report failed states. We want to have a metric that we can query explicitly to see of there are any failed services. This is as opposed to the current scollector process-linux functionality that just doesn't emit any metrics for non-running processes. 

By using `systemctl list-units -t services` one can see any services that are in a failed state (or stuck in an activating state) and report on those.

gbrayut:
This branch needs to be rebased on the current master branch. There is a merge conflict, but you can see an example of the resolved files in the last two commits on https://github.com/bosun-monitor/bosun/commits/services-linux

I had to move the collectors.*Services references out of main.go to prevent the following build error, but we had to do this before for platform specific process metrics (see latest commit on that branch for details):

```
C:\Code\Go\src\bosun.org [services-linux +0 ~1 -0]> go build bosun.org/...
# bosun.org/cmd/scollector
C:\code\Go\src\bosun.org\cmd\scollector\main.go:137: undefined: collectors.RunitServices
C:\code\Go\src\bosun.org\cmd\scollector\main.go:140: undefined: collectors.InitdServices
C:\code\Go\src\bosun.org\cmd\scollector\main.go:143: undefined: collectors.SystemdServices
```

Using the above branch and the following entry in the scollector.toml file:
```
[[SystemdServices]]
  WhiteList = "docker|puppet|influxdb|vmtoolsd"
  BlackList = ""
```
I was able to see metrics under linux.proc.cpu with the same name as specified in the WhiteList. Docker had multiple entries (with ID tag values starting at 0) so it appears to be working.

Like mentioned above it would be nice to add linux.systemd.running and linux.systemd.failed metrics that could be used to alert on monitored services that are in a bad state. Could try and combine them into something like linux.systemd.service.status using metadata.StatusCode where 0=active/running and you assign other codes to exited/waiting/failed/etc but this can get tricky. win.service metrics include .started and .status, so maybe linux.systemd.running and linux.systemd.failed would be good to start with.

dimamedvedev:
Very old PR, gonna rewrite and open new one.

PR #1167, opened 2015-07-15
bosun: Validating user and message on action. Requiring message to send notifications.

captncraig:
Simply highlights erroneous fields and doesn't submit.

screen shot 2015-07-15 at 10 06 37 am

I am still not sure if this is a good idea or not.


gbrayut:
Fine with me when the Send Notification box is checked. I often don't want to send notifications when I am just clearing the dashboard, so having it prompt would remind me to uncheck the box :-P

PR #1176, opened 2015-07-17
Add jsonq which allows JSON object handling in templates

kylebrandt:


cmd/bosun/sched/template.go
```diff
@@ -427,6 +427,27 @@ func (c *Context) HTTPGet(url string) string {
...
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode >= 300 {
+		return nil, fmt.Errorf("%v: returned %v", url, resp.Status)
+	}
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+	data := map[string]interface{}{}
```
captncraig: `make(map[string]interface{})`

PR #1181, opened 2015-07-21
cmd/scollector: WIP RabbitMQ

dimamedvedev:
Submitted only to discuss specific parts. Review not needed yet.

### DIFF
--- a/cmd/scollector/collectors/rabbitmq.go
+++ b/cmd/scollector/collectors/rabbitmq.go
@@ -1,0 +1,863 @@
+package collectors
+
+import (
+	"crypto/tls"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/url"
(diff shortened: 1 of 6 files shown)


cmd/scollector/collectors/rabbitmq.go
```diff
@@ -0,0 +1,691 @@
...
+		return nil, err
+	}
+	var md opentsdb.MultiDataPoint
+	for _, nv := range nodes {
+		name, ok := nv["name"].(string)
+		if !ok {
+			err := fmt.Errorf("rabbitmq: invalid node name '%s'", name)
+			return nil, err
+		}
+		if !strings.HasSuffix(name, util.Hostname) {
```
dimamedvedev: `/api/nodes` exposes metrics for all nodes in cluster, given that scollector runs on each I just skip everything that's not for current node.

PR #1185, opened 2015-07-22
collect: Add total_bytes and sample batchsize. Fixes #1184

### LINKED ISSUE #1184: collect: post.error/bad_status/restore need to be zeroed out so they don't go unknown
The add methods for "collect.post.error", "collect.post.bad_status", and "collect.post.restore" don't get called until an error occurs, which means they can go unknown when scollector or another system using collect is restarted. See http://goo.gl/60nInR

I'll be fixing this in a PR shortly

gbrayut:
Also zero out some post metrics on start to prevent them from going unknown.


gbrayut:
Running as a second instance on a few systems pointed at devbosun. See http://goo.gl/aMHXg3

PR #1187, opened 2015-07-22
bosun: making save rely on fewer goroutines needing to acquire the lock.

captncraig:
Uses a single, buffered channel to signal a save is needed. A single worker goroutine monitors this channel and saves as needed. 

In the case of multiple saves in fast succession, this should reduce lock contention on the schedule lock.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: go
 sudo: false
 go:
-  - tip
+  - 1.4.2
 
 notifications:
   email: false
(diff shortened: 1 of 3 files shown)


gbrayut:
code LGTM, but we need to rewrite the i package name now that go 1.5 doesn't let you import internal packages. See https://github.com/golang/go/commit/d70943431397b297fe00c114dd3e97376277f09a and travisci failure:

imports bosun.org/_third_party/golang.org/x/net/internal/iana: use of internal package not allowed

PR #1188, opened 2015-07-22
bosun: Reducing time spent locked during save.

captncraig:
Previously save didn't cause problems because it did so entirely unlocked. I am surprised we haven't corrupted our state file by now. Just gotta try to keep time in lock to minimum, so increasing throttle time to 10sec.

### DIFF
--- a/cmd/bosun/sched/bolt.go
+++ b/cmd/bosun/sched/bolt.go
@@ -83,6 +83,7 @@ func (s *Schedule) save() {
 		enc := gob.NewEncoder(cw)
 		if err := enc.Encode(data); err != nil {
 			log.Printf("error saving %s: %v", name, err)
+			s.Unlock()
 			return
 		}
 		if err := gz.Flush(); err != nil {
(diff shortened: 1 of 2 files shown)


cmd/bosun/sched/bolt.go
```diff
@@ -103,6 +95,7 @@ func (s *Schedule) save() {
 		log.Printf("wrote %s: %v", name, conf.ByteSize(cw.written))
 		collect.Put("statefile.size", opentsdb.TagSet{"object": name}, cw.written)
 	}
+	s.Unlock()
```
gbrayut: What about the return above... do we need to defer the call to unlock or make sure to unlock before that return?

captncraig: Don't want to defer becuase I don't want to include the bolt io time. Bolt does its own locking on transactions for that part. I'll add the unlock explicitly.

gbrayut: Sounds good. fix that and LGTM

PR #1190, opened 2015-07-23
cmd/scollector: enable pprof = "ip:port" settings in TOML file

gbrayut:


cmd/scollector/main.go
```diff
@@ -71,6 +73,15 @@ func main() {
 	} else if conf.Tags["host"] != "" {
 		slog.Fatalf("host not supported in custom tags, use Hostname instead")
 	}
+	if conf.PProf != "" {
+		go func() {
+			slog.Infof("Starting pprof at http://%s/debug/pprof/", conf.PProf)
+			err := http.ListenAndServe(conf.PProf, nil)
```
captncraig: `slog.Fatal(http.ListenAndServe())` is how I most often see this. If it ever returns something bad happened.

gbrayut:
Update with slog.Fatal(http.ListenAndServe())

PR #1191, opened 2015-07-23
Expose port 4242 in Quickstart

SlashmanX:
The docker registry docs say to expose port 4242 as well as 8070, and the opentsdb exits without this port exposed.

### DIFF
--- a/quickstart.md
+++ b/quickstart.md
@@ -38,7 +38,7 @@ If you do not already have docker installed on your system, you can install it f
 
 Once Docker is installed, you will want to pull the latest version of Bosun from the Docker Registry. To do this, execute the following command:
 
-	$ docker run -d -p 8070:8070 stackexchange/bosun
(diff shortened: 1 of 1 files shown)


captncraig:
I don't think opentsdb exits if that is not exposed, you just have no way of talking to it from the outside. Bosun should still be able to talk to it inside the container. Are you sure it actually is exiting for you without the port forwarded?

SlashmanX:
Yep. 100% sure, it was exiting with `exit status 1; not expected` and the OpenTSDB logs were saying `TableNotFoundException: tsdb-uid` simply restart with port 4242 exposed fixed everything.

Beside, the readme on the docker registry says to expose 4242: https://registry.hub.docker.com/u/stackexchange/bosun/

So that's inconsistent at least

captncraig:
sounds good. Thanks!

PR #1192, opened 2015-07-23
making save run on a fixed clock

captncraig:
Save is now taking too long every 10 seconds in our install, because of the time needed to gob encode everything. 

This reduces the save frequency to once a minute, and removes all of the tricky triggers and just does it in a loop.


captncraig:
I don't think opentsdb exits if that is not exposed, you just have no way of talking to it from the outside. Bosun should still be able to talk to it inside the container. Are you sure it actually is exiting for you without the port forwarded?

PR #1194, opened 2015-07-23
cmd/bosun: Making search centralize indexing.

captncraig:
Instead of locking, use a channel, and a central indexing goroutine. If data channel overflows we lose some data points, but very few have new information anyway.

### DIFF
--- a/cmd/bosun/search/search.go
+++ b/cmd/bosun/search/search.go
@@ -5,9 +5,10 @@ import (
 	"regexp"
 	"sort"
 	"strings"
-	"sync"
 	"time"
 
+	"bosun.org/collect"
+	"bosun.org/metadata"
 	"bosun.org/opentsdb"
 )
 
@@ -31,10 +32,7 @@ type Search struct {
 	// value.
 	Read *Search
 
-	sync.RWMutex
-
-	// copy is true when there is a Copy() event pending.
-	copy bool
+	dps chan opentsdb.MultiDataPoint
 }
 
 type pair struct {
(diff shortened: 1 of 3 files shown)


cmd/bosun/search/search.go
```diff
@@ -108,58 +106,65 @@ func (s *Search) Copy() {
...
-			s.Tagk[dp.Metric][k] = now
+func (s *Search) indexLoop() {
+	ticker := time.NewTicker(20 * time.Second)
+	for {
+		select {
+		case <-ticker.C:
+			s.Copy()
+		case mdp := <-s.dps:
+			now := time.Now().Unix()
+
```
kylebrandt: Grrrr.... empty lines

kylebrandt:
I like the tests. I'm not sure we should be optimizing with a buffered chan unless this is solving an active problem. Does the CPU cycles outweigh the risk of unanticipated consequences or more difficult debugging? http://www.slideshare.net/cloudflare/a-channel-compendium

We definitely want to keep the tests either way. If you do go with the buffered channel, can we instrument its length and overflows with collect somehow?

captncraig:
We can instrument drops for sure. The whole reason for this is that pprof shows consistently hundreds of goroutines waiting on the search lock. Generally shouldn't have too high of a cost, but that many sockets held open any longer than necessary for search indexing seems unnecessary. Yes, we should definitely instrument overflows, but the idea is that occasional overflows are ok, since we should see metric/tag sets many many times.

captncraig:
redis migration should make this entirely different holding off for now.

PR #1196, opened 2015-07-23
bosun: capturing timings around template rendering.

captncraig:


cmd/bosun/sched/check.go
```diff
@@ -132,40 +133,8 @@ func (s *Schedule) RunHistory(r *RunHistory) {
...
-					emailbody = body
-					attachments = nil
-				}
-				state.Subject = string(subject)
-				state.Body = string(body)
-				state.EmailBody = emailbody
-				state.EmailSubject = emailsubject
-				state.Attachments = attachments
-			}
+
```
kylebrandt: No blank lines within funcs

cmd/bosun/sched/check.go
```diff
@@ -250,6 +219,61 @@ func (s *Schedule) RunHistory(r *RunHistory) {
...
 
+func (s *Schedule) executeTemplates(state *State, event *Event, a *conf.Alert, r *RunHistory) {
+	state.Subject = ""
+	state.Body = ""
+	state.EmailBody = nil
+	state.EmailSubject = nil
+	state.Attachments = nil
+	if event.Status != StUnknown {
+		metric := "template.render"
+
```
kylebrandt: Blank lines again... won't comment on the rest of them

collect/collect.go
```diff
@@ -289,6 +289,21 @@ func Sample(metric string, ts opentsdb.TagSet, v float64) error {
...
 }
 
+// StartTimer records the current time, and returns a function you can call to
+// record the end of your action.
+//
+// Typical usage would be:
+//    done := collect.StartTimer("myMetric", opentsdb.TagSet{})
+//    doMyThing()
+//    done()
+func StartTimer(metric string, ts opentsdb.TagSet) func() {
```
kylebrandt: Consider testing.Benchmark (http://stackoverflow.com/questions/8350609/how-do-you-time-a-function-in-go-and-return-its-runtime-in-milliseconds). Haven't looked into this in depth, maybe using testing here is just wrong. Up to you

kylebrandt:
Also remember commits should be cmd/bosun prefixed

collect/collect.go
captncraig: Benchmark tests are great for build-time benchmarks. It can't be used for run-time instrumentation like this is doing.

PR #1197, opened 2015-07-23
bosun: Requiring no schedule lock to render dashboard.

captncraig:
Two main fixes:

1. Reading silences does not require schedule lock, but a more localized silence lock, which is a RWMutex that should take very few Writes.
2. Schedule now has a `readStatus` that is a copy of the authoritative `status`. This should be copied in full at the end of each `RunHistory`.


kylebrandt:
Need to pass tests of course...

PR #1202, opened 2015-07-25
cmd/scollector: RabbitMQ

dimamedvedev:
Fairly comprehensive collector for RabbitMQ 3.x Probably needs careful review of metric/tags design.


cmd/scollector/doc.go
```diff
@@ -238,6 +238,11 @@ or just one.
 	[[HTTPUnit]]
 	  TOML = "/some/other.toml"
 
+RabbitMQ (array of table, keys are URL): RabbitMQ hosts to poll.
```
captncraig: perhaps document that it will check localhost, even with no config given.

cmd/scollector/collectors/rabbitmq.go
```diff
@@ -0,0 +1,898 @@
...
+}
+func c_rabbitmq_nodes() (opentsdb.MultiDataPoint, error) {
+	return rabbitmqNodes(defaultRabbitmqURL)
+}
+func c_rabbitmq_queues() (opentsdb.MultiDataPoint, error) {
+	return rabbitmqQueues(defaultRabbitmqURL)
+}
+
+func rabbitmqOverview(url string) (opentsdb.MultiDataPoint, error) {
+	tr := &http.Transport{
```
captncraig: I really hate doing this. Why is it necessary?

cmd/scollector/collectors/rabbitmq.go
```diff
@@ -0,0 +1,898 @@
...
+	res, err := client.Get(url + "/api/overview")
+	if err != nil {
+		return nil, err
+	}
+	defer res.Body.Close()
+	content, err := ioutil.ReadAll(res.Body)
+	if err != nil {
+		return nil, err
+	}
+	var overview map[string]interface{}
```
captncraig: I would much rather see a real struct type for this than map[string]interface{}. Can you paste me some of the actual json blobs in a gist or something?

cmd/scollector/collectors/rabbitmq.go
dimamedvedev: Oh, that should be removed.

cmd/scollector/collectors/rabbitmq.go
dimamedvedev: https://gist.github.com/dimamedvedev/a2a62a797bbafd229055

dimamedvedev:
One of the problems I've encountered is `"consumer_utilisation"` field. When no one consumes from queue it's: `"consumer_utilisation": "",` , when there's one it's `"consumer_utilisation": 0.9842390041698995,` .

captncraig:
That would be an acceptable use of an interface{} field. Most others look like they could be mapped into structs. Have you seen this tool: http://mholt.github.io/json-to-go/?

dimamedvedev:
Yeah, seen that, will do some research on how does RMQ output stats in different versions.

kylebrandt:
@dimamedvedev Also the gojson as a local utility is quite nice https://github.com/ChimeraCoder/gojson , you just do `curl ... | gojson` once you have it installed

dimamedvedev:
* Updated PR to use structs
 * Left off a couple of metrics - can't find a way enable them in RabbitMQ
 * Some metrics could be converted to tags

cmd/scollector/collectors/rabbitmq.go
```diff
@@ -0,0 +1,436 @@
...
+		ts := opentsdb.TagSet{"host": host, "queue": q.Name, "vhost": q.Vhost}
+		Add(&md, p+"consumers", q.Consumers, ts, metadata.Gauge, metadata.Consumer, DescRmqConsumers)
+		Add(&md, p+"memory", q.Memory, ts, metadata.Gauge, metadata.Bytes, DescRmqMemory)
+		Add(&md, p+"message_bytes_total", q.MessageBytes, ts, metadata.Gauge, metadata.Bytes, DescRmqMessageBytes)
+		Add(&md, p+"message_bytes_persistent", q.MessageBytesPersistent, ts, metadata.Gauge, metadata.Bytes, DescRmqMessageBytesPersistent)
+		Add(&md, p+"message_bytes_transient", q.MessageBytesRAM, ts, metadata.Gauge, metadata.Bytes, DescRmqMessageBytesRAM)
+		Add(&md, p+"message_bytes_ready", q.MessageBytesReady, ts, metadata.Gauge, metadata.Bytes, DescRmqMessageBytesReady)
+		Add(&md, p+"message_bytes_unack", q.MessageBytesUnacknowledged, ts, metadata.Gauge, metadata.Bytes, DescRmqMessageBytesUnacknowledged)
+		Add(&md, p+"messages_queue_depth", q.Messages, ts, metadata.Gauge, metadata.Message, DescRmqMessages)
+		Add(&md, p+"messages_persistent", q.MessagesPersistent, ts, metadata.Gauge, metadata.Message, DescRmqMessagesPersistent)
```
dimamedvedev: These could be tagged, not 100% sure.

cmd/scollector/collectors/rabbitmq.go
```diff
@@ -0,0 +1,440 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer res.Body.Close()
+	content, err := ioutil.ReadAll(res.Body)
+	if err != nil {
+		return nil, err
+	}
+	var o rmqOverview
+	if err := json.NewDecoder(bytes.NewReader(content[:])).Decode(&o); err != nil {
```
captncraig: You don't even need to read the body prior to this. Just use `json.NewDecoder(res.Body)`

cmd/scollector/collectors/rabbitmq.go
```diff
@@ -0,0 +1,440 @@
...
+		return nil, err
+	}
+	var md opentsdb.MultiDataPoint
+	host := strings.Split(o.Node, "@")[1]
```
captncraig: Please check the result of split here. If something goes wrong and there is no @, this will panic.

cmd/scollector/collectors/rabbitmq.go
```diff
@@ -0,0 +1,440 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer res.Body.Close()
+	content, err := ioutil.ReadAll(res.Body)
+	if err != nil {
+		return nil, err
+	}
+	var qs []rmqQueue
+	if err := json.NewDecoder(bytes.NewReader(content[:])).Decode(&qs); err != nil {
```
captncraig: Same thing here. No need to ReadAll before decoding.

cmd/scollector/collectors/rabbitmq.go
```diff
@@ -0,0 +1,440 @@
...
+		return nil, err
+	}
+	var md opentsdb.MultiDataPoint
+	for _, q := range qs {
+		host := strings.Split(q.Node, "@")[1]
```
captncraig: No unsafe slice indexing.

cmd/scollector/collectors/rabbitmq.go
```diff
@@ -0,0 +1,440 @@
...
+		Add(&md, p+"message_bytes_ready", q.MessageBytesReady, ts, metadata.Gauge, metadata.Bytes, DescRmqMessageBytesReady)
+		Add(&md, p+"message_bytes_unack", q.MessageBytesUnacknowledged, ts, metadata.Gauge, metadata.Bytes, DescRmqMessageBytesUnacknowledged)
+		Add(&md, p+"messages_queue_depth", q.Messages, ts, metadata.Gauge, metadata.Message, DescRmqMessages)
+		Add(&md, p+"messages_persistent", q.MessagesPersistent, ts, metadata.Gauge, metadata.Message, DescRmqMessagesPersistent)
+		Add(&md, p+"messages_transient", q.MessagesRAM, ts, metadata.Gauge, metadata.Message, DescRmqMessagesRAM)
+		Add(&md, p+"messages_ready_total", q.MessagesReady, ts, metadata.Gauge, metadata.Message, DescRmqMessagesReady)
+		Add(&md, p+"messages_ready_transient", q.MessagesReadyRAM, ts, metadata.Gauge, metadata.Message, DescRmqMessagesReadyRAM)
+		Add(&md, p+"messages_unack_total", q.MessagesUnacknowledged, ts, metadata.Gauge, metadata.Message, DescRmqMessagesUnacknowledged)
+		Add(&md, p+"messages_unack_transient", q.MessagesUnacknowledgedRAM, ts, metadata.Gauge, metadata.Message, DescRmqMessagesUnacknowledgedRAM)
+		if sn, ok := q.SlaveNodes.([]interface{}); ok {
```
captncraig: Is this working for you? I would be extremely skeptical that it is.

dimamedvedev: I need some clarification.

Code works
```
2015/08/13 13:41:44 info: queue.go:66: {"metric":"rabbitmq.queue.slave_nodes","timestamp":1439462503,"value":2,"tags":{"host":"rabbit1","queue":"hello","vhost":"/"}}
2015/08/13 13:41:44 info: queue.go:66: {"metric":"rabbitmq.queue.sync_slave_nodes","timestamp":1439462503,"value":2,"tags":{"host":"rabbit1","queue":"hello","vhost":"/"}}
```
or do you mean like do we need those metrics at all? TBH I don't know that _yet_ because we are only planning to use mirror queues, but I'm fairly confident that we could use these.

dimamedvedev:
@captncraig updated PR, will squash when ready

cmd/scollector/collectors/rabbitmq.go
captncraig: I'm actually surprised that `q.SlaveNodes` successfully casts to `[]interface{}` at runtime. But if it works it is ok by me.

cmd/scollector/collectors/rabbitmq.go
```diff
@@ -0,0 +1,437 @@
...
+
+func rabbitmqBackingQueueStatus(p string, ts opentsdb.TagSet, bqs rmqBackingQueueStatus) opentsdb.MultiDataPoint {
+	var md opentsdb.MultiDataPoint
+	Add(&md, p+"avg_rate", bqs.AvgAckEgressRate, ts.Copy().Merge(opentsdb.TagSet{"method": "ack", "direction": "out"}),
+		metadata.Rate, metadata.Message, DescRmqBackingQueueStatusAvgAckEgressRate)
+	Add(&md, p+"avg_rate", bqs.AvgAckEgressRate, ts.Copy().Merge(opentsdb.TagSet{"method": "ack", "direction": "in"}),
+		metadata.Rate, metadata.Message, DescRmqBackingQueueStatusAvgAckIngressRate)
+	Add(&md, p+"avg_rate", bqs.AvgAckEgressRate, ts.Copy().Merge(opentsdb.TagSet{"direction": "out"}),
+		metadata.Rate, metadata.Message, DescRmqBackingQueueStatusAvgEgressRate)
+	Add(&md, p+"avg_rate", bqs.AvgAckEgressRate, ts.Copy().Merge(opentsdb.TagSet{"direction": "in"}),
```
captncraig: You are adding the same two data points twice, with different tagsets (once with direction and method, once with only direction). This scheme seems like it may cause confusing data conditions. Why Not have just the first two Adds and the last one?

dimamedvedev: Oh, my bad, easy to get lost in so many metrics. Used wrong datapoints. Thank you for being thorough.

PR #1204, opened 2015-07-25
cmd/scollector: riak allow custom URL

dimamedvedev:
Allow to specify custom URL for Riak collector.

### DIFF
--- a/cmd/scollector/collectors/riak.go
+++ b/cmd/scollector/collectors/riak.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"net/url"
 	"strings"
 
 	"bosun.org/metadata"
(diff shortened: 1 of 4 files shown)


cmd/scollector/collectors/riak.go
```diff
@@ -463,13 +464,32 @@ const (
 	riakURL string = "http://localhost:8098/stats"
```
captncraig: maybe call this "localRiakURL" or something.

cmd/scollector/collectors/riak.go
```diff
@@ -463,13 +464,32 @@ const (
...
+
 func enableRiak() bool {
 	return enableURL(riakURL)()
 }
 
 func c_riak() (opentsdb.MultiDataPoint, error) {
+	return riak(riakURL)
+}
+
+func riak(_url string) (opentsdb.MultiDataPoint, error) {
```
captncraig: I would prefer `u` to `_url`

PR #1205, opened 2015-07-26
cmd/bosun: Export time.ParseDuration as ParseDuration to templates

### LINKED ISSUE #1014: Expose 'func (Time) Add' function in template engine
Hi,
In our use-case, we're generating Kibana URLs that include timestamps for dashboard time-window display.
Currently, I can put either fixed ```<time>-ago``` notation (4h-ago) or absolute time (from t1 to t2).

A relevant time-frame can be injected to the generated URL by taking the alert timestamp and subtracting  a duration using ```time.Add(<negative_duration>)```. the resulting Dashboard URL will always display a time-window like 1 hour before the alert until 'now'.

Thanks,
Yarden

kylebrandt:
Fixes #1014

### DIFF
--- a/cmd/bosun/conf/conf.go
+++ b/cmd/bosun/conf/conf.go
@@ -675,6 +675,7 @@ var defaultFuncs = ttemplate.FuncMap{
 	"short": func(v string) string {
 		return strings.SplitN(v, ".", 2)[0]
 	},
+	"parseDuration": time.ParseDuration,
 }
 
 func (c *Conf) loadTemplate(s *parse.SectionNode) {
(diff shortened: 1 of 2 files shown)


cmd/bosun/conf/conf.go
```diff
@@ -675,6 +675,7 @@ var defaultFuncs = ttemplate.FuncMap{
 	"short": func(v string) string {
 		return strings.SplitN(v, ".", 2)[0]
 	},
+	"ParseDuration": time.ParseDuration,
```
captncraig: lower case looks to be the convention. `parseDuration`

captncraig:
@kylebrandt what problem does this solve. It looks like @ayashjorden solved #1014 without it.

kylebrandt:
@captncraig See the associated issue. It makes it so we don't have to do some string replacement or ugly constants in the templates.

captncraig:
ok, change to parseDuration and merge.

PR #1206, opened 2015-07-26
cmd/bosun: Refactor filter expr func to also filter out zero valued results

### LINKED ISSUE #1100: Redesign filter func()
I think I screwed up. Instead of filter working purely based on the groups existing or not existing, I think it might be better if it returns groups that exist and the value for the group (in the second argument) is not zero. That way we could do things like:

```
filter(q("sum:rate:os.net.pause_frames{direction=out,host=ny-nexus*,iname=*}", "10h", "8h"), avg(q("sum:rate:os.net.pause_frames{direction=out,host=ny-nexus*,iname=*}", "10h", "8h")) != 0)
```

instead of:

```
filter(q("sum:rate:os.net.pause_frames{direction=out,host=ny-nexus*,iname=*}", "10h", "8h"), limit(sort(avg(q("sum:rate:os.net.pause_frames{direction=out,host=ny-nexus*,iname=*}", "10h", "8h")) != 0, "desc"), 20))
```

Thoughts anyone?

kylebrandt:
…esults (false)

This is a breaking change. Fixes #1100

### DIFF
--- a/cmd/bosun/expr/funcs.go
+++ b/cmd/bosun/expr/funcs.go
@@ -360,7 +360,9 @@ func Filter(e *State, T miniprofiler.Timer, series *Results, number *Results) (*
 	for _, sr := range series.Results {
 		for _, nr := range number.Results {
 			if sr.Group.Subset(nr.Group) || nr.Group.Subset(sr.Group) {
-				ns = append(ns, sr)
+				if nr.Value.Value().(Number) != 0 {
+					ns = append(ns, sr)
+				}
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


cmd/bosun/expr/funcs.go
```diff
@@ -360,7 +360,9 @@ func Filter(e *State, T miniprofiler.Timer, series *Results, number *Results) (*
 	for _, sr := range series.Results {
 		for _, nr := range number.Results {
 			if sr.Group.Subset(nr.Group) || nr.Group.Subset(sr.Group) {
-				ns = append(ns, sr)
+				if nr.Value.Value().(Number) != 0 {
```
captncraig: This type of type assertion will panic if anything but a number makes it here.

kylebrandt: @captncraig I'm aware but throughout the funcs.go we don't check them since the map[string]parse.Func{... ensures we won't get there.

PR #1223, opened 2015-07-30
cmd/scollector: use dmidecode to get serialNumber and model metadata …

gbrayut:
…since the Dell method stopped working


gbrayut:
Working on my test vm pointed at devbosun: http://devbosun.ds.stackexchange.com/host?host=ny-gbraylx01&time=1d-ago&tab=metadata

PR #1231, opened 2015-08-03
bosun: Concurrent checks

captncraig:
This modifies the schedule in many ways. Each alert runs in its own goroutine.


kylebrandt:
Will fix #1228

cmd/bosun/web/web.go
```diff
@@ -570,17 +570,18 @@ func APIRedirect(w http.ResponseWriter, req *http.Request) {
...
-	case checkRunning <- true:
-		// Good, we've got the lock.
-	default:
-		return 0, fmt.Errorf("check already running")
-	}
-	d, err := schedule.Check(t, time.Now(), 0)
-	<-checkRunning
-	return d, err
-}
+//TODO: This makes less sense with every alert running independently.
```
kylebrandt: This can just be yanked as a feature. The reason I created it was so when someone "fixed" an alert, they didn't have to wait for it to return to normal. But because queries cover a time range, this is generally pretty futile. My idea to actually fix this problem is delayed close as described in https://github.com/bosun-monitor/bosun/issues/808

captncraig:
@gbrayut, I would love a thorough review on this if you don't mind.

cmd/bosun/sched/notify.go
```diff
@@ -14,23 +14,27 @@ import (
...
 		}
 	}
+
 }
 
 func (s *Schedule) Notify(st *State, n *conf.Notification) {
-	if s.notifications == nil {
-		s.notifications = make(map[*conf.Notification][]*State)
+	if s.pendingNotifications == nil {
+		s.pendingNotifications = make(map[*conf.Notification][]*State)
```
gbrayut: I'm not well versed in making maps or appending to them, but this smells like something that might need a lock or some other means of preventing duplicate assignments to the variable

captncraig: It does require locking. Anywhere this function is called the lock should be held. We are really bad at specifying those kinds of conventions though. This should almost certainly not be exported if that is the case, and should probably note that.

PR #1242, opened 2015-08-05
cmd/scollector: Identify the interface type via IF-MIB:ifType

kylebrandt:
Instead of string parsing - mostly.

### DIFF
--- a/cmd/scollector/collectors/snmp_ifaces.go
+++ b/cmd/scollector/collectors/snmp_ifaces.go
@@ -28,6 +28,7 @@ const (
 	ifOutDiscards        = ".1.3.6.1.2.1.2.2.1.19"
 	ifOutErrors          = ".1.3.6.1.2.1.2.2.1.20"
 	ifOutPauseFrames     = ".1.3.6.1.2.1.10.7.10.1.4"
+	ifType               = "1.3.6.1.2.1.2.2.1.3"
 )
 
 // SNMPIfaces registers a SNMP Interfaces collector for the given community and host.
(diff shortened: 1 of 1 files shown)


cmd/scollector/collectors/snmp_ifaces.go
```diff
@@ -41,15 +42,27 @@ func SNMPIfaces(cfg conf.SNMP) {
...
+	switch ifType {
+	case 6:
+		return metric
+	case 53:
+		return osNet + ".bond" + strings.TrimPrefix(metric, osNet)
+	case 135:
+		return osNet + ".virtual" + strings.TrimPrefix(metric, osNet)
+	case 131:
+		return osNet + ".tunnel" + strings.TrimPrefix(metric, osNet)
+	case 161:
```
captncraig: You can combine this with the `53` case by using `fallthrough`.

cmd/scollector/collectors/snmp_ifaces.go
```diff
@@ -64,11 +77,23 @@ func c_snmp_ifaces(community, host string) (opentsdb.MultiDataPoint, error) {
 	if err != nil {
 		return nil, err
 	}
+	t, err := snmp_subtree(host, community, ifType)
```
captncraig: can we fix up some of the one letter variable names in here? k and v I don't mind but a,n,t should all be more descriptive. `snmpNames`, `snmpTypes` perhaps? `ifTypes`?

kylebrandt: Sure, the tricky thing is that they are basically the "preprocessed"  versions of those, any suggestions?

captncraig: perhaps `names` and `nameMap`

PR #1246, opened 2015-08-12
Change Elasticsearch backend configuration key

azhurbilo:
logstashElasticHost is not used in code base, only 'logstashElasticHosts'


kylebrandt:
Thanks, this changed when the elastic library we used was updated. The new elastic library takes a set of hosts, and learns about the rest of the cluster from that set (i.e. hosts that have been removed or added).

PR #1247, opened 2015-08-12
cmd/bosun: Add an API route to return all metrics and their associate…

### LINKED ISSUE #1245: Bosun: /api/tagk to get all current tags across all metrics (also support basic filtering)
We were trying to find if a current tag is used, but the API doesn't have a route to return all tags. We could use the /api/metric route to get all metrics and then call /api/tagk/{metric} but it would be better if there was a /api/tagk route that returned a list of all metrics and tags.

/api/metadata/get has the information, but also includes other metadata which makes it an expensive call. The /api/tagk route could use the same format but just exclude the extra data:

[{"Metric":"os.disk.fs.space_used","Tags":{"disk":"/opt/gitlab"}},{"Metric":"web.comment.edited","Tags":{"host":"ny-web04","role":"coreqa","site":"datascience.stackexchange.com"}},{"Metric":"elastic.indices.suggest.current","Tags":{"cluster":"StackExchangeNetwork","index_name":"careers-full-cv-candidate-search-635643440065754201"}}]

The /api/metadata/get route is also close to what we would need when you provide a query string, but ?tagk=tagname doesn't work by itself (error: "unpaired tagk/tagv") or allow globs. The ability to request /api/tagk?tagk=tagname to find all metrics using a specific tag would be very helpful.

kylebrandt:
…d keys

Fixes #1245

### DIFF
--- a/cmd/bosun/search/search.go
+++ b/cmd/bosun/search/search.go
@@ -284,6 +284,21 @@ func (s *Search) TagKeysByMetric(Metric string) []string {
 	return r
 }
 
+func (s *Search) MetricsWithTagKeys() map[string][]string {
+	metricToKeys := make(map[string][]string)
+	searchResult := s.Read.Tagk.Copy()
+	for metric, v := range searchResult {
+		keys := make([]string, len(v))
+		i := 0
+		for tK := range v {
+			keys[i] = tK
+			i++
+		}
+		metricToKeys[metric] = keys
+	}
+	return metricToKeys
+}
+
 func (s *Search) tagValuesByMetricTagKey(Metric, Tagk string, since time.Duration) []string {
 	var t int64
 	if since > 0 {
--- a/cmd/bosun/web/web.go
+++ b/cmd/bosun/web/web.go
@@ -99,6 +99,7 @@ func Listen(listenAddr string, devMode bool, tsdbHost string) error {
 	router.Handle("/api/metadata/put", JSON(PutMetadata))
 	router.Handle("/api/metric", JSON(UniqueMetrics))
 	router.Handle("/api/metric/{tagk}/{tagv}", JSON(MetricsByTagPair))
+	router.Handle("/api/metric/tagkey", JSON(MetricsWithTagKeys))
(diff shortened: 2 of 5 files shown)


cmd/bosun/search/search.go
```diff
@@ -284,6 +284,20 @@ func (s *Search) TagKeysByMetric(Metric string) []string {
 	return r
 }
 
+func (s *Search) TagKeysByMetricTree() map[string][]string {
```
captncraig: This name does not describe to me what it actually does. Which is the top level key? I am assuming the hierarchy is `metric -> all tagks`. Perhaps `MetricsWithTagKeys` would make a bit more sense.

cmd/bosun/search/search.go
```diff
@@ -284,6 +284,20 @@ func (s *Search) TagKeysByMetric(Metric string) []string {
...
+	metricToKeys := make(map[string][]string)
+	searchResult := s.Read.Tagk.Copy()
+	for metric, v := range searchResult {
+		metricToKeys[metric] = make([]string, len(v))
+		i := 0
+		for tK := range v {
+			metricToKeys[metric][i] = tK
```
captncraig: For readability I recommend making and filling the slice, and then putting it into `metricsToKeys` at the end of the loop. That may reduce the chained map/slice indexing.

cmd/bosun/web/web.go
```diff
@@ -99,6 +99,7 @@ func Listen(listenAddr string, devMode bool, tsdbHost string) error {
 	router.Handle("/api/metadata/put", JSON(PutMetadata))
 	router.Handle("/api/metric", JSON(UniqueMetrics))
 	router.Handle("/api/metric/{tagk}/{tagv}", JSON(MetricsByTagPair))
+	router.Handle("/api/metric/keytree", JSON(TagKeysByMetricTree))
```
captncraig: Naming here is harder yet. `keytree` is not super descriptive to me. You can call it what you want though.

PR #1255, opened 2015-08-17
WIP: Add timeouts to Eval-like Template functions

yyljlyy:


captncraig:
@yyljlyy thanks, but @kylebrandt is still working on this. It needs a few more revisions, and we are waiting on #1231 as well.

PR #1257, opened 2015-08-18
cmd/bosun: Add normalize switch to graph page to display series on 0-…

kylebrandt:
…100 scale

Each series is displayed on a scale of 0-100. This helps to visually see if there is apparent correlation between metrics with vastly different scales. I tried to other options, d3.power scales and d3.log scales. Power scales weere not effective for getting good correlation views. Log scales don't implement well for ad-hoc graphing. There can't be 0 values, series cannot cross the positive/negative threshold, and negative series must be handled differently.


kylebrandt:
go generate error is a lie.

kylebrandt:
Here is what it looks like:

image

Vs 

image

captncraig:
LGTM

Dieterbe:
@kylebrandt just curious, what correlates on your first screenie? green and blue?

kylebrandt:
Oh nothing really, those were just metrics with varied scales

PR #1259, opened 2015-08-18
Update logstashElasticHosts format

briantist:
The original says it uses the same format as `tsdbHost`, but it doesn't work if you just use `host:port`. Looking at the code, bosun passes the value(s) here directly to the elastic library which is expecting a URL. Once I used `http://host:port` I was able to query. The error message given before came directly from the elastic library with a generic error message about there being no hosts available so there was no indication that the configuration format was incorrect.


kylebrandt:
I think is already done? someone just needs to regenerate the config? @captncraig can your write how that is done in the developing md?

captncraig:
What config? Do you mean publishing the docs?

kylebrandt:
@captncraig Yes, sorry publishing the docs... so much cold medicine...

PR #1260, opened 2015-08-19
cmd/scollector: Add MTU, AdminStatus, and OperationStatus to SNMP int…

kylebrandt:
…erface polling


kylebrandt:
Can review, but please don't merge as it needs to be tested.

PR #1266, opened 2015-08-19
Adding header to identify bosun host on all outgoing http requests.

### LINKED ISSUE #1264: Include bosun server/ip as a header when sending emails
When sending email it should indicate what server/ip it was generated from. Something like adding X-Bosun-Host: bosun.ds.stackexchange.com (10.7.0.220) as a header when using SMTP would help us identify rogue servers.

captncraig:
Fixes #1264


captncraig:
@gbrayut

gbrayut:
This adds a header to http requests, but the linked issue is for SMTP notifications not http. We should be able to add a header to the email that is visible in gmail when you view the original source

gbrayut:
Has a merge conflict but otherwise LGTM

PR #1267, opened 2015-08-19
collect: instrumenting some basic gc stats

captncraig:
I mostly want to see how upgrading to go 1.5 will affect these statistics, so instrumenting them now.


gbrayut:
LGTM
We can push to prod or branch bosun tomorrow

PR #1271, opened 2015-08-24
smd/scollector: Allowing multipart snmp keys in subtrees.

captncraig:
Snmp is horrible. Some trees index keys

```
.8.1
.8.2
.8.3
...
```

Others index them:
```
.8.1.1
.8.1.2
.8.1.3
...
```

Scollector would previously reject trees of the second form if present in a MIB. This patch makes all subtree paths strings and if a multipart key is  found simply converts it to `.1.3` style string.


gbrayut:
LGTM

PR #1280, opened 2015-08-26
cmd/scollector: Monitoring teams interfaces

kylebrandt:
Currently we monitor bonding which works with CentOS 6, but by default CentOS 7 is using teaming and we want to monitor that as well.

### DIFF
--- a/cmd/scollector/collectors/ifstat_linux.go
+++ b/cmd/scollector/collectors/ifstat_linux.go
@@ -1,8 +1,12 @@
 package collectors
 
 import (
+	"encoding/json"
+	"io/ioutil"
+	"path/filepath"
 	"regexp"
 	"strings"
+	"time"
 
 	"bosun.org/metadata"
 	"bosun.org/opentsdb"
@@ -12,6 +16,8 @@ import (
 func init() {
 	collectors = append(collectors, &IntervalCollector{F: c_ifstat_linux})
 	collectors = append(collectors, &IntervalCollector{F: c_ipcount_linux})
+	collectors = append(collectors, &IntervalCollector{F: c_if_team_linux})
+	collectors = append(collectors, &IntervalCollector{F: c_if_bond_linux})
 }
 
 var netFields = []struct {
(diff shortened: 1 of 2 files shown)


cmd/scollector/collectors/ifstat_linux.go
```diff
@@ -106,3 +112,162 @@ func c_ifstat_linux() (opentsdb.MultiDataPoint, error) {
...
+
+const (
+	linuxNetBondSlaveIsUpDesc = "The status of the bonded or teamed interface."
+	linuxNetBondSlaveCount    = "The number of slaves on the bonded or teamed interface."
+)
+
+func c_if_bond_linux() (opentsdb.MultiDataPoint, error) {
+	var md opentsdb.MultiDataPoint
+	const bondingPath = "/proc/net/bonding"
+	bondDevices, _ := ioutil.ReadDir(bondingPath)
```
captncraig: unchecked errors are bad. Especially in a case like this where the path might not exist.

cmd/scollector/collectors/ifstat_linux.go
```diff
@@ -106,3 +112,162 @@ func c_ifstat_linux() (opentsdb.MultiDataPoint, error) {
...
+	for _, fi := range bondDevices {
+		var iface string
+		var slaveCount int
+		if err := readLine(bondingPath+"/"+fi.Name(), func(s string) error {
```
captncraig: `filepath.Join`. If you ever type a `/` as a path seperator you should feel bad.

cmd/scollector/collectors/ifstat_linux.go
```diff
@@ -106,3 +112,162 @@ func c_ifstat_linux() (opentsdb.MultiDataPoint, error) {
...
+				return nil
+			}
+			f[0] = strings.TrimSpace(f[0])
+			f[1] = strings.TrimSpace(f[1])
+			if f[0] == "Slave Interface" {
+				iface = f[1]
+				slaveCount++
+			}
+			// TODO: This will probably need to be updated for other types of bonding beside LACP, but I have no examples available to work with at the moment
+			if f[0] == "MII Status" && iface != "" {
```
captncraig: else if may make more sense here.

cmd/scollector/collectors/ifstat_linux.go
```diff
@@ -106,3 +112,162 @@ func c_ifstat_linux() (opentsdb.MultiDataPoint, error) {
...
+	bondDevices, _ := ioutil.ReadDir(bondingPath)
+	for _, fi := range bondDevices {
+		var iface string
+		var slaveCount int
+		if err := readLine(bondingPath+"/"+fi.Name(), func(s string) error {
+			f := strings.SplitN(s, ":", 2)
+			if len(f) != 2 {
+				return nil
+			}
+			f[0] = strings.TrimSpace(f[0])
```
captncraig: can we assign names to these fields instead of referencing f[0] and f[1] everywhere below?

cmd/scollector/collectors/ifstat_linux.go
```diff
@@ -106,3 +112,162 @@ func c_ifstat_linux() (opentsdb.MultiDataPoint, error) {
...
+		if err != nil {
+			return ts, err
+		}
+		err = json.NewDecoder(reader).Decode(&ts)
+		if err != nil {
+			return ts, err
+		}
+		return ts, nil
+	}
+	if _, err := os.Stat("/sys/module/team"); os.IsNotExist(err) {
```
captncraig: What if you get an error besides not exist?

cmd/scollector/collectors/ifstat_linux.go
kylebrandt: Agree on the checking error , but don't really feel like changing logic names etc on something that already works, this was just moved out of the proc_linux collector, but code is the same.

cmd/scollector/collectors/ifstat_linux.go
kylebrandt: Actually don't even need this check, /var/run/teamd is sufficient.

PR #1289, opened 2015-08-31
Influxdb query

nathanielc:
Here is an updated PR from the original 'influxdb-query' branch. It adds support for an `influx` function that can perform a query against InfluxDB and return the results as  a `seriesSet`.  It also adds support for triple quoted strings in expressions so that InfluxDB queries make full use of single and double quotes.

Other small changes where made from the original PR:

* Manipulating the InfluxQL AST instead of string manipulations
* Removed the `format` parameter since it required that the expression double list desired tags. Seems like the original reason for having it was to get around a bug in InfluxDB and is no longer necessary. See https://github.com/influxdb/influxdb/issues/3059


nathanielc:
Since this PR was a bit messy with changes from different people I have squashed all the changes and resubmitted the PR. see https://github.com/bosun-monitor/bosun/pull/1291

PR #1290, opened 2015-09-01
cmd/scollector: Cmdline flag to set reported hostname

kylebrandt:


gbrayut:
I think this was moved to be a conf file only feature, but if it is added back we should update the https://github.com/bosun-monitor/bosun/blob/master/cmd/scollector/doc.go docs as well.

kylebrandt:
I can't remember why I wanted this so just closing

PR #1296, opened 2015-09-03
cmd/scollector: Propagate FullHost setting

giganteous:
I noticed our scollectors reported different host tags for internal metrics.


captncraig:
Good catch. Thanks.

PR #1301, opened 2015-09-04
bosun: Better Error flow

captncraig:
Currently errors checking alerts get muddled with actual alerts in odd ways. Errors are not tagged, so are often difficult to correlate with alerts and hard to get to go away. This is a complete overhaul of errors.

1. Errors are no longer a state. We simply track errors that occur separately from alert data.
2. Dashboard will have a simple summary of active alerts that takes you to the error detail page.

3. Active alerts on the dashboard will get a flame icon to indicate a current error on the specified alert (error happened more recently than a success). This does not affect your ability to close alerts if the last known state was normal.

  screen shot 2015-09-04 at 3 59 58 pm

4. Errors must be "closed" before they will disappear from the system. The error page will list all errors that occured, and the number of times they occured. Once closed, they are forgotten entirely.

5. Errors are coalesced into sets by message. A continual sequence of checks that result in the same error will coalesce into a single line item with a count of events. If a successful check completes, no further errors will coalesce into the same line item on the error page. An entire line item is closed as a unit.


giganteous:
Seeing the nice fire icon with alerts that turned more emergent it'd be kinda cool to have a fire-extinguisher icon on alerts that turned into recoveries?

Issue #1298, opened 2015-09-04
Influx query language improvements


giganteous:
I'm trying to get better support into bosun, and for this I'm trying out different examples from the documentation page and adapting them to the influx() function.

Given the os.diskspace alert, we have this opentsdb query:
```
q("avg:6h-avg:os.disk.fs.percent_free{host=netlog*,disk=*}", "7d", "")
```

I think I have a similar query to get the data from influxdb, correct me if i'm wrong:
```
select mean(value) from "os.disk.fs.percent_free" WHERE host =~ /netlog.*/ and time > now() - 7d group by time(6h), host, disk
```

The current problem:

1) influx() wants the time outside the query, or it will complain with
```
  influx($db, '''select mean(value) from series WHERE host =~ /netlog.*/ group by time(6h), host, disk''', "7d", "")
```
  => **aggregate functions with GROUP BY time require a WHERE time clause**

Adding it, will complain too:
```
  influx($db, '''select mean(value) from series WHERE time < now() - 7d host =~ /netlog.*/ group by time(6h), host, disk''', "7d", "")
```
  => **influx query must not contain time in WHERE**

Is there a dependency somewhere in the code on this argument style? I.e: is it possible to ditch the start/end arguments for influx()?

giganteous:
I've pushed some changes to [my bosun repo in an experimental branch](https://github.com/giganteous/bosun/tree/less-arguments-for-influx). This gives me at least a function to simulate these same alerts with.

giganteous:
Ah. This is the influxTag() function [at line 27 of influx.go](https://github.com/bosun-monitor/bosun/blob/master/cmd/bosun/expr/influx.go#L27). The influx querylanguage must have some info on how to retrieve those. Perhaps @nathanielc can assist me?

nathanielc:
The two errors you are getting are an artifact of the way the query is constructed.

The first `aggregate functions with GROUP BY time require a WHERE time clause` is just complaining that you can't groupby time if you haven't also added  a where clause on time. Pretty self explanatory.

The second error `influx query must not contain time in WHERE`is coming from the bosun specific code saying that the `influx` method wants to add the time clause itself to the query and if there is already one its considered invalid.

The combination of these tow errors creates a cyclic dependency that means you can''t use group by time in your influx queries. 

[...]

PR #1304, opened 2015-09-07
cmd/Bosun: enable group by interval in the influx() query

giganteous:
A first throw at fixing the inconsistencies. See bug #1298. I think I've fixed the tests. Fighting with travis now.

### DIFF
--- a/cmd/bosun/expr/influx.go
+++ b/cmd/bosun/expr/influx.go
@@ -17,7 +17,7 @@ import (
 // Influx is a map of functions to query InfluxDB.
 var Influx = map[string]parse.Func{
 	"influx": {
-		Args:   []parse.FuncType{parse.TypeString, parse.TypeString, parse.TypeString, parse.TypeString},
+		Args:   []parse.FuncType{parse.TypeString, parse.TypeString, parse.TypeString, parse.TypeString, parse.TypeString},
 		Return: parse.TypeSeriesSet,
 		Tags:   influxTag,
 		F:      InfluxQuery,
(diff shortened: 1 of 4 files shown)


gbrayut:
If you rebase this on origin/master travis CI should start working. There is a new version of Esc as well so you will need to run `go get -u github.com/mjibson/esc` if you need to make any changes to the static content.

cmd/bosun/expr/influx.go
```diff
@@ -25,16 +25,27 @@ var Influx = map[string]parse.Func{
...
+		return nil, err
+	}
+	s, ok := st.(*influxql.SelectStatement)
+	if !ok {
+		return nil, fmt.Errorf("influx: expected select statement")
+	}
+
+	t := make(parse.Tags, len(s.Dimensions))
+	for _, d := range s.Dimensions {
+		if _, ok := d.Expr.(*influxql.Call); ok { // && call.Name == "time" {
```
gbrayut: Remove the commented out code // && call.Name == "time" {

gbrayut:
I left a note on a line that has some commented out code that should be removed.

Also the docs from https://github.com/bosun-monitor/bosun/issues/1298#issuecomment-139345964 should be commited into the /docs/expressions.md file as part of this PR, as we usually like to have doc changes in the same PR that the code is changed.

Other than that I tested this on my local instance and it looks good. The following query based on a counter matches what I would expect to get from OpenTSDB:
```
influx("opentsdb", '''SELECT non_negative_derivative(value) FROM "os.cpu" WHERE host =~ /gb.*/ GROUP BY host''', "30m", "", "1s")
```

Still not sure exactly how downsampling or fill values work for counters in Influx. If I change 1s to something larger the graph jumps above 100%, so I assume I would need to wrap the above query in another query if I wanted to get the equivalent of a 5m-avg downsample. I looks like Influx does this using something called [Continuous Queries](https://influxdb.com/docs/v0.9/query_language/continuous_queries.html).

gbrayut:
Found a github issue about [using rates in influxdb](https://github.com/influxdb/influxdb/issues/3976#issuecomment-138755447). looks like the non_negative_derivative function can take an optional argument and aggregate function. So on my test system that sends scollector data to both influxdb and opentsdb the following queries create the same graphs:

```
influx("opentsdb", '''SELECT non_negative_derivative(mean(value),1s) FROM "os.cpu" GROUP BY host''', "30m", "", "2m")

q("sum:2m-avg:rate{counter,,1}:os.cpu{host=*}", "30m", "")
```

giganteous:
Thanks! I'll rebase the work, and fix the issues.

gbrayut:
The docs look good here. Could probably use another rebase to pull in the new influx settings in the conf file, but otherwise LGTM.

giganteous:
Rebased.

I'm running our proof of concept system with the rebased code, and our version of the os.diskspace check looks to be okay.

gbrayut:
LGTM

Thanks for working on this!

Issue #1306, opened 2015-09-08
Typescript 1.6


giganteous:
tsc complains about the following:

```
$ go generate
2015/09/08 12:01:53 running tsc
web/static/js/0-bosun.ts(21,13): error TS2345: Argument of type '{ title: string; templateUrl: string; controller: string; }' is not assignable to parameter of type 'IRoute'.
  Object literal may only specify known properties, and 'title' does not exist in type 'IRoute'.
2015/09/08 12:01:55 run error: tsc: exit status 2
2015/09/08 12:01:55 run complete: tsc
2015/09/08 12:01:55 running esc
2015/09/08 12:01:55 run complete: esc
```

Could someone with typescript fu look at this?

giganteous:
updated title; upstream is 1.6 stable now.

kylebrandt:
Can you PR it or just comment here with the code changes?

PR #1307, opened 2015-09-08
travis: pin Typescript to version 1.5.3

giganteous:
The newest version is 1.6-beta, and is quite strict.

I've created a new issue #1306 for fix the actual code.


gbrayut:
The latest ESC is causing our CI checks to fail (see https://github.com/mjibson/esc/issues/15 ). We probably need to add a flag to esc that disables the ModTime feature.

gbrayut:
Fixed via #1308

PR #1308, opened 2015-09-08
travis: pin Typescript to version 1.5.3 and fix Esc ModTime issues

gbrayut:


gbrayut:
Looks like Esc is working again. We had to override the modtime feature that was added to Esc recently and It looks like on Windows you need to use autocrlf = input to prevent issues with line endings. I also updated the go generate call for Tsc to only replace the file if changes were actually made. I think it was updating the timestamp or line endings which caused some issues.

This is passing travis CI now, I'll have someone else test locally after running `go get -u github.com/mjibson/esc` to make sure I didn't break anything else then this should be good to merge

PR #1314, opened 2015-09-11
bosun: Reverse proxies setting host header properly

captncraig:


captncraig:
#1326 should fix this more thoroughly.

PR #1316, opened 2015-09-12
cmd/scollector: count running linux processes from linux.proc collector

### LINKED ISSUE #1079: Track if a process is running or not
Add some metric to help track if a process is running or not. I guess a simple way to do it is to send the number of processes currently matching each [[Process]] entry

dimamedvedev:
Kind of fixes #1079 . The goal is to have something to track if there are no running processes matching from [[Process]] entry and not use `since()`.

### DIFF
--- a/cmd/scollector/collectors/processes_linux.go
+++ b/cmd/scollector/collectors/processes_linux.go
@@ -39,12 +39,14 @@ func WatchProcesses() {
 
 func linuxProcMonitor(w *WatchedProc, md *opentsdb.MultiDataPoint) error {
 	var err error
+	var processCount int
 	for pid, id := range w.Processes {
 		file_status, e := os.Stat("/proc/" + pid)
 		if e != nil {
 			w.Remove(pid)
 			continue
 		}
+		processCount++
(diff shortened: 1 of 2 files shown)


gbrayut:
We would prefer to only count processes that are registered to run or have previously been seen on a system. We use the same toml file for all hosts to simply setup, so we wouldn't want a metric indicating that our mail server process is not running on our backup servers (since it never was suppose to). 

Maybe you can add an `IncludeCount = true` option to the Config/WatchedProc settings so that this feature is disable by default but can be enabled for specific entries. Example where it is not included for puppet-agent but is included for haproxy-t1:

```
[[Process]]
  Command = "ruby"
  Name = "puppet-agent"
  Args = "puppet"

[[Process]]
  Command = "/haproxy"
  Name = "haproxy-t1"
  Args = "/etc/haproxy-t1/haproxy-t1.cfg"
  IncludeCount = true
```

dimamedvedev:
@gbrayut updated PR accordingly

gbrayut:
LGTM!

PR #1318, opened 2015-09-14
cmd/bosun: Remove metrics from the API host route

kylebrandt:


captncraig:
Don't merge. I don't think this is the problem.

PR #1333, opened 2015-09-18
cmd/scollector. Add powered_state and connection_state for vpshere guests

kylebrandt:


cmd/scollector/collectors/vsphere.go
```diff
@@ -289,6 +291,34 @@ func vsphereGuest(vsphereHost string, v *vsphere.Vsphere, md *opentsdb.MultiData
...
 				}
+			case "VirtualMachinePowerState":
+				s := p.Val.Inner
+				var v int
+				switch s {
+				case "poweredOff":
+					v = 0
+				case "poweredOn":
+					v = 1
+				case "suspended":
```
captncraig: should we log something in the default cases here, so we know if they start sending unknown strings?

PR #1337, opened 2015-09-21
cmd/scollector: Setup a fake url when the -p option is given

### LINKED ISSUE #1270: scollector: -p still requires host
the -p flag is used for local debugging. If you run `scollector -p` you still get:

`2015/08/21 14:29:43 fatal: main.go:166: invalid host : no host specified`

Also, `-p` should probably imply `-m`

giganteous:
Fixes #1270.


captncraig:
fixes #1270

PR #1341, opened 2015-09-22
scollector: enabling json or toml config

captncraig:
After some conversation in stack chat, we decided we don't like toml so much. This allows toml or json files to be passed in. 

At some point in the future we may depricate the toml file format.

This also removes the totoml functionality. If that is needed, the 0.4.0 release of scollector should have it availible.

### DIFF
--- a/cmd/scollector/conf/conf.go
+++ b/cmd/scollector/conf/conf.go
@@ -7,48 +7,48 @@ import (
 
 type Conf struct {
 	// Host is the OpenTSDB or Bosun host to send data.
-	Host string
+	Host string `json:",omitempty"`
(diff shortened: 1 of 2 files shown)


kylebrandt:
Closing this after talking to Craig. The frustrations with toml are coming in when we do more complex data structures, in particular for SNMP definitions. 

I think since this is the problem, we are better keeping the SNMP defs out of the config and treating them as loadable object definitions. This also makes them more sharable (other people may have the same or similar PDUs for example). 

Those definitions could be in JSON, Craig would rather have them go in the scollector code. I'm okay with either as long as keep the main conf clean and avoid switching conf formats yet again unless there is a strong need to.

PR #1342, opened 2015-09-22
scollector: handle SIGINT for somewhat graceful shutdown

captncraig:
On SIGINT scollector will:

1. attempt to stop all running collectors. Any collector in a waiting state will immediately return.
2. Attempt to flush all queued datapoints to opentsdb.
3. Exit no matter what after 5 seconds.


alienth:
lgtm :fish:

PR #1343, opened 2015-09-22
cmd/scollector: Get bridge information from SNMP net devices

kylebrandt:


kylebrandt:
Going to sit on this until CDP is exploreded

gbrayut:
is the mac metadata going to conflict with the existing data we already set?

image

We don't want them overwriting each other. Not sure if we use snmp for hosts or just network gear

gbrayut:
also let me know what format you use... we should have all three report the same way

kylebrandt:
We don't mix polling and scollector. I'm going to updated everything to use
the all caps, no separator format windows is using

cmd/scollector/conf/conf.go
```diff
@@ -82,6 +82,7 @@ type SNMP struct {
 	Community string
 	Host      string
 	MIBs      []string
+	Bridge    bool
```
captncraig: We should not add another field for this. Rather add "bridge" to the list of built in collectors here:https://github.com/bosun-monitor/bosun/blob/master/cmd/scollector/collectors/snmp.go#L17

cmd/scollector/collectors/snmp_ifaces.go
```diff
@@ -125,6 +134,7 @@ func c_snmp_ifaces(community, host string) (opentsdb.MultiDataPoint, error) {
 			}
 			Add(&md, switchInterfaceMetric(sA.metric, ifNames[k], ifTypes[k]), v, tags, sA.rate, sA.unit, sA.desc)
 			metadata.AddMeta("", tags, "alias", ifAliases[k], false)
+			metadata.AddMeta("", tags, "mac", ifPhysAddresses[k], false)
```
captncraig: should we take this opportunity to reduce the set of tags we are adding metadata for? Do we really need iname and direction in the metadata tagset?

kylebrandt: Will do in another PR

PR #1345, opened 2015-09-23
cmd/bosun: add complete influx cred config, docs: update docs to reflect new config

nathanielc:
Adds configuration so that you can connect to InfluxDB with TLS and/or a username and password.

### DIFF
--- a/cmd/bosun/conf/conf.go
+++ b/cmd/bosun/conf/conf.go
@@ -18,6 +18,7 @@ import (
 	"time"
 
 	"bosun.org/_third_party/github.com/MiniProfiler/go/miniprofiler"
+	"bosun.org/_third_party/github.com/influxdb/influxdb/client"
 	"bosun.org/cmd/bosun/conf/parse"
 	"bosun.org/cmd/bosun/expr"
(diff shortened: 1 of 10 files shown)


gbrayut:
LGTM. 

Compiled an ran it locally with no changes to bosun/influxdb settings and it still works as expected. Changed /etc/opt/influxdb/influxdb.conf to use auth-enabled = true under the http section, used /opt/influxdb/influx to create a user for bosun, and added the following settings to the bosun config file

```
[root@gbcentos7 influxdb]# head /home/gbray/go/src/bosun.org/cmd/bosun/gbcentos7.conf
tsdbHost = gbcentos7:4242
influxHost = gbcentos7:8086
influxUsername = bosun
influxPassword = bosun
graphiteHost = gbcentos7:81
smtpHost = gbcentos7:25
emailFrom = bosun@localhost
httpListen = :80
timeAndDate = 202,75,179,136
stateFile = /home/gbray/go/src/bosun.org/cmd/bosun.state
```

After restarting bosun and influxdb the influx("opentsdb",'''SELECT value from "os.mem.percent_free" group by host''',"5m","") expression still works as expected.

PR #1347, opened 2015-09-24
Dependencies for ledis branch

captncraig:


captncraig:
Going to merge dependencies into _third_party before the other branch rather than after. Ignore this.

PR #1350, opened 2015-09-25
cmd/bosun: Allow to query graphite with Auth informations

### LINKED ISSUE #1349: Graphite query with basic auth
i've set up bosun with 
 graphiteHost to smth like "https://user:pass@myhost/render" 
but it seems to still have a 401.... 

my logs says 
```graphite RequestError (https://myhost/render?format=json&from=1443166442): Get failed: 401 Authorization Required```

it seems that somewhere it drop the auth part.

eMerzh:
a small patch to resolve #1349


gbrayut:
Tested and it is working as expected (with and without the credentials and auth requirement)

Thanks!

PR #1358, opened 2015-09-30
cmd/bosun: Collect ack status info by notification

kylebrandt:


cmd/bosun/sched/check.go
```diff
@@ -306,6 +310,8 @@ func (s *Schedule) CollectStates() {
 	severityCounts := make(map[string]map[string]int64)
 	abnormalCounts := make(map[string]map[string]int64)
 	ackStatusCounts := make(map[string]map[bool]int64)
+	ackByNotificationCounts := make(map[string]map[bool]int64)
```
captncraig: the map[bool] here is still odd. I think I would rather see ackCounts and unackedCounts maps[string]int

cmd/bosun/sched/check.go
```diff
@@ -323,10 +329,35 @@ func (s *Schedule) CollectStates() {
 		ackStatusCounts[alert.Name][true] = 0
 		activeStatusCounts[alert.Name][true] = 0
 	}
+	for notificationName, _ := range s.Conf.Notifications {
+		unAckOldestByNotification[notificationName] = time.Unix(1<<63-62135596801, 999999999)
```
captncraig: wtf is this number?

cmd/bosun/sched/check.go
kylebrandt: I believe it follows the existing pattern though, no?

cmd/bosun/sched/check.go
kylebrandt: Mac value of time.Time, so I can loop and get the min. http://stackoverflow.com/a/32620397/107156

PR #1361, opened 2015-09-30
bosun: search keeps set of all metric/tagsets

captncraig:
@kylebrandt


kylebrandt:
Can you hook the web api back into this?

PR #1365, opened 2015-10-01
Send the linux net interfaces bits count

gregdel:


kylebrandt:
Why is this needed? You can multiply the series by 8 in Bosun.

gregdel:
I'm just using scollector to push metrics to opentsdb. 
For network metrics it's easier to work with bits instead of bytes.

dimamedvedev:
It's probably easier when one works mostly with network devices in mind, situation differs quite a bit when e.g. comparing process disk vs. network io.

gregdel:
Rebased to avoid conflicts

kylebrandt:
Since this tends to be a lot of data, I don't want to double the data. Also network information ties into the host API and other expectations in bosun, so don't want to change it at this point. Since expressions make this manipulation straight forward. I am aware bits are more common for consumption, but nearly everything exposes this as octets, and multiplication could overflow types in theory. Appreciate the PR even though we are not accepting it - sorry and hope you are not discouraged from future contributions.

PR #1366, opened 2015-10-02
Update the ifstat regexp to match all the linux interfaces

gregdel:
Related to https://github.com/bosun-monitor/bosun/pull/1353

I'm not sure why the regexp was limited to some interfaces in the first place. Should scollector match some network interfaces or _all_ the network interfaces ??

What do you think ?

### DIFF
--- a/cmd/scollector/collectors/ifstat_linux.go
+++ b/cmd/scollector/collectors/ifstat_linux.go
@@ -43,8 +43,7 @@ var netFields = []struct {
 	{"compressed", metadata.Counter, metadata.Count},
 }
 
-var ifstatRE = regexp.MustCompile(`\s+(eth\d+|em\d+_\d+/\d+|em\d+_\d+|em\d+|` +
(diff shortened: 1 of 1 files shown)


kylebrandt:
The problem is that the number of tags slows down OpenTSDB. So it is prudent to filter out stuff that isn't important to monitor by default.

gregdel:
So what's the best way to add custom interface names ?

I did a PR to make the regexp more readable: https://github.com/bosun-monitor/bosun/pull/1367, but I'm not sure it's the right approach to extend the regexp to add a new interface. It's still more readable though.

Maybe we should add a new parameter in the configuration file to allow custom interface matching ?

gregdel:
Wrong PR link in my last comment, here's the good one: https://github.com/bosun-monitor/bosun/pull/1353

kylebrandt:
I like the idea of making it configurable. However, since OpenTSDB does aggregation, we separate different classes of interfaces into different metrics (i.e. bonds become os.net.bond.bytes where as os.net.bytes should only be physical interfaces.

So if you make this configurable, the configuration will need to provide a way to make sure they get classified appropriately.

PR #1367, opened 2015-10-02
Add metricFilters to filter only the metrics we want to send

PouuleT:
Add metricFilters to filter only the metrics we want to send
If no filters are specified, nothing is filtered to keep
current behaviour

### DIFF
--- a/cmd/scollector/doc.go
+++ b/cmd/scollector/doc.go
@@ -101,6 +101,9 @@ Default is 500.
 
 Filter (array of string): filters collectors matching these terms.
 
+MetricFilters (array of string): filters metrics matching these regular
+expressions.
+
 PProf (string): optional IP:Port binding to be used for debugging with pprof.
 Examples: localhost:6060 for loopback or :6060 for all IP addresses.
 
--- a/cmd/scollector/collectors/collectors.go
+++ b/cmd/scollector/collectors/collectors.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"regexp"
 	"sort"
 	"strings"
 	"sync"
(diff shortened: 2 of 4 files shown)


PouuleT:
There is more and more cloud services that allow people to easily send metrics, but by default, even while filtering collectors with the -f option, there is still way to many metrics sent by default
I think it would be great to be able to send only the few metrics we really want

What do you think?

cmd/scollector/doc.go
```diff
@@ -24,6 +24,9 @@ The flags are:
 	-f=""
 		Filters collectors matching these terms, separated by
 		comma. Overrides Filter in conf file.
+	-F=""
```
captncraig: We have been moving a lot of things into the conf file only, and removing the flag way to do them. Do you have a compelling reason why this needs to be a flag, and not just specified in the config file? What do you think @gbrayut?

PouuleT: I don't have any specific reason to keep the flag, it was just easier for me to test it this way. I'll delete it

PouuleT: I've just deleted it from the flags

digrich:
@captncraig In which release, we can get this change? Do we have daily binary build for scollector to try  this change?

captncraig:
It will be in the 0.5.0 release which will probably not be out until a few weeks from now, due to the redis conversions I am working on.

I can make a prerelease build for this and some other things.

digrich:
It would be good if u can make a prerelease, so we can start testing.

captncraig:
https://github.com/bosun-monitor/bosun/releases/tag/0.5.0-alpha

digrich:
Thanks @captncraig

gbrayut:
Just adding an example here for reference:

you can use ^ and $ to match the start and end of a string, ( a | b ) to say a or b, and .​* to match anything.  

`MetricFilters = ['^(win\.cpu|win\.system\..*)$', 'free']` would match win.cpu, win.system.*​ or any metric that includes the word free

PR #1369, opened 2015-10-02
Updating dependencies

captncraig:


giganteous:
You could change the import for snmp to bosun-monitor/snmp now? I don't know if it fits in your schedule though.

captncraig:
Yeah, I'll be doing that. Actually playing with godep instead of party for vendoring at the moment. I'll move snmp as part of that.

giganteous:
godep or gb :). :+1: for anything that uses a 'vendor' dir, and no rewriting.

captncraig:
@giganteous we are probably going to stick with rewriting until GOVENDOREXPERIMENT becomes a default. Maybe. It is important to us that `go get bosun.org.cmd.bosun` works for people.

Setting the env var is not difficult, but it is another step. Bring it up in slack if you feel strongly.

PR #1370, opened 2015-10-02
Trying out godep.

captncraig:
We are currently using party for vendoring packages.

This has a few pain points:

1. No way to update only a single package. Makes re-vendoring packages take a long time with the number of dependencies we have. 
2. No manifest of which revisions we have of the dependencies. Makes it hard to correlate vendored code with upstream repo.
3. GOVENDOREXPERIMENT support. I don't think we will use this until go 1.6, but godeps has a clearer way forward on it.

I propose switching to godep, still rewriting imports.

Primary difference will be imports will be `bosun.org/Godeps/workspace/src/github.com/whoever/whatever` instead of `bosun.org/_third_party/github.com/whoever/whatever`


captncraig:
I don't think this will work fully until godep fixes https://github.com/tools/godep/issues/271. It only copies files with the current build tags. We have some linux only files and some windows only files, etc.

captncraig:
Closing this. If godep gets cross-platform support may reconsider. https://github.com/kardianos/govendor looks to be another viable alternative.

PR #1374, opened 2015-10-05
Add collector for systemd service status.

alienth:
Utilizes dbus API to determine state of services managed by systemd. It will track the state of services listed under `[[SystemdService]]` entries in the config (modeled off of `processes_windows` collector).

It writes metric information to `linux.systemd.unit`, and to `os.service.running`. The `systemd` metric tracks each possible `ActiveState` of a systemd service unit. The `os` metric simply tracks whether the service is running or not.

If `WatchProc` was set to `true` in the config, it also attempts to lookup the service process and sends process metrics via the `processes_linux` collector methods.

Example config:

```
[[SystemdService]]
   Name = "scollector"
   WatchProc = true

[[SystemdService]]
   Name = "redis-.*"

```

### DIFF
--- a/cmd/scollector/collectors/dotnet_windows.go
+++ b/cmd/scollector/collectors/dotnet_windows.go
@@ -9,6 +9,7 @@ import (
 	"bosun.org/cmd/scollector/conf"
 	"bosun.org/metadata"
 	"bosun.org/opentsdb"
+	"bosun.org/util"
 )
 
 var regexesDotNet = []*regexp.Regexp{}
@@ -60,7 +61,7 @@ func c_dotnet_loading() (opentsdb.MultiDataPoint, error) {
 	}
 	var md opentsdb.MultiDataPoint
 	for _, v := range dst {
-		if !nameMatches(v.Name, regexesDotNet) {
+		if !util.NameMatches(v.Name, regexesDotNet) {
 			continue
 		}
 		id := "0"
@@ -132,7 +133,7 @@ func c_dotnet_memory() (opentsdb.MultiDataPoint, error) {
 		var name string
 		service_match := false
 		iis_match := false
-		process_match := nameMatches(v.Name, regexesDotNet)
+		process_match := util.NameMatches(v.Name, regexesDotNet)
 		id := "0"
 		if process_match {
 			raw_name := strings.Split(v.Name, "#")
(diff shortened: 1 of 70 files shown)


cmd/scollector/collectors/dotnet_windows.go
```diff
@@ -60,7 +60,7 @@ func c_dotnet_loading() (opentsdb.MultiDataPoint, error) {
 	}
 	var md opentsdb.MultiDataPoint
 	for _, v := range dst {
-		if !nameMatches(v.Name, regexesDotNet) {
+		if !util.NameMatches(v.Name, regexesDotNet) {
```
gbrayut: This file is missing a "bosun.org/util" import. Adding that should fix travis ci builds

cmd/scollector/main.go
```diff
@@ -117,6 +117,9 @@ func main() {
 	for _, p := range conf.Process {
 		check(collectors.AddProcessConfig(p))
 	}
+	for _, p := range conf.SystemdService {
```
captncraig: I made a new way to do this kind of init stuff, but unfortunately didn't convert all these old ones in main. See what I did to the haproxy one here: https://github.com/bosun-monitor/bosun/blob/master/cmd/scollector/collectors/haproxy_unix.go#L16

That should make the collectors more self contained.

cmd/scollector/collectors/systemd_linux.go
```diff
@@ -0,0 +1,137 @@
...
+
+type systemdServiceConfig struct {
+	regex *regexp.Regexp
+	watch bool
+}
+
+var systemdServices []*systemdServiceConfig
+
+func init() {
+	// The following two func defs have no-op stubs in collectors.go, as they are
```
captncraig: the self-registering of init functions should alleviate this problem as well.

cmd/scollector/collectors/systemd_linux.go
```diff
@@ -0,0 +1,137 @@
...
+				if config.regex.MatchString(shortName) {
+					if config.watch {
+						err = watchSystemdServiceProc(&md, conn, unit)
+						if err != nil {
+							return nil, err
+						}
+					}
+					systemdTags := opentsdb.TagSet{"name": unit.Name}
+					osTags := opentsdb.TagSet{"name": shortName}
+					Add(&md, "linux.systemd.unit.activestate", activeState[unit.ActiveState], systemdTags, metadata.Gauge, metadata.StatusCode, descActiveState)
```
captncraig: If we get an unexpected state, it will register as active (0). Perhaps we should check before this and replace it with a higher value just to be safe?

gbrayut:
I would add osServiceRunningDesc = "0: active, 1: inactive" as a const in scollector/collectors/collecors.go and use that for the os.service.running description. That way it can be shared with windows. You can also move the metric name to a const in that same file (ex: osServiceRunning = "os.service.running")

alienth:
@captncraig  / @gbrayut Requested changes made.

I pushed the followup commits as-is to make review easier. I plan to squash things down before merge.

gbrayut:
LGTM... will try and test it tomorrow.

alienth:
Squished down. I left the NameMatches change separate because my changes no longer require it, but it should still be done.

PR #1378, opened 2015-10-05
Scollector: Adding MySQL support

MichaelS11:
This is to add support for MySQL. Testing on Windows and Linux.

Updated one html file, and other document files. Not sure what else may need to be done for that.

Welcome feedback

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 cmd/bosun/dev.conf
 docs/.jekyll-metadata
+.project
--- a/cmd/scollector/collectors/continuous.go
+++ b/cmd/scollector/collectors/continuous.go
@@ -1,0 +1,89 @@
+package collectors
+
+import (
+	"reflect"
+	"regexp"
+	"runtime"
+	"time"
+
+	"bosun.org/collect"
+	"bosun.org/metadata"
+	"bosun.org/opentsdb"
+)
+
+type ContinuousCollector struct {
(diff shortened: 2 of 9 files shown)


_third_party/github.com/go-sql-driver/mysql
```diff
@@ -0,0 +1 @@
+Subproject commit 527bcd55aab2e53314f1a150922560174b493034
```
captncraig: Rather than using subprojects we have been using github.com/mjibson/party to copy the source tree into our own repo.

captncraig:
@MichaelS11 If I understand this correctly it is not collecting stats about mysql, but running queries against data populated by an application? 

I think this may be a good candidate for an external collector. I do think it is quite clever, but the schema and usability requirements might make it hard to be generally applicable. The docs are good though, so if you can make a good case we may include it.

MichaelS11:
@captncraig 

The main starting focus was about getting stats about MySQL. The way that I am aware of getting stats about the MySQL database is by running queries. Depending on what version of MySQL one has and how one is using MySQL, different types of queries may be wanted and/or needed.

For example, in the mysql-collector.md document, there is a "Additional queries that may be wanted" section. While the table called "tables" is slandered in most MySQL database version, not everyone may want all that detailed data about table size because of the number of tables and/or schemas they have.

That being said, yes, it is about being able to to run other queries as well. So someone can get application data if wanted.

MichaelS11:
@captncraig 

Created new pull request #1381

PR #1381, opened 2015-10-05
Scollector: Adding MySQL support

MichaelS11:
This is to add support to gather stats about MySQL. Testing on Windows and Linux.

Updated one html file, and other document files. Not sure what else may need to be done for that.

Welcome feedback


MichaelS11:
screenshot

Screenshot of a two node MySQL NDB cluster. These are just a handful stats gather from the basic queries in documents.

captncraig:
I actually like this a lot. It provides a way to convert sql data into relational data. I have a couple of initial thoughts before I even look too close at the code:

1. I would probably like this to be as database agnostic as possible. If there was a config option for `dbType` or something that allowed is to target `mysql` or `mssql` by changing the string and selecting the appropriate driver that would be pretty cool. Even if we just support mysql initially, that would be nice to have.

2. I am a bit uncomfortable with the queries being stored in the database themselves. How about just putting a list of queries into the config?

3. I like the convention-based approach to converting result sets to datapoints, automatically detecting tags and so forth. Can we make it even simpler? Can we autodetect if a timestamp is present in the second column or not? Why is it limited to 5 tags (not that I would ever want to use more, but it seems odd)? That kind of thing. Just like to keep needed config to what we absolutely need for the feature set. 

This looks like it could be a really nice addition to bridge the gap between relational data and time series data.

MichaelS11:
Awesome! :)

1) My thought was to add other databases after this one. Wanted to at least get something done and working so it could be presented. But now that you brought it up, was thinking as you, how much of the same code could be used? Besides the drivers of course, I don't think 100% of the same code can be used, however, probably most of the same code can be used. Was thinking of having a "base" agnostic database code then have smaller database particular code. Since it might change the current MS database code or etc., defiantly would like your detailed thoughts.

As a side note, wanted to add support for Oracle but that uses C drivers. This would mean additional setup for builds that may me unwanted. Not sure how you felt about that.

2) Why uncomfortable?

Each query needs to have name, query, description, and queryinterval at least. Maybe hastime as well. That started to be a very large config file depending on how many queries one wants. Plus also having to restart Scollector every time one wants to change them.

Just trying to think of the pros and cons of both. Not sure which place is "better". One of the pros of having the information in the conf file is if someone uses puppet for pushing changes out, that may be easier.

3) Not sure how one would automatically detect if it was a time stamp? Say that the first tag name can not be a number?

Limited to 5 tags because the default limit of OpenTSDB is 8 tags. The 3 other tags are host, db_name, and inst_id. Technically could remove the limit (remove the error check) and have it error somewhere else if there is an issue. Have not looked at all the code so was not sure what kind of error it would give.


Thank you for all your notes! :)

MichaelS11:
@captncraig 

If you can provide some guidance, happy to recode it so it is more general.

MichaelS11:
2) Thought about this more, will move it over to the configuration file.

alienth:
Could you please move the `_third_party addition` to a separate commit? (ideally before the new scollector code). Makes review and history diving easier.

I've also updated our contributing guidelines to reflect this.

alienth:
Shouldn't need to close the PR to split the commits. Just repushing to your branch will update the PR.

MichaelS11:
@alienth 
Sorry, not sure how to split it :/

MichaelS11:
Created new pull request #1391

PR #1382, opened 2015-10-05
metadata: moving shared constants between bosun and scollector into metadata package

kylebrandt:


metadata/shared.go
```diff
@@ -0,0 +1,14 @@
+package metadata // import "bosun.org/metadata"
```
captncraig: you shouldn't need to repeat the vanity import, but whatever.

Issue #766, opened 2015-03-06
Error State does not Send Notifications


kylebrandt:
We could probably just mirror the unknown template functionality.

Dieterbe:
i actually get unknown alerts when it errors (querying graphite)
how should this behave? getting informed if there were errors sounds good, but if my datasource is having issues i don't want to be reminded every single time for every single alert check. or will this follow the standard rules? (alert on state change, not when state is retained)

kylebrandt:
Maybe look at how the unknown limiting works. You can set it to X, and if
there are more than X unknown notifications in a single run they get
grouped into a single summary notification.

Same would would probably work for err.

On Wed, Mar 18, 2015 at 11:06 AM, Dieter Plaetinck <notifications@github.com

kylebrandt:
https://github.com/bosun-monitor/bosun/blob/master/cmd/bosun/sched/notify.go#L154

Dieterbe:
does #836 solve this?
i'm thinking as the admin i want to know what kind of errors happen in alert runs,
but also as owners of alerting rules i want to know if my checks failed to execute because of alerts.

so i'm thinking when errors happen during rule executions, potentially two different parties should be notified. thoughts?

captncraig:
No, #836 I would say makes this situation slightly worse. The unknown alerts will no longer fire, so that symptom is not going to help you. 

We need to formalize error notifications instead on just having them show on the dashboard I think.

I'm not sure what we want them to look like though.

Do you want a notification *every* time a graphite error occurs, or only after 3 in a row or something like that?

Dieterbe:
not every.
as alert owner i would stay follow the same rules as other bad states: trigger alert on first error, subsequent errors need no further alerts, until i acked or closed or whatever the thing is, when the state gets changes from error to warn or crit, or from warn or crit to err, those are also state changes that deserve an alert imho.

as an admin, when alerts happen we can wait until the end of the alerting run and then send all alerts at once.  we also probably shouldn't send an alert for every run that has alerts, or maybe we should. because figuring out what "state changes" (wrt the set of alerts in each run) should result in alerts gets tricky when you think of scenarios where there's some new errors or some less, or one that looks different than before, etc.

PR #1386, opened 2015-10-06
Sending alerts the first time an error occurs

captncraig:


kylebrandt:
Would fix https://github.com/bosun-monitor/bosun/issues/766

PR #1390, opened 2015-10-08
Fix hbase collector for 1.0.

alienth:
This is a quick fixup to get this collector working for us again.

Note that this entire collector needs a revamp, but for now I just want to get our metrics. I'm planning on a revamp later.


alienth:
Note that this eliminates the 'instance' tag, which is no longer available from the metric data, and was duplicative to begin with. I'll be up dating the hbase alerts we have to reflect this change.

gbrayut:
LGTM... you can merge and deploy when ready

PR #1391, opened 2015-10-08
Scollector: Adding third party MySQL driver: go-sql-driver

MichaelS11:
Scollector: Adding third party MySQL driver: go-sql-driver


MichaelS11:
@captncraig  & @alienth

This should address most of the previous comments.

Welcome comments / feedback.

MichaelS11:
@captncraig & @alienth 

Any thoughts on this?

MichaelS11:
@captncraig & @alienth 

Any thoughts on this?

captncraig:
Sorry, @MichaelS11. Been a bit busy. I think it is looking good. I'll give it a good thorough review in the next day or two.

cmd/scollector/collectors/database.go
```diff
@@ -0,0 +1,345 @@
...
+
+	_ "bosun.org/_third_party/github.com/go-sql-driver/mysql"
+	"bosun.org/cmd/scollector/conf"
+	"bosun.org/collect"
+	"bosun.org/metadata"
+	"bosun.org/opentsdb"
+	"bosun.org/slog"
+)
+
+type DatabaseQuery struct {
```
captncraig: Why are is there a `conf.DatabaseQuery` and a `collectors.DatabaseQuery` struct? Just use the conf one everywhere.

cmd/scollector/main.go
```diff
@@ -134,6 +134,9 @@ func main() {
 	for _, r := range conf.Riak {
 		check(collectors.Riak(r.URL))
 	}
+	for _, cfg := range conf.Database {
```
captncraig: This is the old way of initializing things. For new things we have a different way to do it. See https://github.com/bosun-monitor/bosun/blob/master/cmd/scollector/collectors/haproxy_unix.go#L16 for an example. This lets us add new things without touching main.

cmd/scollector/collectors/database.go
```diff
@@ -0,0 +1,345 @@
...
+func init() {
+	DatabaseVars.vaildDatabases = map[string]bool{"mysql": true}
+}
+
+func Database(c conf.Database) error {
+	return DatabaseAddCollector(c)
+}
+
+func DatabaseAddCollector(c conf.Database) error {
+	_, found := DatabaseVars.vaildDatabases[c.Type]
```
captncraig: I don't really like the DatabaseVars existing just for this check. Just start with `if c.Type != "mysql"` and we'll go from there as appropriate.

cmd/scollector/collectors/database.go
```diff
@@ -0,0 +1,345 @@
...
+	if !found {
+		return fmt.Errorf("%v: %v", "invalid Database Type", c.Type)
+	}
+	if c.DBName != "" {
+		cleaned, _ := opentsdb.Clean(c.DBName)
```
captncraig: Opentsdb's allowed characters probably are not the same as valid names for mysql databases. What are the rules for mysql?

cmd/scollector/collectors/continuous.go
```diff
@@ -0,0 +1,95 @@
...
+	ContinuousCollectorVars.nameWorkerChan = make(map[string]chan *ContinuousCollectorStats)
+}
+
+func (c *ContinuousCollector) Init() {
+	if c.init != nil {
+		c.init()
+	}
+}
+
+func (c *ContinuousCollector) Run(dpChan chan<- *opentsdb.DataPoint, quit <-chan struct{}) {
```
captncraig: Help me understand why continuous collector is needed. It looks like you are still running each query on an interval. Is it because you maintain a set of connections? `database/sql` should maintain good enough connection pooling that I wouldn't worry about that. 

Why won't interval collector work for this?

cmd/scollector/collectors/database.go
MichaelS11: They are not the same. How would you like to handle the time in the collectors.DatabaseQuery? Could convert the interval on each run if you want but the next seems more tricky.
	interval    time.Duration
	next        time.Time

cmd/scollector/collectors/database.go
MichaelS11: DBName is used in a tag, so it needs to conform to openTSDB rules.

cmd/scollector/collectors/database.go
MichaelS11: Sure, will change.

cmd/scollector/collectors/continuous.go
MichaelS11: It wasn't that the interval collector could not be made to work, it was it did not seem ideal.

If you had a query to run every 15 seconds and other query run every 20 seconds, how often would the interval collector have to be set to run? My understanding is you would have to have interval collector run every 5 seconds otherwise the timing would be off. If you had even more queries with different timings the interval collector would have to be set to every second to check if there is a query to be run.

Also what if had a query that ran every 20 seconds and another query that ran every 30 minutes that took 5 minutes to run. Would you run them in the same interval collector? If you did, would you run both the queries at the same time? If you did run them at the same time, the MultiDataPoint could not be returned until both were finished.

It just seems to me a better solution would be to have a continuous collector, just like the external files have.

cmd/scollector/main.go
MichaelS11: Will fix

MichaelS11: Done

cmd/scollector/collectors/database.go
MichaelS11: Done

MichaelS11:
@captncraig 

Have also added Oracle to the database Scollector, it is in it's own branch.

https://github.com/MichaelS11/bosun/tree/oracle

The reason it is in it's own branch is because all the Oracle drivers have to use C. Do not expect a pure Go Oracle database driver anytime soon, if ever. Adding C as a requirement to Scollector may not be wanted.

Welcome to hear your thoughts.

captncraig:
Yeah, we really want to avoid cgo. It really hurts our ability to cross
compile. I would recommend an external collector to Oracle users.

MichaelS11:
@captncraig 

Any more thoughts on this?

MichaelS11:
@captncraig 

Would you be so kind and review my comments?

cmd/scollector/collectors/continuous.go
captncraig: You don't have to do this all in one collector. You can start one interval collector per query and they will all run independently, and slow queries won't affect fast ones.

MichaelS11: If you had say 50 queries, then one would have 50 collectors. Plus say you have one scollector instance that was monitoring 5 databases, that would be 250 collectors for one instance. Sure, one could make it work, but it did not seem the best solution just to make sure to use the interval collector.

MichaelS11:
@captncraig 

Happy New Year!

MichaelS11:
The check failed, could I get some guidance on why?

MichaelS11:
@captncraig 

Would it be possible to get some help please?

captncraig:
sorry it has taken so long. Been taking forever on the branch I just merged. I'll take another good look at this today.

captncraig:
@MichaelS11 Thank you for being patient. Sorry I have not been super responsive on this. I have given this a lot of thought, and I think this is a much better fit for a separate app run as an [external collector](http://bosun.org/scollector/external-collectors), for a number of reasons:

- The external collector system is already set up to stream data in the way you have set up with the continuous collector. You can just print a json line anytime you have a datapoint. That feels better than trying to shoehorn it into an interval collector, as you have pointed out. Just put it in the collectors directory `/0` and scollector will start your app and keep it running.
- The config for this is kinda hairy, and it might be better to have it outside of scollector.toml anyway. I kinda dislike toml in general, and putting complex config in it is probably a mistake. As a seperate app, you can use json or whatever format makes sense for this use case.
- If you make a seperate repo, we'd be happy to link to it from our docs and let people know it is out there.
- You can have more freedom to change it without going through us.

Does that seem reasonable?

MichaelS11:
@captncraig 

Thank you giving this a lot of thought and responding to this request. I hear what you are saying, that this is not super simple like the rest of the collectors and understand your desire to keep things simple. That being said, I do not think it really adds that much complexity to the main repository. Sure it has more configuration then other collectors but the code behind the collector is still fairly small and relatively simple. I think having it as a build in collector will add a lot of value and draw to Scollector, being able to pull data from many different types of databases, starting with MySQL Also as a side note, if the code is already in Scollector to stream data externally, seems just as well to be able to stream data internally as well.

While I really do appreciate the offer to link to external collector project, at this time I am just interested in helping out other projects.

gbrayut:
I'd have to agree with Craig that this is a bit heavy to add directly to scollector. We really only support system diagnostic collectors in scollector, and I don't think we want to do generic queries in MySQL/MSSQL or any other datastore directly from scollector.

Like Craig mentioned, this code would make a good separate binary called scollector-mysql or scollector-sql (supporting multiple database types), that could live in the `/opt/scollector/collectors/0/` folder and have it's own config file in `/opt/scollector/collectors/etc/` and send datapoints to scollector over standard output using the [JSON data output format
](http://bosun.org/scollector/external-collectors). That way it can evolve as needed without adding surface area to the scollector code base.

kylebrandt:
As we get more collectors, this is kind of a general issue. I wonder if we could do some sort of custom build flags, and by default ignore certain collectors?

In this way, people could do custom builds with only the collectors that they have in the environment. It might be nice, but I'm not finding a way to implement this.

kylebrandt:
Ah, actually http://stackoverflow.com/questions/10646531/golang-conditional-compilation/10649653#10649653 , seems like it is doable

gbrayut:
I'm less worried about the binary size increase than the change in scope from buitin collectors generating diagnostic metrics to supporting general purpose sql and all the code that has to be added to support arbitrary sql queries at arbitrary intervals

PR #1405, opened 2015-10-20
cmd/bosun: Add metric stats to host API

kylebrandt:


cmd/bosun/sched/host.go
```diff
@@ -0,0 +1,809 @@
+package sched // import "bosun.org/cmd/bosun/sched"
```
captncraig: thanks for making a new file :)

cmd/bosun/sched/host.go
```diff
@@ -0,0 +1,809 @@
...
+	}
+	for name, host := range hosts {
+		host.Name = name
+		hostTagSet := opentsdb.TagSet{"host": host.Name}
+		hostMetadata, err := s.GetMetadata("", hostTagSet)
+		if err != nil {
+			slog.Error(err)
+		}
+		processHostIncidents(host, states, silences)
+		for _, ts := range icmpTimeOutTags {
```
captncraig: this is a lot of iteration. I each iteration over hosts, you iterate over all tags for a lot of metrics. Perhaps we can turn those arrays into maps by host before we enter the main loop.

cmd/bosun/sched/host.go
```diff
@@ -0,0 +1,809 @@
...
+				continue
+			}
+			// The host tag represents the polling source for these set of metrics
+			source, ok := ts["host"]
+			if !ok {
+				slog.Errorf("couldn't find source tag for icmp data for host %s", host.Name)
+			}
+			// 1 Means it timed out
+			timeout, timestamp, err := s.Search.GetLast("bosun.ping.timeout", ts.String(), false)
+			if err != nil && !(timestamp > 0) {
```
captncraig: I think you mean `err != nil || timestamp <= 0`

cmd/bosun/sched/host.go
```diff
@@ -0,0 +1,809 @@
...
+		}
+		for _, ts := range hwTempsTags {
+			if ts["host"] != host.Name {
+				continue
+			}
+			name, ok := ts["name"]
+			if !ok {
+				slog.Errorf("couldn't find name tag %s for host %s", host.Name, name)
+			}
+			tStatus, timestamp, err := s.Search.GetLast("hw.chassis.temps", ts.String(), false)
```
captncraig: I notice you almost never check errors on these calls. Are you ok with zero values? Use _ if you aren't going to care about a particular error.

cmd/bosun/sched/host.go
```diff
@@ -0,0 +1,809 @@
...
+				if m.Time.Before(time.Now().Add(-timeFilterAge)) {
+					continue
+				}
+				if !m.Tags.Equal(ts) {
+					continue
+				}
+				switch val := m.Value.(type) {
+				case string:
+					switch m.Name {
+					case "psMeta":
```
captncraig: maybe check for name = psMeta earlier in the metadata loop. It should be cheaper than tagset equals.

cmd/bosun/sched/host.go
```diff
@@ -0,0 +1,809 @@
...
+				status = "Ok"
+			}
+			if err == nil && timestamp > 0 {
+				host.Hardware.Storage.Batteries[id] = &Battery{
+					Status:            status,
+					StatusLastUpdated: timestamp,
+				}
+			}
+		}
+		for _, ts := range hwBoardPowerTags {
```
captncraig: if you make these maps then these loops become `for _,ts := range hwBoardPowerTags[host.Name]` and the if/continue is unnecessary.

cmd/bosun/sched/host.go
```diff
@@ -0,0 +1,809 @@
...
+			pd := &PhysicalDisk{}
+			if err == nil && timestamp > 0 {
+				pd.Status = status
+				pd.StatusLastUpdated = timestamp
+				host.Hardware.Storage.PhysicalDisks[id] = pd
+			}
+			for _, m := range hostMetadata {
+				if m.Time.Before(time.Now().Add(-timeFilterAge)) {
+					continue
+				}
```
captncraig: check name before comparing tagsets again.

cmd/bosun/sched/host.go
```diff
@@ -0,0 +1,809 @@
...
+					StatusLastUpdated: timestamp,
+				}
+			}
+		}
+		for _, ts := range hwControllersTags {
+			if ts["host"] != host.Name {
+				continue
+			}
+			id, ok := ts["id"]
+			if !ok {
```
captncraig: perhaps when making maps we could omit datapoints without required tags. That could eliminate a lot of these checks.

cmd/bosun/sched/host.go
```diff
@@ -0,0 +1,809 @@
...
+			if err == nil && timestamp > 0 {
+				c.Status = status
+				c.StatusLastUpdated = timestamp
+				host.Hardware.Storage.Controllers[id] = c
+			}
+			for _, m := range hostMetadata {
+				if m.Time.Before(time.Now().Add(-timeFilterAge)) {
+					continue
+				}
+				if !m.Tags.Equal(ts) {
```
captncraig: same issue here. let's continue as fast as possible.

cmd/bosun/sched/host.go
```diff
@@ -0,0 +1,809 @@
...
+		var timestamp int64
+		var cpu float64
+		if cpu, timestamp, err = s.Search.GetLast("os.cpu", hostTagSet.String(), true); err != nil {
+			cpu, timestamp, _ = s.Search.GetLast("cisco.cpu", hostTagSet.String(), false)
+		}
+		host.CPU.PercentUsed = cpu
+		host.CPU.StatsLastUpdated = timestamp
+		host.Memory.TotalBytes, timestamp, _ = s.Search.GetLast("os.mem.total", hostTagSet.String(), false)
+		host.Memory.UsedBytes, _, _ = s.Search.GetLast("os.mem.used", hostTagSet.String(), false)
+		host.Memory.StatsLastUpdated = timestamp
```
captncraig: can you assign `host.Memory.StatsLastUpdated` two lines above inline?

cmd/bosun/sched/host.go
kylebrandt: In the scenario, where would you recommend I check for the existence of the host in that particular tagset result?

cmd/bosun/sched/host.go
kylebrandt: In this scenario, where would you recommend I check for the existence of the host in that particular tagset result?

cmd/bosun/sched/host.go
```diff
@@ -0,0 +1,809 @@
...
+				continue
+			}
+			var iface *HostInterface
+			if name := m.Tags["iface"]; name != "" {
+				if host.Interfaces[name] == nil {
+					h := new(HostInterface)
+					host.Interfaces[name] = h
+				}
+				iface = host.Interfaces[name]
+			}
```
captncraig: if no iface tag, we should probably skip or continue a lot of this loop.

cmd/bosun/sched/host.go
captncraig: right after you fetch the data. Turn it into a `map[string][]Tagset`. only include tagsets with the tags you will be looking for later.

cmd/bosun/sched/host.go
captncraig: in any case iface will be nil if there is no iface tag.

cmd/bosun/sched/host.go
```diff
@@ -0,0 +1,809 @@
...
+			if name := m.Tags["iname"]; name != "" {
+				iface.Name = name
```
captncraig: this will panic if iname without iface ever occurs

cmd/bosun/sched/host.go
```diff
@@ -0,0 +1,809 @@
...
+						case 0:
+							host.VM.PowerState = "poweredOn"
+						case 1:
+							host.VM.PowerState = "poweredOff"
+						case 2:
+							host.VM.PowerState = "suspended"
+						}
+						host.VM.PowerStateLastUpdated = timestamp
+					}
+					//Is this safe?
```
captncraig: looks safe enough to me.

cmd/bosun/web/web.go
```diff
@@ -604,11 +604,14 @@ func Last(t miniprofiler.Timer, w http.ResponseWriter, r *http.Request) (interfa
 	if r.FormValue("counter") != "" {
 		counter = true
 	}
-	tags, err := opentsdb.ParseTags(r.FormValue("tagset"))
-	if err != nil {
-		return nil, err
-	}
-	return schedule.Search.GetLast(r.FormValue("metric"), tags, counter)
+	val, timestamp, err := schedule.Search.GetLast(r.FormValue("metric"), r.FormValue("tagset"), counter)
```
captncraig: no ui depends on this, correct?

PR #1406, opened 2015-10-20
rule page won't init new search.

captncraig:


kylebrandt:
Add comment on sched's init warning that it gets called from userland

PR #1407, opened 2015-10-20
Add nexpose collector.

alienth:
Looking for an initial review pass on this collector.

Collector is configured as follows:

```
[[Nexpose]]
Username = "nexposeacct"
Password = "nexposesucks"
Host = "nexposehost.blah.com"
Insecure = false
```

For each monitored asset, it sends metrics with the vuln count, malware count, risk score, exploit count, and last scan time (represented in seconds since last scan). The asset `name` is either the detected hostname, or the IP address of the host.

For each site, it sends `nexpose.site` metrics which are an aggregate of the aforementioned device metrics.

It also sends a `nexpose.scan.running` metric indicating if a scan is currently running on a site.

Note that not all of the defined functions are used at this time. They'll probably come in handy for other metrics, so I'd like to keep em in. I'll probably end up creating a generic Go package for nexpose and rip a lot of that stuff out of this collector.


nickman:
Very cool. Forwarding to my infosec group...

cmd/scollector/collectors/nexpose.go
```diff
@@ -0,0 +1,405 @@
...
+	URLv1     string
+	URLv2     string
+	Client    *http.Client
+	SessionID string
+	Cookie    *http.Cookie
+}
+
+func c_nexpose(username, password, host string, insecure bool) (opentsdb.MultiDataPoint, error) {
+	tr := &http.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: insecure},
```
captncraig: Just noting the irony of even having an insecure option for a vulnerability tool.

cmd/scollector/collectors/nexpose.go
```diff
@@ -0,0 +1,405 @@
...
+	}
+	client := &http.Client{Transport: tr}
+	url1 := fmt.Sprintf("https://%s/api/1.1/xml", host)
+	url2 := fmt.Sprintf("https://%s/api/1.2/xml", host)
+
+	var md opentsdb.MultiDataPoint
+
+	c := nexposeConnection{Username: username, Password: password, URLv1: url1, URLv2: url2, Host: host, Client: client}
+
+	err := c.login()
```
captncraig: you can condense a lot of these to the inline if format:
```
if err := c.Login(); err != nil{
    return nil, fmt.Errorf("Login failed: %s", err)
}
```

cmd/scollector/collectors/nexpose.go
```diff
@@ -0,0 +1,405 @@
...
+	err := json.Unmarshal(*jsonData, &newTable)
+	if err != nil {
+		return err
+	}
+	t.Records = append(t.Records, newTable.Records...)
+
+	return nil
+}
+
+func (c *nexposeConnection) getJsonTable(location string, params map[string]string) (*jsonTable, error) {
```
captncraig: these functions are a bit unorganized. I would like to see all methods on a type somewhat grouped together.

cmd/scollector/collectors/nexpose.go
```diff
@@ -0,0 +1,405 @@
...
+			Add(&md, "nexpose.asset.exploit_count", asset.ExploitCount, assetTags, metadata.Gauge, metadata.Vulnerabilities, descExploitCount)
+			Add(&md, "nexpose.asset.malware_count", asset.MalwareCount, assetTags, metadata.Gauge, metadata.Vulnerabilities, descMalwareCount)
+			Add(&md, "nexpose.asset.last_scan", last_scan, assetTags, metadata.Gauge, metadata.Second, descAssetLastScan)
+
+			site.VulnCount += asset.VulnCount
+			site.ExploitCount += asset.ExploitCount
+			site.MalwareCount += asset.MalwareCount
+		}
+
+		Add(&md, "nexpose.site.vuln_count", site.VulnCount, tags, metadata.Gauge, metadata.Vulnerabilities, descVulnCount)
```
captncraig: would it make sense to have an averaged risk score per site?

cmd/scollector/collectors/nexpose.go
```diff
@@ -0,0 +1,405 @@
...
+		Add(&md, "nexpose.site.exploit_count", site.ExploitCount, tags, metadata.Gauge, metadata.Vulnerabilities, descExploitCount)
+		Add(&md, "nexpose.site.malware_count", site.MalwareCount, tags, metadata.Gauge, metadata.Vulnerabilities, descMalwareCount)
+	}
+
+	activeScans, err := c.scanActivity()
+	for _, scan := range *activeScans {
+		if scan.Status == "running" {
+			Add(&md, "nexpose.scan.running", 1, opentsdb.TagSet{"site": siteNames[scan.SiteID]}, metadata.Gauge, metadata.Bool, descScanRunning)
```
captncraig: should this be a count rather than 0/1?

cmd/scollector/collectors/nexpose.go
alienth: Probably not. There are a huge number of devices for some sites and the risk scores have a humongous range, so average loses meaning pretty quickly. Median might be worth considering.

cmd/scollector/collectors/nexpose.go
```diff
@@ -0,0 +1,405 @@
...
+		}
+	}
+
+	return md, nil
+}
+
+const (
```
captncraig: I'd be ok defining these consts inside c_nexpose

cmd/scollector/collectors/nexpose.go
```diff
@@ -0,0 +1,405 @@
...
+	descScanRunning   = "Nexpose scan running."
+	descRiskScore     = "Risk score for a given site/device."
+	descVulnCount     = "Number of known vulnerabilities."
+	descExploitCount  = "Number of vulnerabilities exploitable via Metasploit. Subset of vuln_count."
+	descMalwareCount  = "Number of vulnerabilities susceptible to malware attacks. Subset of vuln_count."
+	descAssetLastScan = "How many seconds ago this asset was last scanned."
+)
+
+func (c *nexposeConnection) jsonRequest(method string, location string, v *url.Values) (*[]byte, error) {
```
captncraig: there is nothing json about this.

cmd/scollector/collectors/nexpose.go
alienth: It's tagged on site, so count doesn't make much sense.

cmd/scollector/collectors/nexpose.go
```diff
@@ -0,0 +1,405 @@
...
+	req, _ := http.NewRequest(method, "https://"+c.Host+location, strings.NewReader(v.Encode()))
+
+	// Some methods require the cookie, some require the header, so we set both.
+	req.AddCookie(c.Cookie)
+	req.Header.Add("nexposeCCSessionID", c.SessionID)
+	req.Header.Add("Content-Type", "application/x-www-form-urlencoded")
```
captncraig: I know their api is garbage, but setting Content-Type on a GET should definitely be an error.

cmd/scollector/collectors/nexpose.go
```diff
@@ -0,0 +1,405 @@
...
+
+	resp, err := c.Client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
```
captncraig: we could lose some of this whitespace.

cmd/scollector/collectors/nexpose.go
```diff
@@ -0,0 +1,405 @@
...
+}
+
+type jsonTable struct {
+	RowsPerPage  int
+	RecordOffset int
+	TotalRecords int
+	Records      []json.RawMessage
+}
+
+func (t *jsonTable) concat(jsonData *[]byte) error {
```
captncraig: I'm not seeing why this function needs `*[]byte` as opposed to simply `[]byte`

alienth: Is it not preferable to pass by reference rather than by value?

cmd/scollector/collectors/nexpose.go
```diff
@@ -0,0 +1,405 @@
...
+	var newTable jsonTable
+	err := json.Unmarshal(*jsonData, &newTable)
+	if err != nil {
+		return err
+	}
+	t.Records = append(t.Records, newTable.Records...)
```
captncraig: The other three fields in jsonTable will be inconsistent after this. It probably doesn't matter.

cmd/scollector/collectors/nexpose.go
captncraig: slices and maps are always passed by reference.

cmd/scollector/collectors/nexpose.go
```diff
@@ -0,0 +1,405 @@
...
+
+	return nil
+}
+
+func (c *nexposeConnection) getJsonTable(location string, params map[string]string) (*jsonTable, error) {
+	v := url.Values{}
+	v.Set("dir", "ASC")
+	v.Set("startIndex", "0")
+	v.Set("results", "500")
+
```
captncraig: whitespace again

cmd/scollector/collectors/nexpose.go
alienth: It is a request to their internal JSON API, which will return a JSON reseponse. The non-JSON apis don't work like this (that header I add only works on the JSON api and will throw errors on the other APIs).

cmd/scollector/collectors/nexpose.go
```diff
@@ -0,0 +1,405 @@
...
+		v.Set(param, value)
+	}
+
+	body, err := c.jsonRequest("POST", location, &v)
+	if err != nil {
+		return nil, err
+	}
+
+	var table jsonTable
+	err = json.Unmarshal(*body, &table)
```
captncraig: err not checked

cmd/scollector/collectors/nexpose.go
```diff
@@ -0,0 +1,405 @@
...
+	total := table.TotalRecords
+
+	for len(table.Records) < total {
+		v.Set("startIndex", fmt.Sprintf("%d", len(table.Records)))
+		body, err = c.jsonRequest("POST", location, &v)
```
captncraig: these can be the `if err := foo(); err != nil{}` style

cmd/scollector/collectors/nexpose.go
```diff
@@ -0,0 +1,405 @@
...
+	login := loginRequest{UserID: c.Username, Password: c.Password}
+	resp, err := c.xmlRequest(&login, 1)
+	if err != nil {
+		return err
+	}
+
+	c.SessionID = resp.SessionID
+
+	// Set the cookie as well, in case we want to use the undocumented JSON API.
+	// Note that this cookie must not be sent to XML requests. It is only added
```
captncraig: Must not be sent to xml endpoints? You can't make this up.

alienth: Yeahhh.. you get a stack trace back from nexpose if you add it :)

captncraig:
For dealing with 2 versions of xml apis and the worst json api I've ever seen, I'd say it looks pretty good.

alienth:
@captncraig Requested changes made. I've left the fixup commits separate to facilitate easy review. Will squash before merge.

captncraig:
@alienth I really like how you address individual issues with separate commits. Good call. Merge at will.

PR #1408, opened 2015-10-21
Add ciscobgp snmp collector

shanemadden:
Add SNMP collector for the gathering of BGP health data from Cisco devices (CISCO-BGP4-MIB).

### DIFF
--- a/cmd/scollector/collectors/snmp_ciscobgp.go
+++ b/cmd/scollector/collectors/snmp_ciscobgp.go
@@ -1,0 +1,225 @@
+package collectors
+
+import (
+	"fmt"
+	"io"
+	"strings"
+	"time"
+
+	"bosun.org/_third_party/github.com/mjibson/snmp"
+	"bosun.org/cmd/scollector/conf"
+	"bosun.org/metadata"
+	"bosun.org/opentsdb"
+)
+
+func SNMPCiscoBGP(cfg conf.SNMP) {
+	collectors = append(collectors, &IntervalCollector{
+		F: func() (opentsdb.MultiDataPoint, error) {
+			return c_snmp_ciscobgp(cfg.Community, cfg.Host)
+		},
+		Interval: time.Second * 30,
+		name:     fmt.Sprintf("snmp-ciscobgp-%s", cfg.Host),
+	})
+}
+
+func c_snmp_ciscobgp(community, host string) (opentsdb.MultiDataPoint, error) {
+	const (
+		state               = ".1.3.6.1.4.1.9.9.187.1.2.5.1.3.1.4"
+		adminStatus         = ".1.3.6.1.4.1.9.9.187.1.2.5.1.4.1.4"
+		localAS             = ".1.3.6.1.4.1.9.9.187.1.2.5.1.8.1.4"
+		localIdentifier     = ".1.3.6.1.4.1.9.9.187.1.2.5.1.9.1.4"
+		remoteAS            = ".1.3.6.1.4.1.9.9.187.1.2.5.1.11.1.4"
+		remoteIdentifier    = ".1.3.6.1.4.1.9.9.187.1.2.5.1.12.1.4"
+		inUpdates           = ".1.3.6.1.4.1.9.9.187.1.2.5.1.13.1.4"
+		outUpdates          = ".1.3.6.1.4.1.9.9.187.1.2.5.1.14.1.4"
+		inTotalMessages     = ".1.3.6.1.4.1.9.9.187.1.2.5.1.15.1.4"
+		outTotalMessages    = ".1.3.6.1.4.1.9.9.187.1.2.5.1.16.1.4"
+		inUpdateElapsedTime = ".1.3.6.1.4.1.9.9.187.1.2.5.1.27.1.4"
+		acceptedPrefixes    = ".1.3.6.1.4.1.9.9.187.1.2.8.1.1.1.4"
+		deniedPrefixes      = ".1.3.6.1.4.1.9.9.187.1.2.8.1.2.1.4"
+		advertisedPrefixes  = ".1.3.6.1.4.1.9.9.187.1.2.8.1.6.1.4"
+		suppressedPrefixes  = ".1.3.6.1.4.1.9.9.187.1.2.8.1.7.1.4"
+		withdrawnPrefixes   = ".1.3.6.1.4.1.9.9.187.1.2.8.1.8.1.4"
+
+		bgpPeerState               = "bgp.peer.state"
+		bgpPeerAdminStatus         = "bgp.peer.admin_status"
+		bgpPeerInUpdates           = "bgp.peer.updates.in"
+		bgpPeerOutUpdates          = "bgp.peer.updates.out"
+		bgpPeerInTotalMessages     = "bgp.peer.messages.in"
+		bgpPeerOutTotalMessages    = "bgp.peer.messages.out"
+		bgpPeerInUpdateElapsedTime = "bgp.peer.update_elapsed.in"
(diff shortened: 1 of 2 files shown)


cmd/scollector/collectors/collectors.go
```diff
@@ -62,6 +62,21 @@ const (
 )
 
 const (
+	bgpPeerState               = "bgp.peer.state"
```
captncraig: you can actually define these inside of the collector function that uses them. I think doing that may help keep the collectors package from having quite so many symbols defined at package scope.

cmd/scollector/collectors/snmp_ciscobgp.go
```diff
@@ -0,0 +1,190 @@
...
+	"time"
+
+	"bosun.org/_third_party/github.com/mjibson/snmp"
+	"bosun.org/cmd/scollector/conf"
+	"bosun.org/metadata"
+	"bosun.org/opentsdb"
+)
+
+const (
+	state               = ".1.3.6.1.4.1.9.9.187.1.2.5.1.3.1.4"
```
captncraig: these could also be declared in `c_snmp_ciscobgp`. They all also seem to share a prefix of `.1.3.6.1.4.1.9.9.187.1.2`. If you declare that separately and then concatenate in your add function, it may make them easier to look at. Or maybe they are easier to have all together for manual testing reasons. your call.

cmd/scollector/collectors/snmp_ciscobgp.go
```diff
@@ -0,0 +1,190 @@
...
+	m := make(map[string]interface{})
+	for rows.Next() {
+		key := ""
+		var a interface{}
+		id, err := rows.Scan(&a)
+		if err != nil {
+			return nil, err
+		}
+		switch t := id.(type) {
+		case []int:
```
captncraig: I think you can declare a as []int instead of interface{} and pass that into Scan. Then you don;t have to do all of the type checking stuff.

cmd/scollector/collectors/snmp_ciscobgp.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+	for _, bA := range oids {
+		if err := add(bA); err != nil {
+			return nil, err
+		}
+	}
+	return md, nil
+}
+
+func snmp_ip_tree(host, community, oid string) (map[string]interface{}, error) {
```
captncraig: should this return a simple map[string][]int if that is what you really expect?

cmd/scollector/collectors/snmp_ciscobgp.go
captncraig: Nevermind me. I'm just confused.

cmd/scollector/collectors/snmp_ciscobgp.go
```diff
@@ -0,0 +1,190 @@
...
+
+func c_snmp_ciscobgp(community, host string) (opentsdb.MultiDataPoint, error) {
+	// Tag: local_as
+	localASesRaw, err := snmp_ip_tree(host, community, localAS)
+	if err != nil {
+		return nil, err
+	}
+	localASes := make(map[string]string, len(localASesRaw))
+	for k, v := range localASesRaw {
+		s := strconv.FormatInt(v.(int64), 10)
```
captncraig: this type assertion needs to be checked.

cmd/scollector/collectors/snmp_ciscobgp.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	// Tag: local_id
+	localIdentifiersRaw, err := snmp_ip_tree(host, community, localIdentifier)
+	if err != nil {
+		return nil, err
+	}
+	localIdentifiers := make(map[string]string, len(localIdentifiersRaw))
+	for k, v := range localIdentifiersRaw {
+		localIdentifiers[k] = snmp_combine_ip_uint8(v.([]uint8))
```
captncraig: check this type assertion.

cmd/scollector/collectors/snmp_ciscobgp.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	// Tag: remote_as
+	remoteASesRaw, err := snmp_ip_tree(host, community, remoteAS)
+	if err != nil {
+		return nil, err
+	}
+	remoteASes := make(map[string]string, len(remoteASesRaw))
+	for k, v := range remoteASesRaw {
+		s := strconv.FormatInt(v.(int64), 10)
```
captncraig: fmt.Sprint(v)

cmd/scollector/collectors/snmp_ciscobgp.go
```diff
@@ -0,0 +1,190 @@
...
+	add := func(bA bgpAdd) error {
+		m, err := snmp_ip_tree(host, community, bA.oid)
+		if err != nil {
+			return err
+		}
+		for k, v := range m {
+			tags := opentsdb.TagSet{
+				"host":      host,
+				"peer":      k,
+				"local_as":  localASes[k],
```
captncraig: If any of these strings are empty, you will have dataset problems. Do we have a good default if they are empty ("na"?)? Should any of these be metadata rather than tags?

cmd/scollector/collectors/snmp_ciscobgp.go
```diff
@@ -0,0 +1,190 @@
...
+	s := make([]string, len(path))
+	for i := range path {
+		s[i] = fmt.Sprint(path[i])
+	}
+	// In some cases there's a .1.1 after the IP (because SNMP is awesome), drop it:
+	s = s[:4]
+	return strings.Join(s, ".")
+}
+
+type bgpAdd struct {
```
captncraig: this struct could even be defined inside of your collector function. I am getting pretty serious about not wanting to add uneccessary things to the collectors package namespace.

captncraig:
Overall looks pretty solid. Just a few minor things to look at, to make sure this won't panic on us.

cmd/scollector/collectors/collectors.go
shanemadden: Will do - was following the conventions from snmp_ifaces which I assumed were an attempt to enable easier localization of these strings?

cmd/scollector/collectors/snmp_ciscobgp.go
shanemadden: Unfortunately, no - the map's values are int64 for some OIDs, []uint8 for others, and uint8 for others.

cmd/scollector/collectors/snmp_ciscobgp.go
shanemadden: Can you clarify? A value from that OID should always be []uint8 - should I switch things around so we won't get a panic on bad data?

cmd/scollector/collectors/snmp_ciscobgp.go
shanemadden: I thought about the metadata route for local_as and local_id, but they could potentially vary for a given host with multiple BGP instances, so I believe they belong as tags.

There shouldn't be a case where any of these values doesn't exist - I think probably the best course is to error out if any are missing, I'll implement that.

cmd/scollector/collectors/snmp_ciscobgp.go
captncraig: I would just do:
```
if uv,ok := v.([]uint8); ok{
   localIdentifiers[k] = snmp_combine_ip_uint8(uv)
}
```
}

captncraig: maybe an `else error` clause too

PR #1419, opened 2015-10-26
Storing error data in redis

captncraig:


gbrayut:
I read through it and nothing jumps out as incorrect, but I'm not very familiar with the redis code or error code really. Probably ready to test on branchbosun or in Denver.

PR #1420, opened 2015-10-26
cmd/bosun: add includeFiles conf setting

mreinhardt:
Set to a glob of the files to include, e.g. `includeFiles = /etc/bosun/conf.d/*.conf`

This allows for separation of global configuration with different confs and alerts for different server types which is helpful for configuration management (via puppet, salt, etc). For example, you might have:

```
/etc/bosun/bosun.conf
/etc/bosun/conf.d/app.conf
/etc/bosun/conf.d/cache.conf
/etc/bosun/conf.d/database.conf
/etc/bosun/conf.d/elasticsearch.conf
# ...etc...
```

### DIFF
--- a/cmd/bosun/conf/conf.go
+++ b/cmd/bosun/conf/conf.go
@@ -10,6 +10,7 @@ import (
 	"net/mail"
 	"net/url"
 	"os"
+	"path/filepath"
 	"regexp"
 	"runtime"
 	"strconv"
@@ -30,6 +31,7 @@ import (
 type Conf struct {
 	Vars
 	Name             string        // Config file name
+	IncludeFiles     string        // Config include files glob string
 	CheckFrequency   time.Duration // Time between alert checks: 5m
 	DefaultRunEvery  int           // Default number of check intervals to run each alert: 1
 	HTTPListen       string        // Web server listen address: :80
@@ -338,6 +340,7 @@ func New(name, text string) (c *Conf, err error) {
 	defer errRecover(&err)
 	c = &Conf{
 		Name:             name,
+		IncludeFiles:     "",
 		CheckFrequency:   time.Minute * 5,
 		DefaultRunEvery:  1,
 		HTTPListen:       ":8070",
@@ -351,13 +354,20 @@ func New(name, text string) (c *Conf, err error) {
 		Templates:        make(map[string]*Template),
 		Alerts:           make(map[string]*Alert),
 		Notifications:    make(map[string]*Notification),
-		RawText:          text,
+		RawText:          "",
 		bodies:           htemplate.New(name).Funcs(htemplate.FuncMap(defaultFuncs)),
 		subjects:         ttemplate.New(name).Funcs(defaultFuncs),
 		Lookups:          make(map[string]*Lookup),
 		Macros:           make(map[string]*Macro),
 	}
-	c.tree, err = parse.Parse(name, text)
+	c.Update(text)
+	return
+}
+
+func (c *Conf) Update(text string) {
+	c.RawText = fmt.Sprint(c.RawText, "\n", text)
+	ctree, err := parse.Parse(c.Name, text)
+	c.tree = ctree
 	if err != nil {
 		c.error(err)
 	}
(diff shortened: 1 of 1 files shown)


captncraig:
I like the idea, but I am a bit worried about the semantics of this.

It looks like included files get appended to the end of the text, but also get parsed inline where they are included. This can cause confusion when things lower in the main config than the include reference included constructs or vice versa. I would almost rather see this implemented as a text preprocessing step so that they appear in the raw text exactly where they are included, and then parsed as a whole.

My bigger concern, however is the rule editor page. All included files will be shown as one big text blob, which will make it a tad tricky to know where in the file tree to commit your changes after you are done editing things. Do you have any ideas for that?

captncraig:
It may also be useful to put a comment in the raw text where you insert the included file.

Perhaps

```
alert mainFileA{
...
}

## INCLUDED FROM foo.conf
alert fooA{
...
}
## END foo.conf
```
or something like that.

mreinhardt:
Thank you for the prompt feedback.

I definitely understand the concern about the inline parsing being different than the way the raw text is concatenated. I can change it to do a regex replacement of the concatenated `*.conf` files where the `includeFiles` is located.

As to the second and bigger concern, I will add the start/end section comments as well. I should be able to make these changes by EOD.

mreinhardt:
Ah, I see that your bigger concern is about how it's handled in the user interface. I had currently only been using that to test alerts via copy/paste of files I was already editing in the our repository. If some sort of separate include file functionality in the rule editor other than the start/end comments is required, I'd like to request that this be punted as an issue so that the functionality itself can be merged and used prior to that (once the above changes are implemented).

kylebrandt:
I share the concern about the config editor. Don't really want to have an includes feature that isn't support in the UI. Basically, the other files would probably need to be expressed as tabs in the editor page.

mreinhardt:
You got it. I retract my EOD estimate then, but will work on it.

IzakMarais:
> If some sort of separate include file functionality in the rule editor other than the start/end comments is required, I'd like to request that this be punted as an issue so that the functionality itself can be merged and used prior to that (once the above changes are implemented).

I agree with this, the separation of global config into different files would be very useful for us. Since it is in any case not currently possible apply changes to the *.conf file from the editor, I don't understand why tabs in the editor would be a strong requirement. It seems more like a nice to have that can be a different feature.

kylebrandt:
Regardless of the implementation, this is adding an extra dimension to the structure of the config. Currently there is just one file. Supporting multiple files means modifying the structure to be two levels deep instead of one.

If Bosun is to actually understand this structure of the underlying config then it should be able to both read in and write out in the same structure via the UI - currently we have this workflow with download (or just copy and paste). We used to have to select out the portions of the config and it was a pain when working with things like Macros. 

So without the UI element done, I'm not confident that any implementation of changing the structure of the config *inside bosun* won't back us into a corner if we were to have the UI reflect this structure. Therefore I don't want to accept this as is. 

For now, if you want to have this ability in the short term you could make a shell script (or a go cmd based on this code) that generates and reads in a conf easily. You could then just gen bosun's config from your repo.

Hope I have explained myself well and haven't discouraged you :-/

mreinhardt:
Sorry, not discouraged, just inundated with other priorities. Someone else may have to pick this up.

kylebrandt:
closing this for now until someone picks it up

IzakMarais:
Was really looking forward to this. Will probably do a go cmd based on this code as @kylebrandt suggested.

IzakMarais:
What we would _really_ like is an extended HTTP REST API to allows alert/template definitions to be PUT. 

Similar to how Grafana allows dashboards to be PUT, this feature will facilitate the use of Bosun by different teams. Teams can store their bosun.conf files in their own, seperate version control repositories. In this way, each monitored applications in the cluster takes responsiblity to manage and update its own bosun.conf rules. When an application is deployed or upgraded it can send its latest bosun.conf settings to the Bosun node without needing file access to the Bosun server or to its salt/chef/puppet configuration.

kylebrandt:
Yes that would be cool. But to remove and add alerts during runtime will
take reloading first which I'm currently working on (appears to work fine
in a brach, but there must be bugs, more testing needed)

The next step after config reloading will be to then turn configuration
into an interface in the code, so it can be implemented by different
providers

PR #1422, opened 2015-10-27
Add google_analytics collector.

alienth:
Config looks like the following:

```
[[GoogleAnalytics]]
  ClientID = "blahblahblah.apps.googleusercontent.com"
  Secret = "foo"
  Token = '''
aHR0cHM6Ly93d3cueW91dHViZS5jb20vd2F0Y2g
'''

  [[GoogleAnalytics.Sites]]
    Name = "stackoverflow.com"
    Profile = "123456a"

  [[GoogleAnalytics.Sites]]
    Name = "stackoverflow-mobile"
    Profile = "2456245"
    Offset = 8

```

The `Token` is a long-lived cached oauth token. The string itself is a base64'd gob. Generating it is done by hand at the moment. I'll investigate writing a util to do this so outszzziders can use it.

`Profile` is a magic GA identifier. `Offset` is what timeslice of the real-time data we want to grab. Mobile real-time data doesn't fully converge until about 5-6 minutes, so we only want to grab the minute slice from 8 minutes ago. If Offset isn't specified or set to 0, it is set to 1 (the minimum offset which makes sense, given the nature of the data).

Note that this has been running on my VM for several months, although the code was broken in a few places. This is the cleaned up version, finally.


cmd/scollector/collectors/google_analytics.go
```diff
@@ -0,0 +1,99 @@
...
+			return md, err
+		}
+		time := time.Now().Add(time.Duration(-1*site.Offset) * time.Minute).Unix()
+		value := "0"
+		// Iterates through the response data and returns the time slice we
+		// actually care about when we find it. If no offset was specified,
+		// the row we care about is minute '0' - or the most recently
+		// gathered datapoint.
+		for _, v := range data.Rows {
+			minute, err := strconv.Atoi(v[0])
```
captncraig: what is the type of v? magic number indexing is a bit scary. Even a comment documenting the expected layout would help.

cmd/scollector/collectors/google_analytics.go
```diff
@@ -0,0 +1,99 @@
...
+			if err != nil {
+				return md, fmt.Errorf("Error parsing GA data: %s", err)
+			}
+			if minute == site.Offset {
+				value = v[1]
+				break
+			}
+		}
+		AddTS(&md, "google.analytics.realtime.pageviews", time, value, opentsdb.TagSet{"site": site.Name}, metadata.Gauge, metadata.Count, "Number of pageviews tracked by GA in one minute")
```
captncraig: Is value a string? I guess we handle that ok, but I'd feel better knowing it is a numeric type.

cmd/scollector/collectors/google_analytics.go
alienth: v is a 'row' containing two numbers. Will add a comment.

cmd/scollector/collectors/google_analytics.go
alienth: Yeah, it's a string representation of a number. I can add a type assertion and convert it.

cmd/scollector/collectors/google_analytics.go
```diff
@@ -0,0 +1,99 @@
...
+	by, err := base64.StdEncoding.DecodeString(tokenstr)
+	if err != nil {
+		return nil, err
+	}
+	b := bytes.Buffer{}
+	b.Write(by)
+	d := gob.NewDecoder(&b)
+	err = d.Decode(&token)
+
+	return config.Client(ctx, token), nil
```
captncraig: I'm not sure why we need the full oauth config with clientId and secret, if they also have to provide a token. Doesn't having a token let you bypass all those flows and just provide an Authorization header?

alienth: This was taken from Google's example code, so I'd assume so. I can try without.

captncraig: You should be able to make a `oauth2.Transport` with a `oauth2.StaticTokenSource`.

alienth: Verified both the clientid and secret are necessary.

alienth:
@captncraig Added fixup commits.

PR #1426, opened 2015-10-28
cmd/bosun: Support seconds when representing "now" on the expression page

kylebrandt:
In particular this is needed for the Grafana bosun backend since it uses this route and specifies the seconds. Using the Grafana Bosun backend on versions prior to this commit will not work.


captncraig:
LGTM

PR #1434, opened 2015-11-02
cmd/scollector: new cadvisor collector

nickethier:
Requesting feedback on some initial work done to add support for collecting metrics from [cadvisor](https://github.com/google/cadvisor). 

Currently this collector will pull all running docker containers through cadvisor and add collect various cpu, memory, fs and network stats, tagging them with the container's ID and name.

### DIFF
--- a/_third_party/github.com/golang/glog/LICENSE
+++ b/_third_party/github.com/golang/glog/LICENSE
@@ -1,0 +1,191 @@
+Apache License
+Version 2.0, January 2004
+http://www.apache.org/licenses/
+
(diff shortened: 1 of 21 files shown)


captncraig:
This looks really nice. Can I assume you have tested it and it is working? The last thing I need is documentation added to https://github.com/bosun-monitor/bosun/blob/master/cmd/scollector/doc.go

nickethier:
@captncraig Yes I've tested locally and with some production machines. 

Added docs, squashed and pushed.

captncraig:
Looks great. Thanks for the contribution!

PR #1435, opened 2015-11-02
cmd/scollector collector: local listener

mathpl:
This adds the local_listener collector, it starts an http server locally and listens for /api/put and /api/metadata/put http requests. It ingest that data and forward it through its usual mechanism, basically acting as a proxy.

### DIFF
--- a/cmd/scollector/collectors/local_listener.go
+++ b/cmd/scollector/collectors/local_listener.go
@@ -1,0 +1,108 @@
+package collectors
+
+import (
+	"compress/gzip"
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+
+	"bosun.org/cmd/scollector/conf"
+	"bosun.org/metadata"
+	"bosun.org/opentsdb"
+)
+
+func init() {
+	registerInit(func(c *conf.Conf) {
+		if c.LocalListener != "" {
+			collectors = append(collectors, &StreamCollector{F: func() <-chan *opentsdb.MultiDataPoint {
+				return c_local_listener(c.LocalListener)
+			},
+				name: fmt.Sprintf("local_listener-%s", c.LocalListener),
+			})
+		}
+	})
+}
+
+func c_local_listener(listenAddr string) <-chan *opentsdb.MultiDataPoint {
+	pm := &putMetric{}
+	pm.localMetrics = make(chan *opentsdb.MultiDataPoint, 1)
+
+	mux := http.NewServeMux()
+	mux.Handle("/api/put", pm)
+	mux.HandleFunc("/api/metadata/put", putMetadata)
+	go http.ListenAndServe(listenAddr, mux)
+
+	return pm.localMetrics
+}
+
+type putMetric struct {
+	localMetrics chan *opentsdb.MultiDataPoint
+}
+
+func (pm *putMetric) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	var (
+		bodyReader io.ReadCloser
+		err        error
+	)
+	defer r.Body.Close()
+
+	if r.Method != "POST" {
+		w.WriteHeader(405)
+		return
+	}
+
+	if r.Header.Get("Content-Encoding") == "gzip" {
+		if bodyReader, err = gzip.NewReader(r.Body); err != nil {
+			w.WriteHeader(500)
+			w.Write([]byte(fmt.Sprintf("Unable to decompress: %s\n", err)))
+			return
+		}
+	} else {
+		bodyReader = r.Body
+	}
+
+	if body, err := ioutil.ReadAll(bodyReader); err != nil {
+		w.WriteHeader(500)
+		return
+	} else {
+		bodyReader.Close()
+
+		var (
+			dp  *opentsdb.DataPoint
+			mdp opentsdb.MultiDataPoint
+		)
+
+		if err := json.Unmarshal(body, &mdp); err == nil {
+		} else if err = json.Unmarshal(body, &dp); err == nil {
+			mdp = opentsdb.MultiDataPoint{dp}
+		} else {
+			w.WriteHeader(500)
+			w.Write([]byte(fmt.Sprintf("Unable to decode OpenTSDB json: %s\n", err)))
+			return
+		}
+
+		for _, dp := range mdp {
+			dp.Tags = AddTags.Copy().Merge(dp.Tags)
+		}
+
+		pm.localMetrics <- &mdp
+
+		w.WriteHeader(204)
+	}
+}
+
+func putMetadata(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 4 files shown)


captncraig:
I'm curious what role this plays for you. Have you seen `cmd/tsdbrelay` in this repo? It is essentially a standalone forwarding proxy for those endpoints. Why would we need this in scollector itself?

mathpl:
It's to simplify deployment of application producing OpenTSDB metrics. The application sends its data to a local scollector (which we need anyway to collect system metric) which adds our usual default tags relating to the host. I see it as a http equivalent to tcollector's udp_bridge.py.

captncraig:
Very interesting. What language is your app in?

mathpl:
Some in Java, some in Go.

captncraig:
In most deployments I have seen, apps will send metrics directly to bosun. In go you can load the collect package directly, or https://bitbucket.org/oneoffcode/go-metrics-bosun/src. I haven't seen a java library, but one would be really useful. Not saying we won't accept this, but making sure to look at alternative approaches.

mathpl:
The benefit of this imo is that the application doesn't need to know anything about its deployment. No need to find out what's its hostname, cluster, product, etc. It's all added by scollector which is in charge of topology related tags.

As for producing libraries looks like we had the same idea on the Go side (https://github.com/mathpl/go-tsdmetrics). For Java our library is tightly coupled with our internal framework currently so we can't open source it easily for now.

cmd/scollector/collectors/stream.go
```diff
@@ -0,0 +1,82 @@
...
+	"runtime"
+	"sync"
+	"time"
+
+	"bosun.org/collect"
+	"bosun.org/metadata"
+	"bosun.org/opentsdb"
+)
+
+type StreamCollector struct {
```
captncraig: I'm not sure the stream collector is really needed here. Why not just have the listener relay data points to the collect package directly. That channel is made here: https://github.com/bosun-monitor/bosun/blob/master/cmd/scollector/main.go#L191. I would suggest putting the http listener in the main package and relaying directly to collect.

cmd/scollector/collectors/local_listener.go
```diff
@@ -0,0 +1,106 @@
...
+	return nil
+}
+
+func c_local_listener(listenAddr string) <-chan *opentsdb.MultiDataPoint {
+	pm := &putMetric{}
+	pm.localMetrics = make(chan *opentsdb.MultiDataPoint, 1)
+
+	mux := http.NewServeMux()
+	mux.Handle("/api/put", pm)
+	mux.Handle("/api/metadata/put", util.JSON(PutMetadata))
```
captncraig: I don't think you really need the json middleware for this. You already do error handling, and there is no response to serialize. Probably should revert moving of that func.

cmd/scollector/collectors/stream.go
mathpl: The main reason for it is to be able to generate scollector.collector.count for this collector.

captncraig: You could just do `collect.Add("localListener.relayed", tags, len(mdp))` or similar in your relay function as you see datapoints.

mathpl: I felt if any other future collectors would follow the same pattern having a StreamCollector in addition of the IntervalCollector would be beneficial. I'll gladly remove it and attach the collector in main.go if you wish though.

cmd/scollector/collectors/stream.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+}
+
+func (s *StreamCollector) Run(dpchan chan<- *opentsdb.DataPoint, quit <-chan struct{}) {
+	if s.Enable != nil {
+		go func() {
+			for {
+				next := time.After(time.Minute * 5)
+				s.Lock()
+				s.enabled = s.Enable()
```
captncraig: is Enabled really a useful concept here? The interval collector uses it to determine "Is X installed on this system" so it can avoid issuing unecessary requests and such. If you real collector is running external to the Stream Collector, all you are doing with enabled is dropping datapoints. I would rather just always be enabled on these, and make the user "opt in" by enabling it in config.

cmd/scollector/collectors/stream.go
captncraig: I suppose at least add a good comment here explaining the purpose of StreamCollector, and when I would use it.

cmd/scollector/collectors/stream.go
mathpl: Agreed, I've removed it.

captncraig:
Can you fix conflicts?

cmd/scollector/collectors/stream.go
```diff
@@ -0,0 +1,61 @@
...
+	"bosun.org/opentsdb"
+)
+
+/* StreamCollector is useful for collectors that do not produces metrics at a
+   preset interval. Instead it consummes directly from a channel provided by
+   the collector and forwards it internally. */
+
+type StreamCollector struct {
+	F      func() <-chan *opentsdb.MultiDataPoint
+	Enable func() bool
```
captncraig: remove Enable from the struct.

cmd/scollector/collectors/local_listener.go
```diff
@@ -0,0 +1,106 @@
...
+		if bodyReader, err = gzip.NewReader(r.Body); err != nil {
+			w.WriteHeader(500)
+			w.Write([]byte(fmt.Sprintf("Unable to decompress: %s\n", err)))
+			return
+		}
+	} else {
+		bodyReader = r.Body
+	}
+
+	if body, err := ioutil.ReadAll(bodyReader); err != nil {
```
captncraig: you don't need to ReadAll, just do `json.NewDecoder(bodyReader)`, and decode from that.

cmd/scollector/collectors/local_listener.go
```diff
@@ -0,0 +1,106 @@
...
+
+		pm.localMetrics <- &mdp
+
+		w.WriteHeader(204)
+	}
+}
+
+func putMetadata(w http.ResponseWriter, r *http.Request) {
+	d := json.NewDecoder(r.Body)
+	var ms []metadata.Metasend
```
captncraig: `defer r.Body.Close()`

cmd/scollector/collectors/local_listener.go
mathpl: We need it because there's 2 ways the body can be decoded, as single data point or an array of data points. Both format are accepted by OpenTSDB. This saves us from decompressing the body 2 times.

mathpl:
Been a while, anything else I can do?

captncraig:
rebase on master and resolve conflicts please.

mathpl:
Done. Not quite sure what's wrong with TravisCI though.

mathpl:
rebased again to get the build fixed.

PR #1436, opened 2015-11-02
c_google_analytics: Gather realtime metrics for browsers, trafficType, deviceCategory, and OS.

alienth:
If the 'Detailed' attribute is set to true for a GA site in the configuration, this will track additional realtime metrics.

The `getPageviews` function is unchanged from the original. Was simply cut out of `c_google_analytics`.

The new metrics are named as follows:

`google.analytics.realtime.activeusers.by_${DIMENSION}`


captncraig:
LGTM

PR #1437, opened 2015-11-02
cmd/scollector: configurable network interface name regexp

mathpl:
This adds config to replace the default interface name regexp used in ifstat_linux.go. We use custom interface names internally hence this PR. I think this should also fill the need expressed in  https://github.com/bosun-monitor/bosun/pull/1366.

### DIFF
--- a/cmd/scollector/collectors/ifstat_linux.go
+++ b/cmd/scollector/collectors/ifstat_linux.go
@@ -2,22 +2,30 @@ package collectors
 
 import (
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"path/filepath"
 	"regexp"
 	"strings"
 	"time"
 
+	"bosun.org/cmd/scollector/conf"
 	"bosun.org/metadata"
 	"bosun.org/opentsdb"
 	"bosun.org/util"
 )
 
 func init() {
-	collectors = append(collectors, &IntervalCollector{F: c_ifstat_linux})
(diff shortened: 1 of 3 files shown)


cmd/scollector/main.go
```diff
@@ -99,6 +99,7 @@ func main() {
 		}
 	}
 	collectors.Init(conf)
+	check(collectors.AddIfstatConfig(conf.IfaceExpr))
```
captncraig: We are going away from having main initialize collectors. Instead conllectors can "self-register" init functions. See https://github.com/bosun-monitor/bosun/blob/master/cmd/scollector/collectors/haproxy_unix.go#L16

cmd/scollector/collectors/ifstat.go
```diff
@@ -0,0 +1,13 @@
+package collectors
+
+import "regexp"
+
+var ifstatRE = regexp.MustCompile(`\s+(eth\d+|em\d+_\d+/\d+|em\d+_\d+|em\d+|` +
```
captncraig: If you do the registerInit thing, this file can be moved back into ifstat_linux.go

captncraig:
Looks good other than the init thing. Please squash to a single commit when you do that and I can merge.

mathpl:
The new registerInit() is much better indeed! Fixed and squashed.

captncraig:
Vlosed via b24f598891d47a47234fd255732badbbf1efa5a8

PR #1439, opened 2015-11-03
Add more info to Clean() err message.

alienth:
The usage of 'key' was erroneous. Corrected it and added the tag which the value was from.

Also added the metric to the error message when a tag failed to clean.


kylebrandt:
LGTM assuming it passes build tests

alienth:
Build passed, but it will appear to have not as I'm about to rebase. Merging shortly.

PR #1441, opened 2015-11-03
Sort data to prevent arbitrary metadata changes.

alienth:
The `macs` and `ips` data from snmp don't come in ordered. As a result, metadata gets rewritten whenever their order is changed. This should prevent that.

:eyeglasses:  @gbrayut @captncraig


captncraig:
LGTM

PR #1445, opened 2015-11-05
puppet_linux: Track puppet.run.status and fix broken error checking logic.

alienth:
We have a couple hosts successfully 'running' puppet and updating the last_run timestamp, but the run has a status of 'failed'. Adding a metric to track that.


captncraig:
LGTM

PR #1447, opened 2015-11-05
Adding logging if empty templates are generated.

captncraig:


gbrayut:
Looks good but failing test now:

--- FAIL: TestCheckFlapping (0.01s)
panic: reflect: indirection through nil pointer to embedded struct [recovered]
	panic: reflect: indirection through nil pointer to embedded struct [recovered]
	panic: reflect: indirection through nil pointer to embedded struct

PR #1448, opened 2015-11-06
WIP: new udp listener for sporadic counter events.

captncraig:
This is the base implementation of a udp collector. I put it in the collect package because it is not in a `cmd/*` package and it seemed logical. 

We have a couple of options for deploying it:

1) scollector listens locally, stores things in redis master. Less udp traffic on network.
2) bosun listens. 
3) tsdbrelay listens. I don't like this a ton.

My vote is for 2. Any or all of them can be implemented fairly easily. 

@gbrayut @kylebrandt @bretcope

### DIFF
--- a/cmd/scollector/collectors/redis_counters.go
+++ b/cmd/scollector/collectors/redis_counters.go
@@ -1,0 +1,81 @@
+package collectors
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
+	"bosun.org/_third_party/github.com/garyburd/redigo/redis"
+	"bosun.org/cmd/scollector/conf"
+	"bosun.org/collect"
+	"bosun.org/metadata"
+	"bosun.org/opentsdb"
+	"bosun.org/slog"
+)
+
+func init() {
+	registerInit(func(c *conf.Conf) {
+		for _, red := range c.RedisCounters {
+			collectors = append(collectors,
+				&IntervalCollector{
+					name: "redisCounters_" + red.Server,
+					F: func() (opentsdb.MultiDataPoint, error) {
+						return c_redis_counters(red.Server, red.Database)
+					},
+				})
+		}
+	})
+}
+
+func c_redis_counters(server string, db int) (opentsdb.MultiDataPoint, error) {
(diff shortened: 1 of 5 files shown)


kylebrandt:
What about putting it in its own process? My concern is stability:
 - This is new
 - If increments start being rapidly sent, it could impact stability. I would rather that impact was mostly localized. If we make sure the keyspace doesn't conflict that it could use its own redis, or share bosun's depending on configuration

Also, we can always merge it into one of the other cmds at a later date.

bretcope:
It seems unlikely that increments will be sent particularly rapidly considering the intended use case for this is very slow moving counters.

kylebrandt:
@bretcope Sure, but if there are no protection mechanisms (such as rate limiting) than it could be used wrong. That could because the person didn't get the intention, or if an unforeseen condition arises that starts to increment the counter more rapidly than expected.

As a consumer in our environment you wouldn't even need to know that it is a different process since we have it behind haproxy anyways.

bretcope:
The same could be said of using Bosun/OpenTSDB in general though. If you send it way too much data, or too often, bad things could happen. It seems to me that this is less likely to impact performance because it's a simple listener which increments a Redis counter. That should be able to handle a tremendous load quite easily.

captncraig:
We can monitor the incoming packet rate and alert on it if it explodes. I'm
not worried about that. "That should be able to handle a tremendous load
quite easily." should be pretty accurate.

captncraig:
To reduce deployment complexity I would like to put it in another cmd.
Although if we expand the set of operations significantly over time I could
see separating it.

kylebrandt:
You're call - if we think the probability of a load impact the stops bosun from alerting is low then putting in bosun seems reasonable

gbrayut:
+1 on not being in bosun. Bosun doesn't have high availability, so we would loose counts during builds (which is frequent). A separate cmd or tsdbrelay (the only current cmd with high availability) is my vote.

captncraig:
Thats an ok point.

kylebrandt:
+1 on Greg's point

PR #1455, opened 2015-11-10
processes_linux: Add a tiebreaker if the process creation time is identical.

alienth:
Had an issue with keepalived where the 3 processes had an identical mtime. Resulted in the metric IDs of said processes hopping around, and then counters getting jacked. This should remedy that.

:eyeglasses: @kylebrandt  @captncraig


alienth:
Confirmed this fixes the problem in testing, as seen here: 

image

captncraig:
lgtm

PR #1461, opened 2015-11-13
cmd/scollector: dns_windows.go: New metrics & better descriptions for metrics

mhenderson-so:
- Added better descriptions for the metrics. Rather than using the WMI provided description, I have included the descriptions as best as possible from the original Microsoft documentation, which has come from the class information here : https://msdn.microsoft.com/en-us/library/cc448843.aspx

- Added some additional metrics that got missed the first time around due to a faulty filtering rule that I used when gathering the metrics.


captncraig:
closed via ae88267e63b66184988ef5dbe4efd40154bce5a3

PR #1462, opened 2015-11-15
cmd/scollector: fix pseudo file system checking for dfstat on linux

noblehng:
Use string equality testing instead of sub-string matching in the
readLine() function for /proc/filesystems in isPsenduFS(), because the
latter could produce false-positive result.

For a real example:
nodev	selinuxfs
	xfs
these two lines will make the old code think xfs is a pseudo file
system.

### DIFF
--- a/cmd/scollector/collectors/disk_linux.go
+++ b/cmd/scollector/collectors/disk_linux.go
@@ -75,9 +75,9 @@ func removable_fs(name string) bool {
 
 func isPseudoFS(name string) (res bool) {
 	err := readLine("/proc/filesystems", func(s string) error {
-		if strings.Contains(s, name) && strings.Contains(s, "nodev") {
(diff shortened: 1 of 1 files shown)


captncraig:
I'm a bit confused how the old code could cause xfs to be a false-positive. Isn't it on a different line?

noblehng:
It scan every lines in /proc/systems for every file system types output of df command.
If there is one line that contains both the passed in file system type and "nodev", res will be set to true.

So for xfs, "nodev selinuxfs" contains both "xfs" and "nodev", then res will be true, hence
isPseudoFS("xfs") is true in the old code.

captncraig:
Gotcha. Thanks for the catch!

dimamedvedev:
Thank you for fixing my bug.

PR #1466, opened 2015-11-16
Really kill a process when we say we kill it.

nicollet:
BTW, looking at the previous code, it seems odd to randomly kill() or int(). It would be better to signal first, and terminate after a moment (say 5 * time.Millisecond). I would also use SIGTERM signal instead of SIGINT, since SIGINT means: "interrupted from the keyboard".


captncraig:
Right, the kill timeout should be 2 * timeout. I feel bad for breaking this. Could you make that change as well please?
You could handle TERM and INT if you like, but we do use INT while running locally.

captncraig:
Fixed in #1827

PR #1469, opened 2015-11-17
cmd/bosun: Support OpenTSDB 2.2 filters

kylebrandt:
image


cmd/bosun/conf/conf.go
```diff
@@ -82,7 +83,12 @@ func (c *Conf) TSDBContext() opentsdb.Context {
 	if c.TSDBHost == "" {
 		return nil
 	}
-	return opentsdb.NewLimitContext(c.TSDBHost, c.ResponseLimit)
+	version := c.TSDBVersion
```
captncraig: set the default at conf init time, not every time we make a ocntext

gbrayut:
Looks like 2.2 also supports Fill Policies. There are a lot of times using Zero instead of linear interpolation would be better, curious if that is something we could also support in bosun.

http://opentsdb.net/docs/build/html/user_guide/query/aggregators.html#fill-policies

cmd/bosun/web/static/partials/graph.html
```diff
@@ -83,54 +83,114 @@
...
+							<div class="col-sm-3">
+								<input type="text" class="form-control" ng-model="t.nGbFilters[k].filter" ng-options="tv for tv in tagvs[$parent.$index][k]" bs-typeahead min-length="0" limit="100" ng-disabled="!t.metric_tags[k]">
+							</div>
 						</div>
-					</div>
-				</form>
+					</form>
+				</div>
+
+				<div ng-if="! filterSupport">
```
captncraig: I like this duplication now, just because we can delete it when we depricate 2.1. Maybe a comment somewhere reminding us to make changes to both blocks though.

kylebrandt:
This got pushed, just had the wrong git username set so I don't think GH detected it.

PR #1470, opened 2015-11-17
Fix to last logic for template rendering.

captncraig:
Old abnormal events were getting confused with current ones.

Hopefully this will help fix the empty emails.


captncraig:
Superceded by #1473

Issue #1453, opened 2015-11-08
empty e-mails incoming


timurb:
Today I've had an OpenTSDB outage and Bosun as expected emailed me a dozen of "Unknown" alerts.

What's wrong with that is soon I've started to messages with no subject and empty body while all of templates I defined in config file do have both template and body:
```
Date: Sun, 8 Nov 2015 07:40:27 +0000
Mime-Version: 1.0
X-Bosun-Server: tsdb
Content-Type: multipart/mixed;
 boundary=435514121d807855059eaa73cdea71ccbaa92a293f0eaafb0b48d9c5ec2f
To: ops@mydomain.com
From: ops@mydomain.com
Message-ID: <00000150e60a025a-e6d749cc-1edb-4844-93a6-9307544980d4-000000@us-west-2.amazonses.com>
X-SES-Outgoing: 2015.11.08-54.240.27.24
Feedback-ID: 1.us-west-2.LQSZzAmtlzVGV4VEOTpAGIGbe5mPLH9mf6V5OR2giLE=:AmazonSES
X-Original-Sender: ops@mydomain.com
X-Original-Authentication-Results: mx.google.com;       spf=pass (google.com:
 domain of 00000150e60a025a-e6d749cc-1edb-4844-93a6-9307544980d4-000000@us-west-2.amazonses.com
 designates 54.240.27.24 as permitted sender) smtp.mailfrom=00000150e60a025a-e6d749cc-1edb-4844-93a6-9307544980d4-000000@us-west-2.amazonses.com;
       dkim=pass header.i=@amazonses.com
Precedence: list
Mailing-list: list ops@mydomain.com; contact ops+owners@mydomain.com
List-ID: <ops.mydomain.com>
X-Spam-Checked-In-Group: ops@mydomain.com
X-Google-Group-Id: 612457435205
List-Help: <http://support.google.com/a/mydomain.com/bin/topic.py&topic=25838>, <mailto:ops+help@mydomain.com>

--435514121d807855059eaa73cdea71ccbaa92a293f0eaafb0b48d9c5ec2f

--435514121d807855059eaa73cdea71ccbaa92a293f0eaafb0b48d9c5ec2f--
```

I'm running Bosun 0.4.0 on Ubuntu 12.04, gathering incoming data via `/api/index` (sending to OpenTSDB in parallel) and sending out e-mails via AWS SES.

timurb:
Copying here yesterday's slack chat for reference:
```
[...]

PR #1473, opened 2015-11-18
Stopping flapping notifications and empty emails

### LINKED ISSUE #1471: Hardening State Transition Rules
Lets clarify and fix what should happen when states change:

States in increasing order of severity are:
1. Normal
2. Warning
3. Critical
4. Unknown

## Notifications
You should get a notification any time:

1. A new incident is opened. This is the first non-normal incident on a closed alert.
2. The status increases from the maximum we have seen this incident. 

for the following state transitions:

1. Normal
2. **Warning**
3. **Critical**
4. Warning
5. Critical

I would expect notifications only on the bold states. This is not currently the behaviour, because of a bug. We are currently also notifying on 5, which should not be.

## Dashboard

The dashboard is fairly complicated at this point. We show a variety of things:

- Current status for the alert key is indicated by the small indented bar on the edge of the panel header.
- Last Abnormal status and template are represented by the color of the panel header and the text of the subject. We want this to show the info from the most recent non-normal check we have.
- Worst status this incident is not shown, but should probably be made availible in the expanded info panel for an alert key.

I broke this particular functionality in an attempt to only show the template from the maximum status. It has caused problems and should be fixed. It causes the empty emails among other things.

captncraig:
Fixes #1471

### DIFF
--- a/cmd/bosun/sched/check.go
+++ b/cmd/bosun/sched/check.go
@@ -147,6 +147,7 @@ func (s *Schedule) runHistory(r *RunHistory, ak models.AlertKey, event *Event, s
 	}
 	// assign incident id to new event if applicable
 	prev := state.Last()
+	worst := StNormal
 	event.Time = r.Start
 	if prev.IncidentId != 0 {
 		// If last event has incident id and is not closed, we continue it.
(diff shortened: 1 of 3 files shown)


captncraig:
@gbrayut @kylebrandt

gbrayut:
LGTM but I'm not familiar with the check code. Probably should run in Denver for a bit to make sure it works as expected.

captncraig:
fixes #1453

captncraig:
Honestly, I don't think denver will tell us much interesting since most of this is around notifications, which denver doesn't send.

gbrayut:
It will tell us if there are major flaws in the check logic, but I'm fine either way

PR #1483, opened 2015-11-25
Logging on post notifications.

### LINKED ISSUE #1482: Need log entries for post notification attempts
When we send an email there is a [log entry](https://github.com/bosun-monitor/bosun/blob/266a17a6426b69562c079e8dc20f0a07d011914c/cmd/bosun/conf/notify.go#L104)

```
2015/11/25 03:14:01 info: notify.go:104: relayed alert services.down{host=ny-asdf} to [asdf@examlple.com] sucessfully. Subject: 47 bytes. Body: 14158 bytes.
```

these are helpful to see if Bosun tried to send the alert or not. We do NOT currently have any log entries for [get or post notifications](https://github.com/bosun-monitor/bosun/blob/266a17a6426b69562c079e8dc20f0a07d011914c/cmd/bosun/conf/notify.go#L47) unless there is an error.

which makes tracking down missing alerts much more difficult.

captncraig:
Fixes #1482

### DIFF
--- a/cmd/bosun/conf/notify.go
+++ b/cmd/bosun/conf/notify.go
@@ -30,10 +30,10 @@ func (n *Notification) Notify(subject, body string, emailsubject, emailbody []by
 		go n.DoEmail(emailsubject, emailbody, c, ak, attachments...)
 	}
 	if n.Post != nil {
-		go n.DoPost([]byte(subject))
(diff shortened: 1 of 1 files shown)


gbrayut:
Does this work for get notifications too? looked like that was a separate method.

PR #1485, opened 2015-11-26
c_elasticsearch: Refactor collector.

alienth:
This change refactors much of the collector, and also makes it compatible with ElasticSearch 2.

Notable changes:
* All of the JSON we use has been struct-ified.
* Metric metadata and descriptions have been placed into the struct field tags.
* Metrics which are version-specific have been denoted in field tags, and are conditionally sent based on those tags and what version of Elastic we're hitting.
* Namespaces of the metrics have been rearranged to make more intuitive sense.
    * Previously local-node index stats were stuck right on `elastic.`, while cluster-white index stats were stuck in `elastic.indices.`. These have been moved to `elastic.indices.local` and `elastic.indices.cluster`, respectively.
    * Health-related stats have been moved into `elastic.health`.
* Previously we had a mix of millisecond and second metric values. All metric values are now consistently in seconds.

:eyeglasses: @captncraig @gbrayut

### DIFF
--- a/cmd/scollector/collectors/elasticsearch.go
+++ b/cmd/scollector/collectors/elasticsearch.go
@@ -2,32 +2,38 @@ package collectors
 
 import (
 	"encoding/json"
-	"errors"
-	"math"
 	"net/http"
 	"net/url"
+	"reflect"
 	"regexp"
-	"time"
+	"strings"
 
+	"bosun.org/cmd/scollector/conf"
 	"bosun.org/metadata"
 	"bosun.org/opentsdb"
+	"bosun.org/slog"
 )
 
 func init() {
-	collectors = append(collectors, &IntervalCollector{F: c_elasticsearch, Enable: enableURL(esURL)})
-	collectors = append(collectors, &IntervalCollector{F: c_elasticsearch_indices, Interval: time.Minute * 2, Enable: enableURL(esURL)})
+	registerInit(func(c *conf.Conf) {
+		for _, filter := range c.ElasticIndexFilters {
+			err := AddElasticIndexFilter(filter)
+			if err != nil {
+				slog.Errorf("Error processing ElasticIndexFilter: %s", err)
+			}
+		}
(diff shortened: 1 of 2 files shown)


alienth:
@captncraig I've split the changes up into separate commits to aid in the review. Note that this was done retroactively, so I can't guarantee all of the commit states work atomically. The end result is the same, and I plan on remashing when this is ready for merge.

cmd/scollector/collectors/elasticsearch.go
```diff
@@ -35,189 +41,228 @@ func AddElasticIndexFilter(s string) error {
 	if err != nil {
 		return err
 	}
-	esIndexFilters = append(esIndexFilters, re)
+	elasticIndexFilters = append(elasticIndexFilters, re)
 	return nil
 }
 
+type structProcessor struct {
+	elasticVersion string
```
captncraig: is this too specific to elastic to generalize? It seems like struct tags to metrics is a really cool feature if we could make it generally available.

cmd/scollector/collectors/elasticsearch.go
```diff
@@ -35,189 +41,228 @@ func AddElasticIndexFilter(s string) error {
...
 	}
-	for nodeid, nstats := range stats.Nodes {
-		isMaster := nodeid == clusterState.MasterNode
+	var md opentsdb.MultiDataPoint
+	s := structProcessor{elasticVersion: status.Version.Number, md: &md}
+	ts := opentsdb.TagSet{"cluster": clusterStats.ClusterName}
+	for nodeID, nodeStats := range clusterStats.Nodes {
+		isMaster := nodeID == clusterState.MasterNode
 		if isMaster {
-			cstats := make(map[string]interface{})
```
captncraig: hooray for deleted code.

cmd/scollector/collectors/elasticsearch.go
alienth: In this case, yes. The elastic version processing is what first pushed me down this road.

Would definitely like to consider a generalized version for certain other collectors where it makes sense. If I build that I can maybe have the elastic collector use that while tacking on the version logic.

PR #1487, opened 2015-12-01
cmd/bosun: Have auto filter on graph page

kylebrandt:
Auto mimicks the old behavior of bosun with opentsdb 2.1. The behavior is based on OpenTSDB 2.2 "filter conversions"


cmd/bosun/web/static/js/bosun.js
```diff
@@ -1636,6 +1636,9 @@ bosunControllers.controller('ExprCtrl', ['$scope', '$http', '$location', '$route
             }
         };
     }]);
+/// <reference path="angular.d.ts"/>
```
captncraig: did you upgrade typescript? I'm ok with it, but are there docs that need updating?

captncraig:
lgtm

PR #1488, opened 2015-12-01
cmd/scollector: Collect system uptime via snmp

kylebrandt:


cmd/scollector/collectors/snmp_sys.go
```diff
@@ -0,0 +1,34 @@
...
+	collectors = append(collectors, &IntervalCollector{
+		F: func() (opentsdb.MultiDataPoint, error) {
+			return c_snmp_sys(cfg.Community, cfg.Host)
+		},
+		Interval: time.Minute * 1,
+		name:     fmt.Sprintf("snmp-sys-%s", cfg.Host),
+	})
+}
+
+func c_snmp_sys(community, host string) (opentsdb.MultiDataPoint, error) {
```
captncraig: You could probably do this by building a hard-coded MIB struct like the cisco one does. I'm ok with this though. Do what feels good. LGTM

PR #1489, opened 2015-12-02
cmd/scollector: RabbitMQ skip auto-generated queues

dimamedvedev:
RabbitMQ generates a name for a queue if it's not specified. Total numbers are already available in `rabbitmq.overview`.


gbrayut:
Merged with commit 5edb44ce8824fdfbf6906e0afa25819e9fb7ee7e

PR #1494, opened 2015-12-03
cmd/scollector: Covert gauges to counters for os.cpu

kylebrandt:


cmd/scollector/collectors/collectors.go
```diff
@@ -325,3 +325,17 @@ func skipMetric(index string) bool {
 	}
 	return true
 }
+
+type tsIntegrator func(int64, float64) float64
+
+func getTsIntegrator() tsIntegrator {
+	var i float64
```
captncraig: these variable names are not descriptive enough to indicate what is going on here.

kylebrandt: Suggestions?

captncraig: `i,ts,t,v` -> `total,lastTimestamp, time, value`

kylebrandt: oldTs and ts? i to total?

kylebrandt: Sounds good, thanks

captncraig: How about
 ```
type tsIntegrator struct{
   data map[string]struct{
      total floats64
      lastTime
   }
   sync.Mutex
}
```
?

then you can do `cpuIntegrator.Add(key, now, val)` and `cpuIntegrator.Total(key)`. Hide all of the mutexing and map stuff inside the integrator.

Comment (kylebrandt) [inline cmd/scollector/collectors/collectors.go]: I was thinking there might be use cases where you don't want to use it in
the map, so this would make it more reusable

File: cmd/scollector/collectors/snmp_cisco.go [hunk 2]
```diff
@@ -41,11 +35,59 @@ func SNMPCisco(cfg conf.SNMP) {
...
+			F: func() (opentsdb.MultiDataPoint, error) {
+				return GenericSnmp(cfg, mib)
+			},
+			Interval: time.Second * 30,
+			name:     fmt.Sprintf("snmp-cisco-%s", cfg.Host),
 		},
-		Interval: time.Second * 30,
-		name:     fmt.Sprintf("snmp-cisco-%s", cfg.Host),
-	})
+		&IntervalCollector{
```
captncraig: if you make the integrator here and pass it into c_cisco_cpu, you may not need the map or mutex at all.

PR #1496, opened 2015-12-03
Add oracle database and varnish collector

noblehng:
Oracle support is via sqlplus, so doesn't use cgo.

### DIFF
--- a/cmd/scollector/doc.go
+++ b/cmd/scollector/doc.go
@@ -298,6 +298,22 @@ will deletes it.
 	    docker_name = ''
 	    source = 'kubelet'
 
+Oracles (array of table, keys are ClusterName, Instances): Oracle database
+instances to poll. The Instances key is an array of table with keys
+ConnectionString and Role, which are the same as using sqlplus.
+
+	[[Oracles]]
+	  ClusterName = "oracle rac name"
+	  [[Oracles.instances]]
+	    ConnectionString = "/"
+	    Role = "sysdba"
+	  [[Oracles.instances]]
+	    ConnectionString = "username/password@oraclehost/sid"
+	  [[Oracles.instances]]
+	    ConnectionString = "/@localnodevip/sid"
+	    Role = "sysdba"
+
+
 Windows
 
 scollector has full Windows support. It can be run standalone, or installed as a
--- a/cmd/scollector/collectors/oracle.go
+++ b/cmd/scollector/collectors/oracle.go
@@ -49,7 +49,7 @@ func c_oracle(name string, inst conf.OracleInstance) (opentsdb.MultiDataPoint, e
 		},
 	}
 
-	args := []string{"-S", inst.ConnectString}
+	args := []string{"-S", inst.ConnectionString}
 	if role := inst.Role; role != "" {
 		args = append(args, "as")
 		args = append(args, role)
--- a/cmd/scollector/collectors/oracle.go
+++ b/cmd/scollector/collectors/oracle.go
@@ -1,0 +1,388 @@
+package collectors
+
+import (
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"strconv"
+	"strings"
+	"time"
+
+	"bosun.org/cmd/scollector/conf"
+	"bosun.org/metadata"
+	"bosun.org/opentsdb"
+	"bosun.org/slog"
+	"bosun.org/util"
+)
+
+func init() {
+	os.Setenv("NLS_LANG", "AMERICAN_AMERICA.AL32UTF8")
+
+	registerInit(func(c *conf.Conf) {
+		for _, o := range c.Oracles {
+			name := o.ClusterName
+			for _, inst := range o.Instances {
+				i := inst
+				collectors = append(collectors, &IntervalCollector{
+					F: func() (opentsdb.MultiDataPoint, error) {
+						return c_oracle(name, i)
+					},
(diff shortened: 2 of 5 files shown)


gbrayut:
Thanks for the pull request! A few notes:

- This is out of date and should be rebased on current master. Can just do that as another commit and then squash everything down once we are ready to merge

- I grabbed [Varnish 4.0.3](https://www.varnish-cache.org/releases) to see what the stats look like for the default /etc/varnish/default.vcl config and it looks like there are a few things that should be changed:

  - metric names are usually all lower case. I see varnish.LCK.backend... varnish.MAIN.pools... varnish.VBE.default127.0.0.18080.bereq_bodybytes... which should be run thru strings.ToLower() like you did in [Oracle.go](https://github.com/bosun-monitor/bosun/pull/1496/files#diff-db29c7cbeb6adcd73c49bff454dd3990R386)

  - The metric names and tag keys could use some work. Seems like type always matches the varnish.xxx. part of the metric, so that tag should be removed (or removed from the metric name if it makes sense to SUM or AVG across different types). And we usually put IP/Server information in as tags instead of part of the metric (since you often want to aggregate across those). I'm not very familiar with varnish but you probably want to reorganize them like:

```
current: metric.name{tagk=tagv,tagk=tagv}
varnish.VBE.default127.0.0.18080.bereq_hdrbytes{host=ny-gbraylx01,subtype=default127.0.0.18080,type=VBE}: 0

recommended: metric.name{tagk=tagv,tagk=tagv}
varnish.vbe.be_req_hdr_bytes{host=ny-gbraylx01,backend=default,endpoint=127.0.0.1_8080}: 0
```

  note the host=tag is reserved for which system is sending the metric (the varnish server in this case), so for the backend you'll need to use server or endpoint. Or if the backend tag is descriptive enough and 1to1 mapping with endpoints you can omit the endpoint detail.

I didn't get a chance to play with the oracle collector yet, but let me know if you have any questions.

gbrayut:
Some notes (for future me) on how to test the Oracle collector:


Docker image for a test target: https://hub.docker.com/r/wnameless/oracle-xe-11g/
On client need sqlplus, which can be downloaded [here](http://www.oracle.com/technetwork/database/features/instant-client/index-097480.html ) (need basic and SQL*Plus packages)

Add folder to path: (PowerShell example)
`$env:path += ";C:\Installs\Oracle\instantclient_12_1"`

Then login with default user:
`sqlplus.exe system/oracle@ny-gbraylx01:49161`

This worked, but in my testing generated an error which I think caused problems with the parsing:

> ERROR:
> ORA-28002: the password will expire within 7 days

So I ran the following to change the password:
`alter user system identified by newpassword replace oracle;`

Built scollector with the oracle collector and ran it with print and filter options:
`.\scollector.exe -p -f oracle`

and the following in scollector.toml (Note ConnectString not ConnectionString as per docs)

```
[[Oracles]]
  ClusterName = "Docker Testing"
  [[Oracles.Instances]]
    ConnectString = "system/newpassword@ny-gbraylx01:49161"
```

And I see metrics like:
```
oracle.database_cpu_time_ratio_15s{host=gbray03,oracle_cluster=DockerTesting,oracle_database=XE,oracle_instance=XE}: 0
oracle.execute_count{host=gbray03,oracle_cluster=DockerTesting,oracle_database=XE,oracle_instance=XE}: 2.21429
```

which look great! You can drop the oracle_ prefix on the tag names if you want, but otherwise the oracle collector can be merged when the code comments are addressed.

cmd/scollector/doc.go
```diff
@@ -278,6 +278,22 @@ metrics.
...
 	LocalListener = "localhost:4242"
 
+Oracles (array of table, keys are ClusterName, Instances): Oracle database
+instances to poll. The Instances key is an array of table with keys
+ConnectionString and Role, which are the same as using sqlplus.
+
+	[[Oracles]]
+	  ClusterName = "oracle rac name"
+	  [[Oracles.instances]]
+	    ConnectionString = "/"
```
gbrayut: these should be ConnectString not ConnectionString, or the struct should be changed

noblehng:
Thank you for your review.

I have rebased the branch on current master and modify the code according to above comments.

noblehng:
@gbrayut

Do you have the time to continue the review? I have updated the branch against the latest master.

gbrayut:
Thanks! I'm out of town this weekend but will take another look on Monday or Tuesday.

gbrayut:
LGTM... I'll merge this shortly.

Thanks!

Not sure why but my Windows system wasn't able to connect to the docker oracle instance (likely a firewall/network issue on my side) so here are the steps I used to install sqlplus on my Linux system for testing (in case I need to configure it again in the future)

download RPMs using link from above and install using yum localinstall:
oracle-instantclient12.1-basic-12.1.0.2.0-1.x86_64.rpm
oracle-instantclient12.1-sqlplus-12.1.0.2.0-1.x86_64.rpm

Update PATH and LD_LIBRARY_PATH so sqlplus can start:
LD_LIBRARY_PATH=/usr/lib/oracle/12.1/client64/lib
export LD_LIBRARY_PATH
export PATH=$PATH:/usr/lib/oracle/12.1/client64/bin

And then you should be able to connect to Oracle using something like:
sqlplus system/password@ny-gbraylx01:49161

gbrayut:
Merged in db69b5aab7bcc5893541bdc138cce7ac219d1b96

noblehng:
Thanks for merging this!

PR #1499, opened 2015-12-08
cmd/scollector: Add fortinet collector

kylebrandt:


cmd/scollector/collectors/snmp_fortinet.go
```diff
@@ -0,0 +1,117 @@
...
+	"fmt"
+	"strconv"
+	"time"
+
+	"bosun.org/cmd/scollector/conf"
+	"bosun.org/metadata"
+	"bosun.org/opentsdb"
+)
+
+const (
```
captncraig: can we define these inside the c_fortinet_os function plz?

captncraig:
lgtm otherwise

cmd/scollector/collectors/snmp_fortinet.go
kylebrandt: @captncraig  That would be change form all the other snmp collectors. We can do that, but would rather it be a commit changing all of them.

kylebrandt:
pushed manually to master

PR #1500, opened 2015-12-08
Tune elasticsearch collector.

alienth:
The cluster index stats are identical across all nodes in the cluster, so it only makes sense to collect them from a single node (in this case, I'm using the master).

Also collecting index counts by status in `elastic.health.index_status_count`. This will allow us to measure index health without tracking each individual index in a separate tag, including indices that are excluded by the filter.

:eyeglasses:  @kylebrandt  @captncraig @gbrayut

### DIFF
--- a/cmd/scollector/collectors/elasticsearch.go
+++ b/cmd/scollector/collectors/elasticsearch.go
@@ -180,8 +180,10 @@ func c_elasticsearch(collectIndices bool) (opentsdb.MultiDataPoint, error) {
 	var md opentsdb.MultiDataPoint
 	s := structProcessor{elasticVersion: status.Version.Number, md: &md}
 	ts := opentsdb.TagSet{"cluster": clusterStats.ClusterName}
+	isMaster := false
+	// As we're pulling _local stats here, this will only process 1 node.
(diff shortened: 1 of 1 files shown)


cmd/scollector/collectors/elasticsearch.go
```diff
@@ -173,42 +173,52 @@ func c_elasticsearch(collectIndices bool) (opentsdb.MultiDataPoint, error) {
...
+	}
 	var md opentsdb.MultiDataPoint
 	s := structProcessor{elasticVersion: status.Version.Number, md: &md}
 	ts := opentsdb.TagSet{"cluster": clusterStats.ClusterName}
+	isMaster := false
+	// As we're pulling _local stats here, this will only process 1 node.
 	for nodeID, nodeStats := range clusterStats.Nodes {
-		isMaster := nodeID == clusterState.MasterNode
+		isMaster = nodeID == clusterState.MasterNode
 		if isMaster {
```
captncraig: can we do `if !isMaster {continue}` ?

captncraig: nvm

captncraig:
LGTM

PR #1501, opened 2015-12-08
c_keepalived: Add new SNMP entry so we work with new keepalived 1.2.19,  and fix off-by-one bug.

alienth:
:eyeglasses:  @captncraig  @gbrayut @kylebrandt


captncraig:
lgtm

PR #1502, opened 2015-12-09
c_nexpose: Gather data on asset groups.

alienth:
Will allow us to compare risk scores for different asset groups.

:eyeglasses: @captncraig @kylebrandt


captncraig:
lgtm

PR #1504, opened 2015-12-09
cmd/scollector: Break up cisco collectors by OS, and normalize mem in…

kylebrandt:
…formation

This will be a breaking change to the config file since the built-in SNMP names will change


captncraig:
LGTM. This essentially confirms to me that generic snmp has limited usefulness. Or that cisco is silly for having such inconsistent mibs. Not sure which.

kylebrandt:
I think has two use cases:
 1) More one off stuff (like PDUs)
 2) To be used within the code

kylebrandt:
pushed manually to master

PR #1512, opened 2015-12-11
Move snmp into the bosun.org repo since third party source is gone

kylebrandt:


gbrayut:
This was moved to the bosun repo and already has a few issues / pull requests https://github.com/bosun-monitor/snmp/pulls

kylebrandt:
Will work on getting the PRs moved to bosun. Rather have it our repo since
changes happen in both places. This isn't really a "snmp" for everyone I
don't think, it is an snmp for scollector's uses of it.

We also never changed anything over to using that repo in Bosun /
scollector it seems

gbrayut:
yeah... I think I moved it because someone started working on new features and Matt indicated he wasn't working on it anymore.

Sounds good to me. Just delete the bosun-monitor/smtp repo when things are moved over.

PR #1515, opened 2015-12-14
cmd/scollector+bosun: get vsphere mounted datastores and expose via h…

kylebrandt:
…ost api

### DIFF
--- a/cmd/bosun/sched/host.go
+++ b/cmd/bosun/sched/host.go
@@ -440,6 +440,24 @@ func (s *Schedule) Host(filter string) (map[string]*HostData, error) {
 					if iface != nil {
 						iface.Description = val
 					}
+				case "dataStores":
+					dataStores := []string{}
+					err = json.Unmarshal([]byte(val), &dataStores)
+					if err != nil {
+						slog.Errorf("error unmarshalling datastores for host %s while generating host api: %s", host.Name, err)
+					}
+					for _, dataStore := range dataStores {
+						tags := opentsdb.TagSet{"name": dataStore}.String()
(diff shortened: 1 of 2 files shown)


cmd/bosun/sched/host.go
```diff
@@ -440,6 +440,24 @@ func (s *Schedule) Host(filter string) (map[string]*HostData, error) {
 					if iface != nil {
 						iface.Description = val
 					}
+				case "dataStores":
+					dataStores := []string{}
+					err = json.Unmarshal([]byte(val), &dataStores)
+					if err != nil {
+						slog.Errorf("error unmarshalling datastores for host %s while generating host api: %s", host.Name, err)
```
captncraig: continue or return here. Don't keep going.

kylebrandt: @captncraig Since the host API generates a lot of information, I don't want single errors preventing the rest of the information from being returned

captncraig: yeah, then we chould continue the metadata loop. Should not iterate over dataStores if we just failed to deserialize it.

kylebrandt: @captncraig My assumption was that in we fail to deserialize it, the array will be empty, so the range is a no-op.

captncraig: It is possible it has a few partial objects already created there. I'd rather skip it to be safe.

kylebrandt: @captncraig The code makes it sounds like it checks for this first, see https://golang.org/src/encoding/json/decode.go?s=2621:2669#L64

PR #1530, opened 2015-12-18
cmd/bosun: Make graph links from OpenTSDB 2.1 work with 2.2

### LINKED ISSUE #1528: New opentsdb 2.2 GUI on graph page breaks tag filters on existing short links
We have a lot of short links to saved graphs: Some in chat room logs, some in personal notes, and some embedded in alerts. The new opentsdb 2.2 UI of the graph tab added new ways to filter tags, but it also broke all the existing links. It would be nice if those links still worked or redirected to a new valid url with the same filters. If not they should at least indicate that the previously saved tag filters are no longer applied and/or display what those were so the user can re-create the graph.

kylebrandt:
Fixes #1528


captncraig:
lgtm

PR #1542, opened 2015-12-30
[WIP, Delay]: Amend cAdvisor collector

PeterGrace:
This PR does three things:

 - Add a docker_host field that regexp-collects the non-qualified hostname of the API endpoint
 - Remove the "name" tagk, as it is essentially superfluous vs. docker_id field
 - Add a config variable "ShortContainerIds" to limit the docker_id field to 12 character container ids, which is what is shown for `docker ps` command and saves opentsdb tagv working-space (we don't really need those ultra long container ids, do we?)

Before this is merged, I'd really like input from @nickethier on his motivations behind having `name` and `docker_id` being distinct tagk's.  Please comment at your convenience, thanks!

### DIFF
--- a/cmd/scollector/collectors/cadvisor.go
+++ b/cmd/scollector/collectors/cadvisor.go
@@ -1,6 +1,7 @@
 package collectors
 
 import (
+	"regexp"
 	"strconv"
 
 	"bosun.org/_third_party/github.com/google/cadvisor/client"
@@ -15,6 +16,9 @@ import (
 func init() {
 	registerInit(startCadvisorCollector)
 }
+
+var shortContainerIds bool
+var apiHostName string
 
 var cadvisorMeta = map[string]MetricMeta{
 	"container.cpu.system": {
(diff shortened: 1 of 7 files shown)


nickethier:
The rational behind using the name tagk was for use with kubernetes. K8s embeds some meta data like the pod name and I'd into the container name. See https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/dockertools/docker.go#L216

I believe k8s now uses docker labels so we could do 1 of 2 things if we want to remove name.

 * If the name starts with the k8s prefix then parse out the k8s attributes as their own tags.
 * Add an option to enable consuming the docket labels as independent tags.

Thoughts?

mathpl:
I actually have an upcoming [PR](https://github.com/mathpl/bosun/compare/feature/local_listener...mathpl:feature/collector_tag_override) related to this which I think you could benefit from. The idea is to do a bit of post processing on tags to get the best value out of the collected data independently of a specific collector, it applies for cadvisor generated tags but we also use it for data generated by the snmp collector. Here's an example config file which I think would replicate the behavior you want:
```
filter = [ 'cadvisor' ]
hostname = 'k8s-1.my.corp'

[[Cadvisor]]
    URL='http://localhost:4194/'

[[TagOverride]]
    CollectorExpr = 'cadvisor'
    [TagOverride.MatchedTags]
        docker_name = 'k8s_(?P<container_name>[^\.]+)\.[0-9a-z]+_(?P<pod_name>[^\.]+)'
        docker_id = '^(?P<docker_id>.{12})'
        host = '^(?P<docker_host>[^\.]+)'
    [TagOverride.Tags]
        name = ''
        docker_name = ''
        provider = 'kubelet'
```
Which would yields:
```
{"metric":"container.cpu.usage","timestamp":1451493764,"value":2349620289,"tags":{"container_name":"my-container-name-in-pod","docker_host":"k8s-1","cpu":"12","docker_id":"8d6ecb2a8b2c","host":"k8s-1.my.corp","pod_name":"my-pod-name","provider":"kubelet"}}
```

I didn't take the time to look at getting dockers labels for tags but I believe pod labels are not applied to containers yet. From my testing they don't show up in the cadvisor api responses. 

I agree name could be removed. Do you think this could work for you?

PeterGrace:
I like where you're going with this pr, @mathpl!  If your PR is accepted then it would really make my changes superfluous.  I'll wait to see where it goes!

PR #1543, opened 2015-12-31
Add ASA-specific checks to snmp_cisco collector

PeterGrace:
The snmp-cisco collector did not have any asa-specific checks in it, even though ASAs were strutted out in the code.  I have added the first of what i hope to be many more ASA-specific checks to aid in ascertaining the fitness of a Cisco ASA firewall.


PeterGrace:
I'm going to resubmit this PR with those additional superfluous commits removed.

PR #1544, opened 2015-12-31
Add ASA-specific checks to snmp_cisco collector

PeterGrace:
The `snmp_cisco.go` file has struts for ASAs but currently no specific checks for ASAs.  This branch will add a new c_cisco_asa parent function to allow ASA-specific checks, and also adds the first specific check to the collector: reporting current/max open connections on the ASA.


PeterGrace:
need to gofmt to make this a single commit.

Issue #1564, opened 2016-01-19
Allow port configuration option for hbase unit


pEdler:
Currently the ports for the hbase unit are hardcoded in the 60000 range. However, by default HortonWorks Ambari sets the ports in the 16000 range. So unless the default port is changed on the hbase end no metrics are reported.

PR #1567, opened 2016-01-23
cmd/scollector: Added hadoop host and port config option.

jareksm:


jareksm:
This commit fixes #1564

gbrayut:
There was an issue with the append call happening before the URLs were set by registerInit, but I fixed that and this LGTM now.

I tested using the docker image and exposing the 16010 port used by HBASE:

    docker run --name bosun --restart=always -d -p 4242:4242 -p 8070:8070 -p 16010:16010 stackexchange/bosun

On another system (note hbase_unix.go is not usually built on windows) I then added `HadoopHost = "mydockerhost:16010"` to the scollector.toml file, and after a few minutes the hbase.* metrics started working. When HadoopHost is excluded from the config it defaults to localhost:60030 as before.

I'll rebase and merge this commit into master shortly.

Issue #1069, opened 2015-06-13
scollector: df command not working as non root with docker on the host


gregdel:
c_dfstat_blocks_linux and c_dfstat_inodes_linux use the "df" command and parse the output.

Running the df command as a non root user on debian with docker installed on the host, the "df" command returns a "Permission denied" error which is not handled by scollector.

Here is the output of the df command:
```
greg@dev:~$ df -lP --block-size 1
df: `/var/lib/docker/aufs': Permission denied
Filesystem                                                1-blocks        Used  Available Capacity Mounted on
rootfs                                                 20995723264 10222010368 9683828736      52% /
udev                                                      10485760           0   10485760       0% /dev
tmpfs                                                    210792448      237568  210554880       1% /run
/dev/disk/by-uuid/424d4290-b08a-5e38-b55e-ea8f66023ce2 20995723264 10222010368 9683828736      52% /
tmpfs                                                      5242880           0    5242880       0% /run/lock
tmpfs                                                    421580800           0  421580800       0% /run/shm
cgroup                                                  1053954048           0 1053954048       0% /sys/fs/cgroup

```

scollector logs:
```
2015/06/13 14:53:19 error: interval.go:61: bosun.org/cmd/scollector/collectors.c_dfstat_blocks_linux: exit status 1
2015/06/13 14:53:19 error: interval.go:61: bosun.org/cmd/scollector/collectors.c_dfstat_inodes_linux: exit status 1
```

Debian version: 7.8
Kernel: 3.16-0.bpo.3-amd64

kylebrandt:
scollector generally needs root for various things, so closing this as running scollector without elevated privileges isn't really on our roadmap at this time.

PR #1570, opened 2016-01-27
cmd/scollector: Optimise performance for linux stat collection.

jareksm:
On a big server with lots of network interfaces, disks, processes, VMs, etc... the stat collection is very heavy. After running pprof I saw the biggest offending functions and I have stream lined them.

Replacing simple regex with strings methods.
Replace IsDigit and IsAlNum with strconv methods.
Use /proc/mounts and Statfs() instead of 'df', simplifies the code. Less
string processing since it returns int.
Handle filesystems with spaces, use internal functions
Rewrote isPseudo() as a closure. At the moment /proc/filesystems
is read for every filesystem being checked, now it is only read once per
check.
removed IsDigit and IsAlNum, replaced them with strconv methods.
Use Sysinfo() instead of /proc/(uptime|loadavg), faster and less string
processing. But it results in losing linux.uptime_now and
linux.loadavg_runnable metrics.
Changed "linux.loadvg_" to "linux.loadvg."


jareksm:
Updated the pull-request with better rebase/squash.
The original had a bug where it would skip checking filesystems.
This one works correctly.

cmd/scollector/collectors/disk_linux.go
```diff
@@ -158,23 +141,45 @@ func c_iostat_linux() (opentsdb.MultiDataPoint, error) {
...
+		return true
+	}
+
+	err := readLine("/proc/mounts", func(line string) error {
+		var sfs unix.Statfs_t
 		fields := strings.Fields(line)
-		// TODO: support mount points with spaces in them. They mess up the field order
-		// currently due to df's columnar output.
-		if len(fields) != 7 || !IsDigit(fields[2]) {
+		if isPseudo(fields[2]) {
```
kylebrandt: Need to check len(fields) to avoid potential panic

jareksm:
Update with check for number of fields in /proc/mounts.

kylebrandt:
This might fix https://github.com/bosun-monitor/bosun/issues/1069

cmd/scollector/collectors/procstats_linux.go
```diff
@@ -36,171 +28,165 @@ var CPU_FIELDS = []string{
...
+	Add(&md, "linux.loadavg.1_min", sys.Loads[0], nil, metadata.Gauge, metadata.Load, "")
+	Add(&md, "linux.loadavg.5_min", sys.Loads[1], nil, metadata.Gauge, metadata.Load, "")
+	Add(&md, "linux.loadavg.15_min", sys.Loads[2], nil, metadata.Gauge, metadata.Load, "")
+	Add(&md, "linux.loadavg.total_threads", sys.Procs, nil, metadata.Gauge, metadata.Process, "")
+	if sys.Totalram != 0 {
+		Add(&md, osMemPctFree, (sys.Freeram)/(sys.Totalram)*100, nil, metadata.Gauge, metadata.Pct, osMemFreeDesc)
 	}
 	if err := readLine("/proc/vmstat", func(s string) error {
-		m := vmstatRE.FindStringSubmatch(s)
+		m := strings.Split(s, " ")
```
kylebrandt: Would strings.Fields be appropriate here?

cmd/scollector/collectors/procstats_linux.go
```diff
@@ -36,171 +28,165 @@ var CPU_FIELDS = []string{
...
 		if m == nil {
 			return nil
 		}
-		i, err := strconv.ParseFloat(m[2], 64)
-		if err != nil {
-			return err
-		}
-		mem[m[1]] = i
-		Add(&md, "linux.mem."+strings.ToLower(m[1]), m[2], nil, metadata.Gauge, metadata.KBytes, "")
+		m[1] = strings.TrimSpace(m[1])
```
kylebrandt: Could panic if len(m) == 1

cmd/scollector/collectors/procstats_linux.go
```diff
@@ -36,171 +28,165 @@ var CPU_FIELDS = []string{
...
+		switch m[0] {
+		case "pgpgin", "pgpgout", "pswpin", "pswpout":
+			switch {
+			case strings.HasSuffix(m[0], "in"):
+				Add(&md, "linux.mem."+strings.TrimSuffix(m[0], "in"), m[1], opentsdb.TagSet{"direction": "in"}, metadata.Counter, metadata.Page, "")
+			case strings.HasSuffix(m[0], "out"):
+				Add(&md, "linux.mem."+strings.TrimSuffix(m[0], "out"), m[1], opentsdb.TagSet{"direction": "out"}, metadata.Counter, metadata.Page, "")
 			}
+		case "pgfault", "pgmajfault":
+			Add(&md, "linux.mem."+m[0], m[1], nil, metadata.Counter, metadata.Page, "")
```
kylebrandt: same here, need to know len m > 1

cmd/scollector/collectors/procstats_linux.go
```diff
@@ -36,171 +28,165 @@ var CPU_FIELDS = []string{
...
+		switch {
+		case strings.HasPrefix(m[0], "cpu"):
+			tag_cpu := strings.TrimPrefix(m[0], "cpu")
+			if tag_cpu != "" {
+				num_cores++
 			}
-			fields := strings.Fields(m[2])
-			for i, value := range fields {
-				if i >= len(CPU_FIELDS) {
+			for i, value := range m[1:] {
```
kylebrandt: check len here to. Basically m will never be nil now that split is being uses, so the nil check is pointless, and need to check len.

cmd/scollector/collectors/procstats_linux.go
jareksm: It could be, but there is no need to check for tabs, multiple spaces since the separator is always a single space. If you look at the code for fields, it's a lot more complex and there's a lot more overhead. I have checked it in simple benchmark and for a single space Split is roughly 3 times faster than Fields. The longer the line the larger the margin.

The idea behind this patches is to get the checks faster (lighter) so I can increase their frequency with minimal load increase.

cmd/scollector/collectors/procstats_linux.go
jareksm: Fixed all of the above checks in the latest patch

PR #1576, opened 2016-02-03
cmd/scollector: Configurable MaxQueueLen

mathpl:
This adds MaxQueueLen as option in the config file. If undefined it keeps the current default of 200000. If defined it must be greater or equal than the batch size.


captncraig:
merged via 1022108854b869f78102f1026e382ea7754406d3

Issue #1523, opened 2015-12-16
scollector: Auto-discover non-default instances of SQL Server


floyd-may:
I'm not getting SQL Server monitoring from instances of SQL server whose instance name isn't the default, "MSSQLServer".

From [code-surfing here](https://github.com/bosun-monitor/bosun/blob/master/cmd/scollector/collectors/sql_windows.go), it appears that the WMI queries are generated from the struct names, which correspond to WMI class names (I think?). If so, that doesn't work for named instances of SQL Server. If your instance name is `DERP`, rather than getting general stats from a WMI class named `Win32_PerfRawData_MSSQLSERVER_SQLServerGeneralStatistics`, you get them from a class called `Win32_PerfRawData_MSSQLDERP_MSSQLDERPGeneralStatistics`. Fun :confused:

floyd-may:
Bueller?

gbrayut:
This was also on my list... I think we were going to look at opt in monitoring of named instances. The different WMI class names make it a bit tricky, and they may end up needing to be under a different namespace (mssql.instance.*). I'll take another swing at it and let you know later this week what I find.

floyd-may:
Thanks, and sorry for being a pest =)

What's the rationale for putting it under a different namespace? That seems like a prime example of what tags are for. If there's only ever one instance per host, regardless of whether or not it's the default instance, I would expect a given metric query to Just Work (tm). If I have to alter the metric query based on some sort of out-of-band knowledge that some instances are configured with the default name and some aren't, that seems strange. If I need to break it down by instance, I can add the tag to the query.

gbrayut:
We often are a bit tag heavy on Bosun (most opentsdb instances use much more verbose metric names instead of lots of tags) and it has hurt us before (our os.net.bytes metric is so slow we had to denomalize it into __hostname.os.net.bytes for queries that return more than one day of data), so I just have a knee jerk reaction on some of these.

It does fit the tag model pretty well, so I'll try and get it in that way. In my defense named instances do often behave differently at a networking/firewall level (ever tried to setup saved network credentials in windows for one?), which is one reason why we only use the default instance here at Stack Overflow. I've used them a bunch in the past and they work great for development, so I'll see what I can do.

floyd-may:
No argument here about the pain that named instances cause =)

VictoriaD:
Has any progress been made with this one? In the meantime, is there a way to manually configure scollector to look for these statics?

[...]

PR #1579, opened 2016-02-08
cmd/scollector/collectors/sql_windows.go scollector: Auto-discover non-default instances of SQL Server

VictoriaD:
Checks for named instances of sql server and runs the queries for all of these.
Uses an instance tag.


VictoriaD:
This is for issue #1523 scollector: Auto-discover non-default instances of SQL Server

gbrayut:
Thanks for working on this. I will try and merge it under #1671

gbrayut:
merged in #1671

PR #1590, opened 2016-02-12
cmd/bosun: Shorten-only http proxy

mathpl:
This adds shortenProxy as option in bosun.toml to add a proxy only to requests out to googleapis' shortener api. The alternative is making use of the default http client's http.ProxyFromEnvironment but it results in all http calls to use it. That includes calls to OpenTSDB which isn't a something we want.

### DIFF
--- a/cmd/bosun/main.go
+++ b/cmd/bosun/main.go
@@ -136,6 +136,12 @@ func main() {
 			c.TSDBHost = tsdbHost.Host
 		}
 	}
+	if c.InternetProxy != "" {
+		web.InternetProxy, err = url.Parse(c.InternetProxy)
+		if err != nil {
+			slog.Fatalf("InternetProxy error: %s", err)
+		}
+	}
 	if *flagQuiet {
 		c.Quiet = true
 	}
--- a/cmd/bosun/conf/conf.go
+++ b/cmd/bosun/conf/conf.go
@@ -64,6 +64,7 @@ type Conf struct {
 	Quiet            bool
 	NoSleep          bool
 	ShortURLKey      string
+	InternetProxy    string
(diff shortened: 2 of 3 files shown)


captncraig:
Silly question: what problem does this solve?

mathpl:
Our outbound proxies are on a different network than where OpentTSDB is. In our case we simply cannot use the same for both sets of calls. I also think going through a proxy to fetch data from OpenTSDB is generally unneeded in most deployments, unlike reaching out to googleapis.

gbrayut:
yeah... Internet calls vs Intranet calls makes sense. I'd change the name to InternetProxy so it can be reused if we ever add any other APIs, but otherwise sounds fine to me.

cmd/bosun/main.go
```diff
@@ -136,10 +136,17 @@ func main() {
...
+		shortenProxy, err = url.Parse(c.ShortenProxy)
+		if err != nil {
+			slog.Fatalf("ShortenProxy error: %s", err)
+		}
+	}
 	if *flagQuiet {
 		c.Quiet = true
 	}
-	go func() { slog.Fatal(web.Listen(c.HTTPListen, *flagDev, c.TSDBHost)) }()
+	go func() { slog.Fatal(web.Listen(c.HTTPListen, *flagDev, c.TSDBHost, shortenProxy)) }()
```
captncraig: I don't love the idea of yet another argument to Listen. How about a package variable in web that you set when reading the config above here?

mathpl:
How about this?

captncraig:
Thanks for those changes. I like it.

gbrayut:
LGTM

PR #1591, opened 2016-02-12
Esfix

kylebrandt:


kylebrandt:
Waiting on confirmation of fix from @awesensepaul before we should merge

awesensepaul:
I can confirm that this fix is **awesome**! (and both commits/fixes were necessary to get correct ES 2.x behaviour).

FYI It is a huge relief to be able to use the same queries we use in Kibana (instead of re-writing them into regexes just for `bosun`).

Thanks!

captncraig:
Merge anytime

kylebrandt:
I forgot to pushed my rebased branch, already merged.

PR #1598, opened 2016-02-16
cmd/bosun: Add over, shift, and merge funcs

kylebrandt:
This adds the ability to graph multiple seriesSets in expression graphs, and also graph things like week-over-week.

### DIFF
--- a/cmd/bosun/expr/funcs.go
+++ b/cmd/bosun/expr/funcs.go
@@ -101,6 +101,18 @@ var TSDB = map[string]parse.Func{
 		Tags:   tagQuery,
 		F:      Band,
 	},
+	"shiftBand": {
+		Args:   []models.FuncType{models.TypeString, models.TypeString, models.TypeString, models.TypeScalar},
+		Return: models.TypeSeriesSet,
+		Tags:   tagQuery,
+		F:      ShiftBand,
+	},
+	"over": {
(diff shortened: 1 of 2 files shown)


cmd/bosun/expr/funcs.go
```diff
@@ -371,6 +396,43 @@ func Filter(e *State, T miniprofiler.Timer, series *Results, number *Results) (*
 	return series, nil
 }
 
+func Merge(e *State, T miniprofiler.Timer, series ...*Results) (*Results, error) {
+	if len(series) < 1 {
```
captncraig: I'd prefer `== 0` and `== 1` for these two short circuit clauses.

cmd/bosun/expr/funcs.go
```diff
@@ -732,6 +796,91 @@ func Band(e *State, T miniprofiler.Timer, query, duration, period string, num fl
 	return
 }
 
+func ShiftBand(e *State, T miniprofiler.Timer, query, duration, period string, num float64) (r *Results, err error) {
```
captncraig: does this function need to be exported?

captncraig: It looks like it should be, but it is not documented.

kylebrandt: By convention, in funcs.go, all the captial functions work with Sets, and the lowercase ones work within results within the set. So I am following that.

PR #1599, opened 2016-02-16
cmd/bosun: Implementing purge and forceCLose actions

captncraig:
Purge - For completely removing all traces of an alert key. Like forget without validation. Useful for decomissioning a host with alerts stuck open.

Force Close - For closing an alert stuck open. Usually an "ignore unkown" with no new data coming in. Will reopen if expression is still in a critical state.


captncraig:
cc @gbrayut

PR #1610, opened 2016-02-22
cmd/bosun: annotate support

kylebrandt:


kylebrandt:
yiwa6ucoka

cmd/bosun/web/web.go
```diff
@@ -119,6 +124,25 @@ func Listen(listenAddr string, devMode bool, tsdbHost string) error {
...
+			index = "annotate"
+		}
+		annotateBackend, err = backend.NewElastic(schedule.Conf.AnnotateElasticHosts, index)
+		if err != nil {
+			return err
+		}
+		if err := annotateBackend.InitBackend(); err != nil {
+			return err
+		}
+		web.AddRoutes(router, "/api", []backend.Backend{annotateBackend}, false, false)
```
captncraig: I like this way of doing it a lot. What are the last two bool parameters for? Mostly out of curiosity.

captncraig: nvm. read the code.

captncraig: are we gonna try to get the ui working at some point?

kylebrandt: enable the independent UI, and use local dev files when doing so

kylebrandt: The UI does work, but I don't intend to just import the UI into bosun, Bosun will integrate with it using it's own UI.

captncraig:
LGTM

cmd/bosun/web/web.go
captncraig: was thinking it might be nice if your ui has direct query/filter/ raw viewing support.

kylebrandt: Agreed, approaching this incremently and thinking of it as probably a "preview" feature when we release

PR #1615, opened 2016-02-23
cmd/bosun: pprof for bosun

mathpl:
Option to enable a pprof http endpoint for bosun. Essentially the same code as the one in scollector.


gbrayut:
I thought this already worked. we have http://bosun.ds.stackexchange.com/debug/pprof working right now

mathpl:
Woops, looks like I missed it. Indeed it works, disregard. :)

gbrayut:
I'll add it to the list of things that need documentation :-P

PR #1616, opened 2016-02-23
cmd/tsdbrelay: accept responses with 2xx code

ipeters:
opentsdb may return 204 or 200 on successful PUTs to /api/put. 204 indicates success with no response body, but when the details parameter is specified, a 200 is returned along with a simple response document. tsdbrelay should relay all successful PUTs to bosun.

I am explicitly not changing relayMetadata, since bosun itself appears to only return a 204.

(Also fix a missing format specifier in the verbose() call.)


gbrayut:
Travis issues were unrelated. Fixed and merged your commit to master. Closed via #1617

PR #1617, opened 2016-02-24
travis: change go to version 1.6

gbrayut:


gbrayut:
Looks like Matt merged a few PRs on esc. If you are getting travis errors for go generate you need to:
go get -u -v github.com/mjibson/esc
go install github.com/mjibson/esc

Also Kyle pushed right in the middle, so I rebased and pushed to master

PR #1622, opened 2016-02-25
cmd/scollector: add metadata to redis collector

kylebrandt:
also add rejected_connections stat


gbrayut:
running now as a test instance pointed at http://ny-devbosun.ds.stackexchange.com/items

gbrayut:
Made some comments about typos and missing metrics. Otherwise looks good to me.

This will make reading redis graphs much easier. Thanks!

PR #1624, opened 2016-02-26
cmd/scollector: fix redis and misc errors

gbrayut:
I hacked in some code to get rid of a few errors we are seeing. @alienth thinks there is a better way to detect redis running in docker instead of natively on the system, but I tested this and it works on ny-intdiscourse01 so I'd rather fix it now and clean things up when we have more time.

@captncraig @kylebrandt  let me know if there are any bugs you can see

### DIFF
--- a/cmd/scollector/collectors/redis_unix.go
+++ b/cmd/scollector/collectors/redis_unix.go
@@ -285,7 +285,7 @@ func redisInit() {
 				return nil
 			}
 			port := strings.Split(sp[2], ":")[1]
-			if port != "0" {
+			if port != "0" && InContainer(sp[0]) == false {
 				add(port)
 			}
 			return nil
--- a/cmd/scollector/collectors/processes_linux.go
+++ b/cmd/scollector/collectors/processes_linux.go
@@ -336,3 +336,14 @@ func (i *idPool) get() int {
 func (i *idPool) put(v int) {
 	i.free = append(i.free, v)
 }
+
+// InContainer detects if a process is running in a Linux container.
+func InContainer(pid string) bool {
+	pidNameSpaceFile := fmt.Sprintf("/proc/%v/ns/pid", pid)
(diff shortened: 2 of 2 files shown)


cmd/scollector/collectors/redis_unix.go
```diff
@@ -286,7 +286,10 @@ func redisInit() {
 			}
 			port := strings.Split(sp[2], ":")[1]
 			if port != "0" {
-				add(port)
+				_, err := util.Command(time.Second, nil, "grep", "-q", "cpu:/docker", fmt.Sprintf("/proc/%s/cgroup", sp[0]))
```
alienth: Would definitely prefer that this compare the /proc/%s/ns/proc link to pid 1's proc NS, rather than checking for the 'docker' string in the cgroups. Docker isn't the only container system folks are likely to use.

If that is truly difficult, then I'm fine with leaving this as-is for the time being, but it needs a FIXME comment.

gbrayut: I was able to get it to work but it will add a dependency on bash (which I'm fine with). Seems to work testing on ny-intdiscourse01 and ny-gbraylx01, and it's failure mode is to default to be enabled.

If someone is using namespaces but the redis instance is still accessible via localhost then this will cause redis to not be monitored, where it was previously so long as the connection to :6379 or whatever the port was worked. I'm also fine with addressing this in the future if that is ever reported.

alienth:
lgtm

gbrayut:
hmm... this PR got screwed up, but I added Kyle's code from https://github.com/bosun-monitor/bosun/commit/5fc091ef37e14900675bf88c7cc32ce3af8cb18e#commitcomment-16436167 and pushed to master in e412126c63a253f4dfb3a3ed5b3dc79a8e1ba9e3

PR #1263, opened 2015-08-19
Compile error fixes + negative filter patterns + total_time metric for httpunit collector

rajder:
Can now disable collectors using filter patterns prefixed by "-", eg. scollector-windows-amd64.exe -p -d -l -f="network_windows,-tcp" on my dev. machine prints:
bosun.org/cmd/scollector/collectors.c_network_windows

(excluding c_network_windows_tcp)

I needed this specifically for disabling c_network_windows_tcp while keeping c_network_windows enabled, trying to find performance issues caused by one of the two collectors, or both. High CPU usage by WmiPrvSE.exe and blocked collectors seems to be the symptoms (have not solved the issue or confirmed the offending collector yet)
Edit: It was not one of the network collectors. Could be some WMI related issue. It seems like the issue mostly occurs when the machines are under heavy load.

PR #1630, opened 2016-03-01
cmd/scollector: negative -f filters and total_time metric for httpunit collector 

gbrayut:
Updated version of https://github.com/bosun-monitor/bosun/pull/1263 rebased on current master and with more documentation


cmd/scollector/collectors/httpunit.go
```diff
@@ -10,33 +10,34 @@ import (
...
-	HTTPUnitPlans(filename, &httpunit.Plans{
-		Plans: plans,
-	})
+	HTTPUnitPlans(filename, &httpunit.Plans{Plans: plans}, freq)
 	return nil
 }
 
-func HTTPUnitPlans(name string, plans *httpunit.Plans) {
+func HTTPUnitPlans(name string, plans *httpunit.Plans, freq time.Duration) {
+	if freq < time.Second {
```
captncraig: this seems a bit odd to me. If they didn't set it, and it was zero, I would understand. But if they set `15` expecting it to be seconds or minutes or who knows what, 5 mins is an odd default. Maybe error with a helpful message?

cmd/scollector/main.go
```diff
@@ -124,11 +125,12 @@ func main() {
 		check(collectors.AddProcessDotNetConfig(p))
 	}
 	for _, h := range conf.HTTPUnit {
+		freq := time.Second * time.Duration(h.Freq)
```
captncraig: We use `time.ParseDuration` in some places, but apparently not others. Is a number really better than a duration string if you have to look up the docs to know it is seconds?

captncraig:
Other than craig being confused about timings, it looks good to me.

gbrayut:
@captncraig I changed it to use a duration string... will be clearer in the config files

PR #1640, opened 2016-03-03
 vSphere collector with performance counters

simnv:
New metrics:
- vsphere.perf.* for host stats
- vsphere.guest.perm.* for virtual machine stats

Metrics are populated dynamically, names, units, rates and descriptions are pulled from vSphere itself.


simnv:
Made another pull request: #1644

PR #1645, opened 2016-03-04
Added ** expression operator -> q("q") ** 2 squares every element of the result query "q"

tkrajca:
Hi guys, added ** operator that runs math.Pow() on each element from a timeseries. Should all be pretty simple and standard. Let me know if this needs any work - formatting, docs, tests, etc. I am a python programmer so bosun and go are all pretty new to me. Cheers, Tomas


gbrayut:
sorry... master branch was broken because build tags don't work like we thought they did. If you rebase this on the current master the travis ci check should work now.

tkrajca:
Should be good now, cheers.

kylebrandt:
The standard arithmetic order of operations needs to be maintained. Currently that is not the case. Exponentiation should take precedence over multiplication, but it doesn't. So the following should be true, but isn't in the branch:

image

tkrajca:
Thanks @kylebrandt, good catch, that should be fixed now.

kylebrandt:
@tkrajca Can you rebase this on master? Will test one more time after rebase and then should be good to go.

tkrajca:
@kylebrandt I got quite a few merge conflicts when rebasing so I created a new branch of master Pow_operator2 and put all my changes in there as PR https://github.com/bosun-monitor/bosun/pull/1674

PR #1647, opened 2016-03-04
cmd/scollector: validate datapoints before adding them

mathpl:
Slightly related to https://github.com/bosun-monitor/bosun/pull/1646, this will validate a datapoint before sending it off in the datapoints channel. A problem I noticed in the bug fixed in the previous PR is that if we fail to json encode a batch nothing will be sent, thus a single faulty collector can break the pipeline. 

Example of the problem in question:
```
Mar  4 05:41:02 myhost scollector[1234]: error: queue.go:105: json: error calling MarshalJSON for type *opentsdb.DataPoint: Unparseable number MyKeyspace.my_idx
Mar  4 05:41:02 myhost scollector[1234]: info: queue.go:124: restored 500, sleeping 5s
...
```

### DIFF
--- a/cmd/scollector/collectors/collectors.go
+++ b/cmd/scollector/collectors/collectors.go
@@ -17,6 +17,7 @@ import (
 	"bosun.org/cmd/scollector/conf"
 	"bosun.org/metadata"
 	"bosun.org/opentsdb"
+	"bosun.org/slog"
 	"bosun.org/util"
 )
 
@@ -249,15 +250,6 @@ func AddTS(md *opentsdb.MultiDataPoint, name string, ts int64, value interface{}
 	} else if host == "" {
 		delete(tags, "host")
 	}
-	if rate != metadata.Unknown {
-		metadata.AddMeta(name, nil, "rate", rate, false)
-	}
-	if unit != metadata.None {
-		metadata.AddMeta(name, nil, "unit", unit, false)
-	}
-	if desc != "" {
-		metadata.AddMeta(name, tags, "desc", desc, false)
-	}
 
 	tags = AddTags.Copy().Merge(tags)
 	d := opentsdb.DataPoint{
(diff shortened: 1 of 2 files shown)


gbrayut:
sorry... master branch was broken because build tags don't work like we thought they did. If you rebase this on the current master the travis ci check should work now.

cmd/scollector/collectors/collectors.go
```diff
@@ -266,7 +258,20 @@ func AddTS(md *opentsdb.MultiDataPoint, name string, ts int64, value interface{}
 		Value:     value,
 		Tags:      tags,
 	}
-	*md = append(*md, &d)
+	if d.Valid() {
```
kylebrandt: Better to:
```
if !d.Valid() {
   slog.Errorf(...)
   return
}
```

That way the rest of the logic doesn't have to be indendented. See https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

cmd/scollector/collectors/collectors.go
```diff
@@ -266,7 +258,20 @@ func AddTS(md *opentsdb.MultiDataPoint, name string, ts int64, value interface{}
...
+			metadata.AddMeta(name, nil, "rate", rate, false)
+		}
+		if unit != metadata.None {
+			metadata.AddMeta(name, nil, "unit", unit, false)
+		}
+		if desc != "" {
+			metadata.AddMeta(name, tags, "desc", desc, false)
+		}
+	} else {
+		slog.Errorf("Invalid datapoint received for: %s", name)
```
kylebrandt: Would probably be useful to print all fields of the datapoint so one can see what was wrong with it.

kylebrandt:
https://github.com/bosun-monitor/bosun/pull/1647 Might prevent this.

mathpl:
Actually this problem is slightly more complicated than I previously though. Currently what would be considered an invalid metric by Valid() can be marshaled successfully thanks to this a call to clean() beforehand https://github.com/bosun-monitor/bosun/blob/eedf53ff4e87df1fc508414dce63994e52c6bde9/opentsdb/tsdb.go#L73 . However a metric that has been cleaned can still be invalid when the value can't be marshaled.

I'll reword this PR so clean() is done then Valid() is called when we consume DataPoints out of the channel, probably in there https://github.com/bosun-monitor/bosun/blob/master/collect/queue.go#L16. This way we'll retain the current behavior of fixing cleanable DataPoint without discarding them when it's possible to do so.

Let me know what you think.

captncraig:
Closed via #1779

PR #1651, opened 2016-03-06
cmd/bosun/(conf|sched) add Ledis bind address config option

Oded-B:
This will allow multiple instances to run on a single host.

Not sure I understood everything written in the CONTRIBUTING.md file(not a Go dev) but it did seem to pass the build/validate.sh script and it's fairly simple.


captncraig:
Thanks!

PR #1659, opened 2016-03-10
cmd/bosun: remove tags from response that were not requested

### LINKED ISSUE #1657: Tagset When Return of OpenTSDB Filter doesn't Equal Query
When you pass something like `host=*`, if some of the results have another tag say foo, and foo has only one possible value, it will be added to the response.

In the case of the group by tags, we make sure you always get back what you asked for. This needs to be done with Filters as well.

kylebrandt:
fixes #1657


captncraig:
LGTM

PR #1660, opened 2016-03-10
cmd/bosun: don't autocomplete old metrics on graph page

### LINKED ISSUE #1658: Hide old Metrics in the Graph Page
For day to day use, having all the metrics Bosun has ever seen in autocomplete makes it messy. Need to limit to metrics seen in the past X.

You could still query old metrics, they just wouldn't autocomplete. In the items page old metrics could still be show, but with a visual indicator that they are out of date.

kylebrandt:
Won't autocomplete metrics older than 2 days. Fixes #1658


captncraig:
LGTM

Issue #1655, opened 2016-03-08
Function returning a constant-valued time series


gnosek:
I'd like to graph the current alert threshold along with metric values. Can I pretty please have a function that returns a time series containing a stream of constant values?

kylebrandt:
Will be something like:

`hline(number, "tagkey", startDuration, endDuration)`

And will return a series with tagged with the tagkey. Will return a series with two datapoints, and (now-StartDuration, number) and (now-endDuration, number). Stream isn't needed since to draw a line you just need two points.

gbrayut:
can we generalize it to just `line(startNumber, endNumber, "tagkey", startDuration, endDuration)`?

kylebrandt:
Use case for that?

gbrayut:
wanting to show a linear increase of a noisy metic. Like linelr but manually specifying values

gbrayut:
Id also rather see one line function than hline() and slopeline()

kylebrandt:
Can use the series func to do this now.

PR #1667, opened 2016-03-14
cmd/bosun: series func to create series from scalars

kylebrandt:
Also fix nv() so it creates a result when the entire set is empty

```
$now = epoch()
$hourAgo =  $now-d("1h")
$s = merge(series("foo=bar", $hourAgo, 5, $now, 10), series("foo=bar2", $hourAgo, 6, $now, 11))
```

### DIFF
--- a/cmd/bosun/expr/parse/parse.go
+++ b/cmd/bosun/expr/parse/parse.go
@@ -31,13 +31,14 @@ type Tree struct {
 }
 
 type Func struct {
-	Args     []models.FuncType
-	Return   models.FuncType
-	Tags     func([]Node) (Tags, error)
-	F        interface{}
-	VArgs    bool
-	VArgsPos int
-	Check    func(*Tree, *FuncNode) error
+	Args      []models.FuncType
+	Return    models.FuncType
+	Tags      func([]Node) (Tags, error)
+	F         interface{}
+	VArgs     bool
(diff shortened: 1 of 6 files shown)


kylebrandt:
Also fixes #1655

cmd/bosun/expr/parse/parse.go
```diff
@@ -31,13 +31,14 @@ type Tree struct {
...
-	VArgs    bool
-	VArgsPos int
-	Check    func(*Tree, *FuncNode) error
+	Args      []models.FuncType
+	Return    models.FuncType
+	Tags      func([]Node) (Tags, error)
+	F         interface{}
+	VArgs     bool
+	VArgsPos  int
+	VArgsOmit bool
```
captncraig: What does VArgsOmit do?

kylebrandt: @captncraig Makes it so the vargs are optional. You are allowed to omit them.

captncraig: How about inverting it. Maybe `VArgsRequired` or even `MinVArgs` would be clearer.

kylebrandt: Most of the time it isn't true. So when defining funcs you don't have to
include it. So false is a better default

PR #1672, opened 2016-03-16
cmd/bosun: series operations

kylebrandt:
image

### DIFF
--- a/cmd/bosun/expr/expr.go
+++ b/cmd/bosun/expr/expr.go
@@ -189,6 +189,10 @@ func (s Series) MarshalJSON() ([]byte, error) {
 	return json.Marshal(r)
 }
 
+func (a Series) Equal(b Series) bool {
+	return reflect.DeepEqual(a, b)
+}
+
 type ESQuery struct {
 	Query elastic.Query
 }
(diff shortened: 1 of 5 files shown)


kylebrandt:
Useful for bools:

```
$count = q("sum:traffic.haproxy.route_tr_count{host=literal_or(ny-logsql01),route=Questions/Show}", "30m", "")
$avg = q("sum:traffic.haproxy.route_tr_avg{host=literal_or(ny-logsql01),route=Questions/Show}", "30m", "")
$avgCount = avg($count)
dropbool($avg, $count < $avgCount-$avgCount*.01 || $count > $avgCount+$avgCount*.01)
```

Says get me the average response time datapoints, but only when there was between +-1% of the average hits per seconds.

iamamberkulkarni:
Hey we are using bosun version 0.5.0-rc2.
Want to try out series operations so have created an expression as follows:
```
$a = q("sum:metric_name{filters...}", "7h", "")
$b = q("sum:metric_name{filters...}", "7h", "")
$a/$b
```
above expression gives the error
`expr: parse: type error in q("sum:filter{filters...}", "7h", "") / q("sum:filter{filters...}", "7h", ""): at least one side must be a number`

Was trying this out after seeing doc.
`Series Operations
If you combine two seriesSets with an operator (i.e. q(..) + q(..)), then operations are applied `
`for each point in the series if there is a corresponding datapoint on the right hand side (RH). A corresponding datapoint is one which has the same timestamp (and normal group subset rules apply).` `If there is no corresponding datapoint on the left side, then the datapoint is dropped. This is a new feature as of 0.5.0.`

Are we doing something wrong here ?
Please help. Thanks.

p.s. have not put actual metrics.

iamamberkulkarni:
Hey used latest code and it works. thanks a ton

PR #1676, opened 2016-03-18
cmd/bosun: link multiple queries from graph UI

kylebrandt:
This makes a simple expression so when you have multiple query tabs in the graph page, they all get linked to the expr or rule page. This avoid copy / paste and switching between tabs.

image

image


captncraig:
lgtm

PR #1684, opened 2016-03-27
Send actual hostname in SMTP Hello

n-st:
Sending "localhost" in the SMTP Hello can trip some spam filters that expect the Hello hostname to be a valid FQDN.

Note: I don't have a Go development environment set up, so this change is untested.
It seems simple enough, but perhaps someone could run a quick test to see if it runs alright.


captncraig:
util.Hostname will usually be something like "web01" or whatever. Wouldn't the same filters have issues with non-fully-qualified names. May be better to use the hostname as configured in bosun's `hostname` parameter

PR #1685, opened 2016-03-29
cmd/bosun: clear filters when changing metrics in graph view

### LINKED ISSUE #1679: Metric tags applied erroneously when performing subsequent graph lookups
When changing the 'metric' field on the graph page from one metric to another, any tags searched for on the first metric will apply to the second metric, even if the second metric has no such tag. This will result in 'no results' for the second metric if it does not have that tag.

To repro:

1. Graph some metric that has a number of tags. Fill in a value for one of those tags.
1. From the same view, change the 'metric' field to another metric which does not share the specified tag from above. Query this new metric.
1. The resulting graph will always have new results, as the query is still searching for that old tag, despite the tag disappearing from the form page. This can be verified by looking at the generated query at the bottom of the page.

kylebrandt:
Fixes #1679, Creates #??


kylebrandt:
Need to test this more before merging. This code is really messy at this point. There is probably repeated operations on tags and filters that don't need to be there. Once OpenTSDB is at v2.3 we can remove support for v2.1 and at that time I think it will be easier to redo this code.

kylebrandt:
There is still an edge case here where if you switch metrics, the two metrics share a tag key, but the value set for that tag is not present in the new metric it will still be there.

However, in that case this isn't hidden and you can just change the value, so I don't care to fix this at this point.

PR #1696, opened 2016-04-05
cmd/bosun: support actions by incident id

kylebrandt:


cmd/bosun/sched/sched.go
```diff
@@ -574,34 +574,49 @@ func init() {
 		"The running count of actions performed by individual users (Closed alert, Acknowledged alert, etc).")
 }
 
-func (s *Schedule) Action(user, message string, t models.ActionType, ak models.AlertKey) error {
-	if err := collect.Add("actions", opentsdb.TagSet{"user": user, "alert": ak.Name(), "type": t.String()}, 1); err != nil {
-		slog.Errorln(err)
-	}
+func (s *Schedule) ActionByAlertKey(user, message string, t models.ActionType, ak models.AlertKey) (*models.IncidentState, error) {
```
captncraig: I don't see anyone using this return value. Is it just to maintain parity with ActionByIncidentID. Does that need to return the whole incident, or just the alert key?

kylebrandt: Good point, I just want the alert key. Will change

PR #1700, opened 2016-04-08
Passing the username and password to the ElasticSearch client! =>

omar-m-othman:
This is not currently supported by Bosun, but it is supported by the ElasticSearch client it
  uses. So, this commit passes those from Bosun to the client, using, empirically, the first
  username/password it finds among the URLs in the "ElasticHosts" configuration option.

This can be considered as an initial implementation of this feature, we needed it (since our
  ElasticSearch cluster sits behind nginx for access control requirements) so I implemented
  a minimal proof-of-concept. This assumption (that a single username/password combination
  is enough) is due to the restriction of the ElasticSearch client that expects multiple
  URLs, but a single username/password combination (which says that it is a single-cluster
  client currently).

### DIFF
--- a/cmd/bosun/expr/elastic.go
+++ b/cmd/bosun/expr/elastic.go
@@ -214,7 +214,37 @@ type ElasticHosts []string
 func (e ElasticHosts) InitClient() error {
 	if esClient == nil {
 		var err error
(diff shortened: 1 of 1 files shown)


johnewing1:
Thanks for your contribution. This has been superseded by later work. 
Using username and password is supported in the current version using client options

PR #1709, opened 2016-04-13
Add support for fill policies when using opentsdb 2.2

### LINKED ISSUE #1707: [Feature Request] Add support for openTSDB fill policies
Bosun currently doesn't support fill policies from openTSDB 2.2. When used, they return bad "opentsdb: bad query" error.

vitorboschi:
The only problem I've found so far is that bosun cannot handle nan's from openTSDB, so that policy cannot be used yet. Closes #1707


vitorboschi:
ping...

PR #1719, opened 2016-04-22
cmd/scollector: add ability to set global tags by args

ezotrank:


ezotrank:
ping

captncraig:
I'm gonna vote no on this for a few reasons:

- It feels wrong to use flag.Args for this use case. That's a powerful tool that I wouldn't like to reserve at this time for this use case. Everything else uses named flags or else is only available via the config file.

- Global tags can already be set via the config file. If you are using advanced functionality like this, I'm afraid you will have no choice to create a conf file.

- It may be important enough to make a flag alternative for, but I'm leaning towards no on that as well.

Is there a significant reason why setting Tags in the config file is not feasible?

PR #1725, opened 2016-05-06
cmd/bosun: skiplast cmdline switch for development

kylebrandt:


captncraig:
I'm ok with the code. Wondering though...

The last stuff is all to support the host api for opserver. I imagine we are the only people who will use the opserver integration. Would it make sense to have an opt-in flag for that whole side of things?

kylebrandt:
I'm not sure it is only used in the host API? Maybe autocomplete uses it to see the age of things? If people are not indexing data in bosun the load won't take any time. So I'm okay keeping the featured tied to using indexing for now.

PR #1735, opened 2016-05-13
cmd/scollector: fastly status.io monitoring and status.io lib

kylebrandt:


kylebrandt:
Build failure above doesn't have to do with PR, just a timeout when shutting down bosun - so travis thing I think.

cmd/scollector/collectors/fastly.go
```diff
@@ -50,8 +61,89 @@ const (
...
+
+	fastlyStatusPrefix             = "fastly.status."
+	fastlyComponentStatusDesc      = "The current status of the %v. 0: Operational, 1: Degraded Performance, 2: Partial Outage, 3: Major Outage." // see iota for statusio.ComponentStatus
+	fastlyScheduledMaintDesc       = "The number of currently scheduled maintenances. Does not include maintenance that is current active"
+	fastlyActiveScheduledMaintDesc = "The number of currently scheduled maintenances currently in progress. Includes the 'in_progress' and 'verifying'"
+	fastlyActiveIncidentDesc       = "The number of currently active incidents. Includes the 'investingating', 'identified', and 'monitoring' states."
+)
+
+var (
+	fastlyStatusPopRegex = regexp.MustCompile(`(.*)\(([A-Z]{3})\)`)
```
captncraig: a comment describing what a regex is looking for is always useful.

cmd/scollector/collectors/fastly.go
```diff
@@ -50,8 +61,89 @@ const (
...
+	for _, si := range statusio.StatusIndicatorValues {
+		scheduledMaintByImpact[si] = 0
+		activeScheduledMaintByImpact[si] = 0
+	}
+	// Group by scheduled vs inprogress/verifying
+	for _, maint := range summary.ScheduledMaintenances {
+		switch maint.Status {
+		case statusio.Scheduled:
+			scheduledMaintByImpact[maint.Impact]++
+		case statusio.InProgress, statusio.Verifying:
```
captncraig: Is it worth logging in a default case? What happens if they add a new status?

captncraig:
My only issue is the size of the "library" component compared to what we actually use. I'm feeling that one of the following should be true:

1. It is a general purpose status.io library that should be an external repo and vendored in.
2. It is specific to the needs of this collector and should be included in bosun, but only include the bits we use now. 

If this is too pedantic, I'm ok either way.

kylebrandt:
I can vendor it. Make a new repo under bosun-monitor?

cmd/scollector/collectors/fastly.go
kylebrandt: I believe the unmarshal will err since it maps to the enum

captncraig:
sounds good to me

PR #1736, opened 2016-05-13
cmd/scollector: Support collector freq overrides

### LINKED ISSUE #1681: Scollector: Add support to configure polling interval at the collector level.
It's good to have support in some way to override the default/global frequency interval at the individual collector level. We have seen the cases, where we don't want to collect the metrics from all collectors at the same frequency interval.

kylebrandt:
Fixes #1681


kylebrandt:
I haven't actually tested this at all, anyone that wanted this feature willing to give it a whirl?

digrich:
@kylebrandt  Is this available in 0.5.0-rc3? And how to override frequency interval in the configuration file ?

captncraig:
Can you please add docs to https://github.com/bosun-monitor/bosun/blob/master/cmd/scollector/doc.go

Issue #1726, opened 2016-05-07
Ability to get alert vars in notification, and/or specify per-notification templates


angadsingh:
Currently its hardcoded, that notification POST body (HTTP notifications) only has access to the subject of the rendered notification template (or I believe the body if useBody = true). This is too limited, and is causing us to create too many redundant notification definitions.

E.g. one for each pagerduty service (if only the service_key changes, but we copy/paste the entire notification definition:

```
notification x-team-pagerduty {
  post = https://events.pagerduty.com/generic/2010-04-15/create_event.json
  contentType = application/json
  next = x-team-pagerduty
  timeout = 30m
  runOnActions = false
  print = true
  body = `{
      "service_key": "safasdfasfda12412asfs",
      "incident_key": {{.|json}},
      "event_type": "trigger",
      "description": {{.|json}},
      "client": "Bosun",
      "client_url": "http://bosun.corp.company.com:8070"
    }`
}

notification y-team-pagerduty {
  post = https://events.pagerduty.com/generic/2010-04-15/create_event.json
  contentType = application/json
  next = x-team-pagerduty
  timeout = 30m
  runOnActions = false
  print = true
  body = `{
      "service_key": "safasdfasfda12412asfs",
      "incident_key": {{.|json}},
      "event_type": "trigger",
      "description": {{.|json}},
      "client": "Bosun",
      "client_url": "http://bosun.corp.company.com:8070"
    }`
}

#you get the idea
```

It gets worse, for example, when we need to send events to graphite. We need to send the alert name as a tag of the event, so that bosun alerts can be displayed on top of grafana/graphite graphs of their respective components:

```
notification graphite-event-tag-team-x-component-y {
  post = http://hb4101.grid.lhr1.inmobi.com/events/
  contentType = application/json
  next = graphite-event-tag-team-x-component-y
  timeout = 30m
  runOnActions = true
  print = true
  body = `{
      "what": {{.|json}},
      "tags": "bosun,team-x,component-x"
    }`
}

notification graphite-event-tag-team-x-component-z {
  post = http://hb4101.grid.lhr1.inmobi.com/events/
  contentType = application/json
  next = graphite-event-tag-team-x-component-z
  timeout = 30m
  runOnActions = true
  print = true
  body = `{
      "what": {{.|json}},
      "tags": "bosun,team-x,component-z"
    }`
}
```

[...]

PR #1737, opened 2016-05-15
cmd/bosun: alternate templates for POST notifications

mkuzmin:
This is an attempt to support rich Slack notifications with markup and weblinks:
slack

Templates should contain JSON code, but here are the current issues:
- if the code is put into `subject`, Bosun shows ugly incident titles on a web page: https://github.com/bosun-monitor/bosun/issues/1096#issue-90477041.
- `useBody` setting does not help - HTML markup is automatically added to JSON content, and Slack servers do not accept such calls.
- Acknowledgement notifications do not work.

Ideally, we need additional variable resolution in notification settings: #689 & #1726. But this is a complex feature, I found no easy way how to implement it.

Instead, I've added two more settings: 
- `postBody` provide alternate template for common notifications.
- `actionBody` is used for action notifications.

Unfortunately, they have to be configured in different places.

Here is a full example:

```
template lowDatastoreSpace {
    subject = `Datastore free space = {{ bytes (.Eval .Alert.Vars.freeSpace)}}`
    body = `{{ .Graph .Alert.Vars.metric }}`
    postBody = `{ "attachments": [{
        "color": "danger",
        "title": "Incident #{{ .Id }}",
        "title_link": "{{ .Incident }}",
        "text": "Datastore free space = {{ bytes (.Eval .Alert.Vars.freeSpace)}}",
        "fallback": "Incident #{{ .Id }}: Datastore free space = {{ bytes (.Eval .Alert.Vars.freeSpace)}}"
    }]}`
}

notification slack {
    post = https://hooks.slack.com/services/<ID>
    actionBody = `{ "attachments": [{"color": "good", "text": "{{ . }}"}] }`
}
```

If anyone wants to try, here are [binaries](https://github.com/mkuzmin/bosun/releases/tag/0.5.dev-mkuzmin1) and a [Docker image](https://hub.docker.com/r/mkuzmin/bosun/).

### DIFF
--- a/cmd/bosun/conf/conf.go
+++ b/cmd/bosun/conf/conf.go
@@ -329,6 +329,7 @@ type Notification struct {
 	Email        []*mail.Address
 	Post, Get    *url.URL
 	Body         *ttemplate.Template
+	ActionBody   *ttemplate.Template
 	Print        bool
 	Next         *Notification
 	Timeout      time.Duration
@@ -336,10 +337,11 @@ type Notification struct {
 	RunOnActions bool
 	UseBody      bool
 
-	next      string
-	email     string
-	post, get string
-	body      string
+	next       string
+	email      string
+	post, get  string
+	body       string
+	actionbody string
 }
 
 func (n *Notification) MarshalJSON() ([]byte, error) {
@@ -1135,6 +1137,14 @@ func (c *Conf) loadNotification(s *parse.SectionNode) {
 				c.error(err)
 			}
 			n.Body = tmpl
+		case "actionBody":
+			n.actionbody = v
+			tmpl := ttemplate.New(name).Funcs(funcs)
+			_, err := tmpl.Parse(n.actionbody)
+			if err != nil {
+				c.error(err)
+			}
+			n.ActionBody = tmpl
 		case "runOnActions":
 			n.RunOnActions = v == "true"
 		case "useBody":
--- a/cmd/bosun/conf/conf.go
+++ b/cmd/bosun/conf/conf.go
@@ -84,6 +84,7 @@ type Conf struct {
 	node            parse.Node
 	unknownTemplate string
 	bodies          *htemplate.Template
+	postbodies      *ttemplate.Template
 	subjects        *ttemplate.Template
 	squelch         []string
 }
@@ -313,11 +314,12 @@ func (c *Conf) parseNotifications(v string) (map[string]*Notification, error) {
 type Template struct {
 	Text string
 	Vars
-	Name    string
-	Body    *htemplate.Template `json:"-"`
-	Subject *ttemplate.Template `json:"-"`
+	Name     string
+	Body     *htemplate.Template `json:"-"`
+	PostBody *ttemplate.Template `json:"-"`
+	Subject  *ttemplate.Template `json:"-"`
 
-	body, subject string
+	body, postbody, subject string
 }
 
 type Notification struct {
@@ -376,6 +378,7 @@ func New(name, text string) (c *Conf, err error) {
 		Notifications:    make(map[string]*Notification),
 		RawText:          text,
 		bodies:           htemplate.New(name).Funcs(htemplate.FuncMap(defaultFuncs)),
+		postbodies:       ttemplate.New(name).Funcs(defaultFuncs),
 		subjects:         ttemplate.New(name).Funcs(defaultFuncs),
 		Lookups:          make(map[string]*Lookup),
 		Macros:           make(map[string]*Macro),
@@ -825,6 +828,14 @@ func (c *Conf) loadTemplate(s *parse.SectionNode) {
 					c.error(err)
 				}
 				t.Body = tmpl
+			case "postBody":
+				t.postbody = v
+				tmpl := c.postbodies.New(name).Funcs(funcs)
+				_, err := tmpl.Parse(t.postbody)
+				if err != nil {
+					c.error(err)
+				}
+				t.PostBody = tmpl
 			case "subject":
 				t.subject = v
 				tmpl := c.subjects.New(name).Funcs(funcs)
--- a/cmd/bosun/conf/notify.go
+++ b/cmd/bosun/conf/notify.go
@@ -26,12 +26,12 @@ func init() {
 		"The number of email notifications that Bosun failed to send.")
 }
 
-func (n *Notification) Notify(subject, body string, emailsubject, emailbody []byte, c *Conf, ak string, attachments ...*models.Attachment) {
+func (n *Notification) Notify(subject, body string, emailsubject, emailbody []byte, postbody string, c *Conf, ak string, attachments ...*models.Attachment) {
 	if len(n.Email) > 0 {
 		go n.DoEmail(emailsubject, emailbody, c, ak, attachments...)
 	}
 	if n.Post != nil {
-		go n.DoPost(n.GetPayload(subject, body), ak)
+		go n.DoPost(n.GetPayload(subject, body, postbody), ak)
 	}
 	if n.Get != nil {
 		go n.DoGet(ak)
@@ -45,7 +45,10 @@ func (n *Notification) Notify(subject, body string, emailsubject, emailbody []by
 	}
 }
 
-func (n *Notification) GetPayload(subject, body string) (payload []byte) {
+func (n *Notification) GetPayload(subject, body, postbody string) (payload []byte) {
+	if postbody != "" {
+		return []byte(postbody)
+	}
 	if n.UseBody {
 		return []byte(body)
 	} else {
--- a/cmd/bosun/conf/notify.go
+++ b/cmd/bosun/conf/notify.go
@@ -61,7 +61,14 @@ func (n *Notification) DoPrint(payload string) {
 }
 
 func (n *Notification) DoPost(payload []byte, ak string) {
-	if n.Body != nil {
+	if ak == "actionNotification" && n.ActionBody != nil {
+		buf := new(bytes.Buffer)
+		if err := n.ActionBody.Execute(buf, string(payload)); err != nil {
+			slog.Errorln(err)
+			return
+		}
+		payload = buf.Bytes()
+	} else if n.Body != nil {
 		buf := new(bytes.Buffer)
 		if err := n.Body.Execute(buf, string(payload)); err != nil {
 			slog.Errorln(err)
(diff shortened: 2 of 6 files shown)


kylebrandt:
I haven't played with anything outside of email notifications. So I'm probably a bottleneck here :-/ I *think* I like it though.

This continues the model of:

 - Alert Def: When to notify, point to who to notify, point to what that notification should look like
 - Template: What the notification should look like
 - Notification: Who to notify, and how to notify them

So our core issue, as I understand it, is that we want to the notification to look different depending on *how* you are notified. Currently we have mixed in the "What notification should look like" into the notification in order to get HTTP post working, which messed up the above model. Your method fits the model better, with the downsides being:

 - Still kind of messed up since actionBody is in the notification instead of a template - as you pointed out
 - There will be duplication of who+how combinations, because who+how each requires its own notification, but with global variables the repetition there can probably be sane.
 - This would only support one type of HTTP post notification per alert

I think if we go with this it would be better to support N postBodies in a template, and the notification can know what postBody it should reference. Same for the actions.

Thoughts?

kylebrandt:
Another thought that complicates this. Certain JSON endpoints are probably going to need variables in the post line to. i.e. an api like `/notifiy/kyle/serivce/widgetMaker`

mkuzmin:
That's interesting. Initially I requested additional functionality in notifications, but after your explanation I realize templates suit this task better. I've just started adopting Bosun, and there is a lot to learn.

I'll check how to move `actionBody` there.

Also there will be repetition between several templates. I think I can extract a common part into a global setting. Something like this:
```
postBody = `{ "attachments": [{
    "color": "{{if .Action}}good{{else}}danger{{end}}",
    "title": "Incident #{{ .Id }}",
    "title_link": "{{ .Incident }}",
    "text": "{{ .Body }}",
    "fallback": "Incident #{{ .Id }}: {{ .Body }}"
}]}`

template lowDatastoreSpace {
    postBody = `Datastore free space = {{ bytes (.Eval .Alert.Vars.freeSpace)}}"`
}
```

Do you know use cases for multiple parallel POST integrations? I see how to use Slack and email in parallel (like critical to Slack, warnings to email).
But with PagerDuty it's different. Bosun and PagerDuty/OpsGegine/VictorOps overlap in features, and even compete in incident management.
Bosun can generate alerts, and send them to PagerDuty. But that's all - all email and slack notifications are sent by PagerDuty, not Bosun.
I don't use these tools yet, that's just my assumption.

kylebrandt:
Right, the overlap in incident management does make things tricky.

I'm not sure if you aware, but we do support subtemplates. So generally our templates include a header template like so:

```
template  logstash.vrrp.check {
    body = `{{template "header" .}}
    More stuff....`
   ...
```

And then our header:

```
template header {
    body = `
    <style>
    td, th {
        padding-right: 10px;
    }
    </style>
    <p style="font-weight: bold; text-decoration: underline;">
        <a style="padding-right: 10px;" href="{{.Ack}}">Acknowledge</a>
        <a style="padding-right: 10px;" href="{{.Rule}}">View Alert in Bosun's Rule Editor</a>
        {{if .Group.host}}
            <a style="padding-right: 10px;" href="https://status.stackexchange.com/dashboard/node?node={{.Group.host}}">View {{.Group.host}} in Opserver</a>
            <a href="http://kibana.ds.stackexchange.com/app/kibana?#/discover?_g=(refreshInterval:(display:Off,pause:!f,value:0),time:(from:now-15m,mode:quick,to:now))&_a=(columns:!(_source),index:%5Blogstash-%5DYYYY.MM.DD,interval:auto,query:(query_string:(analyze_wildcard:!t,query:'logsource:{{.Group.host}}')),sort:!('@timestamp',desc))">View {{.Group.host}} in Kibana</a>
        {{end}}
    </p>
    <table>
        <tr>
            <td><strong>Key: </strong></td>
            <td>{{printf "%s%s" .Alert.Name  .Group }}</td>
        </tr>
        <tr>
            <td><strong>Incident: </strong></td>
            <td><a href="{{.Incident}}">#{{.Last.IncidentId}}</a></td>
        </tr>
    </table>
    <br/>
    {{if .Alert.Vars.notes}}
        <p><strong>Notes:</strong> {{html .Alert.Vars.notes}}</p>
    {{end}}
    {{if .Alert.Vars.additionalNotes}}
        <p>
        {{if not .Alert.Vars.notes}}
            <strong>Notes:</strong>
        {{end}}
        {{ html .Alert.Vars.additionalNotes }}</p>
    {{end}}
    `
}
```

kylebrandt:
Oh, with email we also inline CSS since email doesn't like style blocks, just attributes. So that does change when it goes out to email.

kylebrandt:
In terms of incident management, I think it might best to start with an either/or sort of logic. Use Bosun's state management, or don't when it comes to specific alerts.

What we miss now I think is the feature to not use it and just use pagerduty instead. In one of these issues (or maybe it was chat) someone was going to work on a notification option that just sends all state changes for the incident. Then people can opt into using another incident management system, and Bosun just sources it.

The reason I'm leaning in this direction is our state machine is already complicated. So when people want features to that don't really fit with the design of it, I imagine more bugs emerging for all the possible states. 

So the *possible* win-win might be to stay opinionated on Bosun's state management, and let people feed bosun state changes to something else if they want to handle alerts differently.

bmandrade:
Hey @mkuzmin,

Do you have this changes working? I tried to use your binaries without success. This is how I configured the template and notifications... https://gist.github.com/bmandrade/757b45cb21ec3983e52532d9456d235f

mkuzmin:
I'm not going to finish this work

kylebrandt:
@mkuzmin I've got a spec in progress in order to improve http post
notifications.

If you have any time for review feedback it would be appreciated

https://github.com/bosun-monitor/bosun/issues/1952

angadsingh:
@kylebrandt there have been many takes on the "post notifications crappy" collection of issues.

one is what I suggested here: https://github.com/bosun-monitor/bosun/issues/1726
another one is: https://github.com/bosun-monitor/bosun/issues/689

to summarize, here are the approaches being discussed:
(1) having .Alert and all the other alert variables accessible in the notification. This would be the death of templates as we know it and doesn't make much sense actually.
(2) having the body (useBody) in the template - not generic enough
(3) the above solution by @mkuzmin - having a single "template type" and just keep adding special fields for each notification type - subject/body for email, "postBody" for POST, etc.

and then there is this:
(4) having different "template types" for different "notification types" - "email template" would have the subject and body fields (body being html'ed and inline css'ed, etc.), "http post template" - having header and payload fields, etc. Then make it possible to specify the template in the "critNotification/warnNotiication" statement as "critNotification = pagerduty-notification(pagerduty-template), slack-notification(slack-template)". bosun would validate if the "type" of template specified matches the "type" of the notification and error otherwise. You could still refer templates within templates to avoid repetition of the same parts.
+
actionBody comes in a very weird place in the abstraction in solution (3) - it is useful for BOTH email and POST notification types - so instead of being a field in the template, it should rather be a different template itself, specified in the alert as "actionTemplate". That way you could have "recoveryTemplate" as well, in the alert, when bosun supports recovery notifications!

IMO (4) is the most generic solution and checks all the abstraction boxes. I guess at the inception of the project, bosun was internally used only with email notifications, and then, when post notifications were added, we just overloaded the use of email template's subject and made it accessible in the POST payload. Now since multiple notifications types are supported (and more may be added in the future), it makes sense to have different types of template definitions for them.

And by the way, there *is a real need* for multiple POST notifications for the same alert - the simplest example is sending events to graphite (so that you can then use them as annotations for alerts on your dashboards!), or the very use case you are denying - notifying BOTH pagerduty and slack. We do not just let pagerduty send messages back to slack - because that sends it in a trim'ed pagerduty style slack message. We like to customize our slack message (using all the rich slack API messaging support) with different fields of the alert rendered differently, etc. Not to mention - being able to add the bosun incident link and links to ack/resolve/silence the alert right in the slack message.

And you're right apart from the above - it should be ok to allow alert variables atleast in the post url of the post notification too.

kylebrandt:
Intended solution is in Google doc linked to on this issue
https://github.com/bosun-monitor/bosun/issues/1952. @captncraig currently
has that on his todo.

On May 24, 2017 09:12, "angadsingh" <notifications@github.com> wrote:

@kylebrandt <https://github.com/kylebrandt> there have been many takes on
the "post notifications crappy" collection of issues.

one is what I suggested here: #1726
<https://github.com/bosun-monitor/bosun/issues/1726>
another one is: #689 <https://github.com/bosun-monitor/bosun/issues/689>

to summarize, here are the approaches being discussed:
(1) having .Alert and all the other alert variables accessible in the
notification. This would be the death of templates as we know it and
doesn't make much sense actually.
(2) having the body (useBody) in the template - not generic enough
(3) the above solution by @mkuzmin <https://github.com/mkuzmin> - having a
single "template type" and just keep adding special fields for each
notification type - subject/body for email, "postBody" for POST, etc.

and then there is this:
(4) having different "template types" for different "notification types" -
"email template" would have the subject and body fields (body being html'ed
and inline css'ed, etc.), "http post template" - having header and payload
fields, etc. Then make it possible to specify the template in the
"critNotification/warnNotiication" statement as "critNotification =
pagerduty-notification(pagerduty-template),
slack-notification(slack-template)".
bosun would validate if the "type" of template specified matches the "type"
of the notification and error otherwise. You could still refer templates
within templates to avoid repetition of the same parts.
+
actionBody comes in a very weird place in the abstraction in solution (3) -
it is useful for BOTH email and POST notification types - so instead of
being a field in the template, it should rather be a different template
itself, specified in the alert as "actionTemplate". That way you could have
"recoveryTemplate" as well, in the alert, when bosun supports recovery
notifications!

IMO (4) is the most generic solution and checks all the abstraction boxes.
I guess at the inception of the project, bosun was internally used only
with email notifications, and then, when post notifications were added, we
just overloaded the use of email template's subject and made it accessible
in the POST payload. Now since multiple notifications types are supported
(and more may be added in the future), it makes sense to have different
types of template definitions for them.

And by the way, there *is a real need* for multiple POST notifications for
the same alert - the simplest example is sending events to graphite (so
that you can then use them as annotations for alerts on your dashboards!),
or the very use case you are denying - notifying BOTH pagerduty and slack.
We do not just let pagerduty send messages back to slack - because that
sends it in a trim'ed pagerduty style slack message. We like to customize
our slack message (using all the rich slack API messaging support) with
different fields of the alert rendered differently, etc. Not to mention -
being able to add the bosun incident link and links to ack/resolve/silence
the alert right in the slack message.

And you're right apart from the above - it should be ok to allow alert
variables atleast in the post url of the post notification too.

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/bosun-monitor/bosun/pull/1737#issuecomment-303719268>,
or mute the thread

PR #1739, opened 2016-05-16
Add support for NTLM authentication. By default NTLM will not be used…

GROpenSourceDev:
…, unless a configuration property UseNtlm is supplied.


kylebrandt:
Mind sharing your use case? Are you putting bosun behind ntlm, or is this for auth against opentsdb or something like that?

GROpenSourceDev:
Hi,
We are using TSDB behind an NTLM-authenticated proxy, so need to extend scollector to support this.
Thanks

kylebrandt:
ntlm should probably live somewhere else and we just vendor it. Does the library live somewhere else already, or did you write it?

GROpenSourceDev:
Hi Kyle

There's no NTLM library available at the moment as far as we know, so we put this together. What are the other options if we can't include it in this way?

Thanks

kylebrandt:
Host it under https://github.com/GROpenSourceDev/ntml, and then import it as "github.com/GROpenSourceDev/ntml". Then run govendor, and it ends up being part of the Bosun repo inside https://github.com/bosun-monitor/bosun/tree/master/vendor , but is then just something that the Bosun repo imports and stores a local copy - this signifies that the basically the "bosun" group doesn't mantain the NTLM library, GROpenSourceDev does.

kylebrandt:
Also, what about https://github.com/ThomsonReutersEikon/go-ntlm ?

kylebrandt:
Also see https://golanglibs.com/top?q=ntlm

GROpenSourceDev:
Hi Kyle,
We have moved the ntlm code out from scollector, created our package, and referenced it via govendor as you suggested.
Thanks for pointing out the ThomsonReuters implementation. We have seen it however their implementation and many others require providing the username and password in plain text. Our implementation differs that it uses default account credentials that the scollector is already running to authenticate by.

Thanks

vendor/github.com/GROpenSourceDev/go-ntlm-auth/ntlm/ntlm_windows.go
```diff
@@ -0,0 +1,304 @@
...
+	"unsafe"
+)
+
+var (
+	secur32_dll           = syscall.NewLazyDLL("secur32.dll")
+	initSecurityInterface = secur32_dll.NewProc("InitSecurityInterfaceW")
+	sec_fn                *SecurityFunctionTable
+)
+
+func init() {
```
captncraig: I would feel more comfortable if this package did not make calls to windows functions unless specifically asked to. If someone is not using ntlm in scollector, I would rather not make this call.

GROpenSourceDev: Hi,
We've added a configuration property "UseNtlm" which controls whether or not to use NTLM authentication. By default this is false so none of this NTLM vendor code will be used.

GROpenSourceDev: Thanks, we've moved the initialisation of the security interface into a function which will be called when first used.

GROpenSourceDev:
Hi, we've made a few changes based on the team's comments. Could we check if there's anything else we need to do to get this approved? Thanks

GROpenSourceDev:
Hi, we've made a few changes based on the team's comments. Could we check if there's anything else we need to do to get this approved?

captncraig:
Not sure what the conflicts are, but if you can resolve them, I'll merge this.

GROpenSourceDev:
@captncraig Thanks, conflicts have been resolved.

PR #1740, opened 2016-05-16
cmd/bosun: add month func to get end of or start of month

kylebrandt:


captncraig:
LGTM. the doc took me a few readthroughs to grok what it does, but the code looks good.

PR #1756, opened 2016-05-26
Adding monitoring for more types of interfaces

krish7919:
- Scollector monitors interfaces whose interface names have a specific prefix.
- Adding tun, wlan and ppp to be monitored too

### DIFF
--- a/cmd/scollector/collectors/ifstat_linux.go
+++ b/cmd/scollector/collectors/ifstat_linux.go
@@ -52,7 +52,8 @@ var netFields = []struct {
 }
 
 var ifstatRE = regexp.MustCompile(`\s+(eth\d+|em\d+_\d+/\d+|em\d+_\d+|em\d+|` +
-	`bond\d+|team\d+|` + `p\d+p\d+_\d+/\d+|p\d+p\d+_\d+|p\d+p\d+):(.*)`)
+	`bond\d+|team\d+|` + `p\d+p\d+_\d+/\d+|p\d+p\d+_\d+|p\d+p\d+|wlan\d+|` +
+	`ppp\d+|tun\d+):(.*)`)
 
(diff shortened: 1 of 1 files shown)


kylebrandt:
https://github.com/bosun-monitor/bosun/pull/1353 also adds things.

What these need is for the interfaces to be categorized into the proper metric. So the scheme should be:

```
 [os|linux].net (physical interfaces)
 [os|linux].net.bond (for teams and bonds)
 [os|linux].net.tunnel (for tunnel interfaces)
 [os|linux].net.virtual (vlans)
 [os|linux].net.other (vlans)
```

See reference at bottom.

This way the aggregation of os.net.bytes should be total traffic and we don't double count things.

image

krish7919:
@kylebrandt : On it. Will send another PR soon.

krish7919:
@kylebrandt : I went through the source this weekend. However, I could not find where to add the metrics aggregation property. Can you please guide?

kylebrandt:
I am not sure what you mean by 'metrics aggregation property'. Can you expand?

krish7919:
1. "interface to be categorized into proper metric" and the scheme. => I renamed it to metics aggregation. And am confused where to edit for taking into account other interfaces.

2. The sample code is for snmp isn't very helpful. Shed some light please?

P.S. I am new to bosun and this is my first PR.
Bosun is awesome; we are just testing it out for now.

kylebrandt:
@krish7919 Sorry for the delay, was focused on getting 0.5.0 out the door. There is no aggregation property, but tsdb, i.e. opentsdb will do aggregation operations on tags for a metric when querying. So if you query say os.net.bytes with no tags specified and using the sum aggregator, all the series will be summed over time.

So the metric layout matches this schema so only i.e. physical interfaces are under os.net.bytes, so when you query it and aggregate you are not double counting bytes that also pass through virtual / tunnel / bond.

Make sense?

PR #1762, opened 2016-06-01
Send throttling

GROpenSourceDev:
We've added the ability to throttle the rate at which SCollector sends data to TSDB. This is configurable by an optional property and by default will work as-is with no throttling.
This is useful if you need to regulate the load on your TSDB instance (or proxy in between).

### DIFF
--- a/collect/queue.go
+++ b/collect/queue.go
@@ -66,6 +66,7 @@ func send() {
 			// If there is no backlog, wait for the minimum send interval before sending any more data
 			if i < BatchSize && time.Since(lastSent) < MinimumSendInterval {
 				qlock.Unlock()
+				time.Sleep(time.Second)
 				continue
 			}
 
--- a/collect/queue.go
+++ b/collect/queue.go
@@ -66,7 +66,8 @@ func send() {
 			// If there is no backlog, wait for the minimum send interval before sending any more data
 			if i < BatchSize && time.Since(lastSent) < MinimumSendInterval {
 				qlock.Unlock()
-				time.Sleep(time.Second)
+				waitTime := lastSent.Add(MinimumSendInterval).Sub(time.Now())
+				time.Sleep(waitTime)
 				continue
 			}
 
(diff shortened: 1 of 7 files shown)


GROpenSourceDev:
@captncraig Fair comment, we've changed it to wait for the min send interval as suggested

collect/queue.go
```diff
@@ -56,9 +56,21 @@ func Flush() {
...
 
 func send() {
+
+	lastSent := time.Now()
+
 	for {
 		qlock.Lock()
 		if i := len(queue); i > 0 {
+
+			// If there is no backlog, wait for the minimum send interval before sending any more data
```
captncraig: I am confused with the logic here. If there is high volume of metrics you are likely to always have a backlog, thus continuing to send at high volume. Am I wrong?

Additionally, it looks like you throw away the last partial batch before sleeping. Should we not sleep only after sending?

GROpenSourceDev: @captncraig 
We are using a larger batch size, expecting to not normally have a backlog, and we want to limit the frequency at which we send these larger batches. In the scenario where there is a backlog, we don't want to limit the sending frequency, to avoid the backlog growing.

What do you mean regarding throwing away a partial batch? We don't intend to discard anything, if you have spotted something then please let us know.

captncraig:
I guess my issue is that it is not really throttling. If I have 10x my batch size ready, it will send 10 batches in a quick burst before sleeping. That could be bad if your backend is struggling to keep up. If you have a large batch, it obviously won't be so much of a problem. It seems ok to me with that in mind, but it also needs documentation in https://github.com/bosun-monitor/bosun/blob/master/cmd/scollector/doc.go. Maybe @gbrayut should take a look and see if there are any pitfalls with this approach.

GROpenSourceDev:
@captncraig @gbrayut We've updated the docs to match. Thanks

GROpenSourceDev:
@captncraig @gbrayut @kylebrandt Hi, Is it possible to have an update on this PR? We're keen to move back onto Master if possible. Thanks

PR #1764, opened 2016-06-01
List open incidents

kylebrandt:


kylebrandt:
Still needs docs.

cmd/bosun/conf/conf.go
```diff
@@ -295,6 +295,31 @@ func (ns *Notifications) Get(c *Conf, tags opentsdb.TagSet) map[string]*Notifica
 	return nots
 }
 
+func GetNotificationChains(c *Conf, n map[string]*Notification) ([][]string) {
```
captncraig: this needs a comments about exactly what it returns. I'm kinda confused on the myself. A list of lists of chains for a single alert? With no duplicates. Maybe writing it out will help.

cmd/bosun/conf/conf.go
```diff
@@ -340,9 +365,9 @@ type Notification struct {
 	body      string
 }
 
-func (n *Notification) MarshalJSON() ([]byte, error) {
-	return nil, fmt.Errorf("conf: cannot json marshal notifications")
-}
+// func (n *Notification) MarshalJSON() ([]byte, error) {
```
captncraig: delete rather than comment.

cmd/bosun/sched/sched.go
```diff
@@ -22,6 +22,8 @@ import (
 	"github.com/MiniProfiler/go/miniprofiler"
 	"github.com/boltdb/bolt"
 	"github.com/bradfitz/slice"
+	"github.com/kylebrandt/boolq"
+	"github.com/kylebrandt/boolq/parse"
```
captncraig: can we alias this reference please. We have way too many things called parse. Better yet, maybe it should just be `boolq.Parse` and nobody else needs to know about the internal parse package.

cmd/bosun/sched/sched.go
```diff
@@ -391,7 +396,18 @@ func (s *Schedule) MarshalGroups(T miniprofiler.Timer, filter string) (*StateGro
 				}
 				continue
 			}
-			if matches(s.Conf, a, v) {
+			if parsedExpr == nil {
```
captncraig: I feel like the logic here is a bit odd. I would like to see the `status[k] = v` bit only once, and continue of it doesn't match the filter. 
Something like `if filter and !filter.Matches(is) { continue}   status[k] = v`. If that makes any sense.

PR #1765, opened 2016-06-02
cmd/scollector: fix rabbitmq backing queue status

kercdbh:


dimamedvedev:
:+1: 
Was that metric useful for you? Just curious.

kercdbh:
It's not yet decided; I am still trying to figure out which metrics I'll be keeping.

PR #1771, opened 2016-06-07
cmd/scollector: Adding .PS1 to supported program extensions and changing command exec…

### LINKED ISSUE #1693: Always allow .ps1 scripts to be run as external collectors
We've had a few reports now of .ps1 scripts not working as external collectors. The [isExecutable code](https://github.com/bosun-monitor/bosun/blob/e412126c63a253f4dfb3a3ed5b3dc79a8e1ba9e3/cmd/scollector/collectors/program.go#L72) used to determine what files are executable on Windows is based on the PATHEXT environmental variable. By default this variable has:

.COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC;.CPL

We should add .PS1 as an executable file extension so that users don't need to update their PATHEXT.

TrentScholl:
This PR explicitly adds .PS1 to the supported executable extensions list.

It also adds code to support execution of PowerShell scripts.

Resolves #1693

### DIFF
--- a/cmd/scollector/collectors/program.go
+++ b/cmd/scollector/collectors/program.go
@@ -73,6 +73,7 @@ func isExecutable(f os.FileInfo) bool {
 	switch runtime.GOOS {
 	case "windows":
 		exts := strings.Split(os.Getenv("PATHEXT"), ";")
+		exts = append(exts, ".PS1")
 		fileExt := filepath.Ext(strings.ToUpper(f.Name()))
 		for _, ext := range exts {
 			if filepath.Ext(strings.ToUpper(ext)) == fileExt {
@@ -115,7 +116,12 @@ func (c *ProgramCollector) Init() {
 var setupExternalCommand = func(cmd *exec.Cmd) {}
 
 func (c *ProgramCollector) runProgram(dpchan chan<- *opentsdb.DataPoint) (progError error) {
-	cmd := exec.Command(c.Path)
+	var cmd *exec.Cmd
+	if runtime.GOOS == "windows" && strings.EqualFold(filepath.Ext(c.Path), ".ps1") {
+		cmd = exec.Command("powershell", "-NoProfile", "-NoLogo", "-NonInteractive", "-ExecutionPolicy", "Bypass", "-File", c.Path)
+	} else {
+		cmd = exec.Command(c.Path)
+	}
 	setupExternalCommand(cmd)
 	pr, pw := io.Pipe()
 	s := bufio.NewScanner(pr)


cmd/scollector/collectors/program.go
```diff
@@ -115,7 +116,12 @@ func (c *ProgramCollector) Init() {
 var setupExternalCommand = func(cmd *exec.Cmd) {}
 
 func (c *ProgramCollector) runProgram(dpchan chan<- *opentsdb.DataPoint) (progError error) {
-	cmd := exec.Command(c.Path)
+	var cmd *exec.Cmd
+	if runtime.GOOS == "windows" && filepath.Ext(c.Path) == ".ps1" {
+		cmd = exec.Command("powershell", "-NoProfile", "-ExecutionPolicy", "Bypass", "-File", c.Path)
```
gbrayut: What do you think about adding -NoLogo and -NonInteractive here? We usually use those in our collector scripts.

I think -ExecutionPolicy Bypass can fail on systems that [set it via GPO](https://absolute-sharepoint.com/2014/03/change-powershell-machinepolicy-execution-policy-windows-server-2012r2.html). Might be best to have this as optional either via an ENV var **SCOLLECTOR_PS_POLICY=byapss** or part of the scollector.toml config file. Or we can wait for someone to report an issue and address it at that time.

gbrayut:
Added a comment. Let me know your thoughts and I can try testing this tomorrow afternoon.

cmd/scollector/collectors/program.go
gbrayut: On second thought... ENV vars don't work well with Windows Services (usually require a reboot to change). How do you feel about adding a PSExecutionPolicy string to https://github.com/bosun-monitor/bosun/blob/master/cmd/scollector/conf/conf.go#L41 and only including it in the command line args if the value is provided? I think that would be the best path forward, but @captncraig might also have an opinion.

gbrayut:
Tell you what... leave it in for now and don't worry about disabling it. Not likely to be used heavily and we can address any issues as they come up

TrentScholl:
Done!

gbrayut:
Thanks I'll test it tomorrow. Also we usually ask to squash, rebase, and force push to the PR branch before we merge. And use cmd/scollector: in the commit message. Check the contributors doc for more details.

TrentScholl:
Done, done and done

Thanks @gbrayut

gbrayut:
Sorry for the delay. I tested this and it works as expected. I used the following sample external collector (will add to docs the next time I update those).

```PowerShell
#Example of a PowerShell external collector. See http://bosun.org/scollector/external-collectors for details
#This file should be saved in C:\Program Files\scollector\collectors\0\mymetrics.ps1 since it is a continuous output script
#scollector.toml should have ColDir = 'C:\Program Files\scollector\collectors'

#Setup format strings and other variables
$epoch = New-Object DateTime (1970,1,1)
$MetricMetadata='{{"metric":"{0}","name":"{1}","value":"{2}"}}'
$MetricData='{{"metric":"{0}","timestamp":{1:F0},"value":{2:G}{3}}}'
$MetricTags=',"tags":{{{0}}}'
$Base="mymetric"

#Send metadata for each metric once on startup (Scollector will resend to bosun periodically)
Write-Output ($MetricMetadata -f "$Base.test","rate","gauge") #See https://godoc.org/bosun.org/metadata#RateType
Write-Output ($MetricMetadata -f "$Base.test","unit","item")  #See https://godoc.org/bosun.org/metadata#Unit
Write-Output ($MetricMetadata -f "$Base.test","desc","A test metric")

#Create tags and send metrics
$tags=$MetricTags -f '"mykey":"myvalue"' #generate static tags here. Can append additional tags in the loop if needed. 
#Use $tags="" to exclude all tags but those added by scollector.
Write-Output ($MetricData -f "$Base.test",[datetime]::UtcNow.Subtract($epoch).TotalSeconds,42.123,$tags)
do {
    $delay = Get-Random -Minimum 5 -Maximum 25
    sleep -Seconds $delay
    Write-Output ($MetricData -f "$Base.test",[datetime]::UtcNow.Subtract($epoch).TotalSeconds,$delay,$tags)
} while ($true)

#If a continuous outpout script ever exits scollector will restart it. If you just want periodic data every 60 seconds you 
#can use a /60/ folder instead of /0/ and allow the script to exit when finished sending a batch of metrics.
```

Thanks for working on this!

gbrayut:
closed via c23f091c158a68242fb0154188145b7a2fab76d1

PR #1781, opened 2016-06-14
Pass IncidentState to notifications for richer templating

gnosek:
This series adds `useFullContext` notification setting. If it's set to true, instead of the template subject/body, the notification template gets passed the whole IncidentState. This allows for richer notifications, e.g.:

```
template common {
    subject = {{.Alert.Vars.title}} on {{.Group.host}}
    body = `<h1><a href="{{.Incident}}">[{{.Last.Status}}] {{.Alert.Vars.title}} on {{.Group.host}}</a></h1>`
}

notification some_webservice {
    post = http://localhost:8000/some_webservice
    useFullContext = true
    body = `{
        "author": "Bosun",
        "thread_id": "{{.AlertKey}}",
        "status": "{{.CurrentStatus}}",
        "subject": "{{.Subject}}",
        "body": {{.Body|json}}
    }`
    contentType = application/json
}
```

### DIFF
--- a/cmd/bosun/conf/conf.go
+++ b/cmd/bosun/conf/conf.go
@@ -352,16 +352,17 @@ type Template struct {
 type Notification struct {
 	Text string
 	Vars
-	Name         string
-	Email        []*mail.Address
-	Post, Get    *url.URL
-	Body         *ttemplate.Template
-	Print        bool
-	Next         *Notification
-	Timeout      time.Duration
-	ContentType  string
-	RunOnActions bool
-	UseBody      bool
+	Name           string
+	Email          []*mail.Address
+	Post, Get      *url.URL
+	Body           *ttemplate.Template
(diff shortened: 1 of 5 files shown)


gnosek:
Not sure I understand what actually failed in the Travis build :(

gnosek:
Oh, it was gofmt, should be fixed now.

gnosek:
gofmt is happy but Travis is not :/

captncraig:
I believe we want to solve this problem a slightly different way, by allowing an template to define an arbitrary set of subtemplates, and the notification then just needs to pick which one, or maybe something like that. It is hard to do in a way that decouples notifications from alerts. Planning to address this for the next release.

kylebrandt:
will be fixed with https://github.com/bosun-monitor/bosun/pull/2135

PR #1782, opened 2016-06-15
cmd/scollector: Record the current_source as a tag rather than a separate metric.

alienth:
Currently, things like ntp.offset are of limited use by themselves. You can't tell which remote the metric is measured against without cross-referencing against ntp.current_source, which seems a bit silly.

This change will allow us to alert things like ntp.offset properly with much more ease, as we can simply filter it using the tag rather than cross referencing. The ntp.current_source metric is currently entirely unused in our config.

:eyeglasses: @kylebrandt  / @captncraig

### DIFF
--- a/cmd/scollector/collectors/ntp_unix.go
+++ b/cmd/scollector/collectors/ntp_unix.go
@@ -70,11 +70,9 @@ func c_ntp_peers_unix() (opentsdb.MultiDataPoint, error) {
 		}
 		remote := fields[0]
 		tags := opentsdb.TagSet{"remote": remote, "refid": fields[1]}
-		var current_source int
 		if fl == "*" {
-			current_source = 1
+			tags = tags.Merge(opentsdb.TagSet{"current_source": "1"})
 		}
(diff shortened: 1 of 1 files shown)


cmd/scollector/collectors/ntp_unix.go
```diff
@@ -70,11 +70,9 @@ func c_ntp_peers_unix() (opentsdb.MultiDataPoint, error) {
 		}
 		remote := fields[0]
 		tags := opentsdb.TagSet{"remote": remote, "refid": fields[1]}
-		var current_source int
 		if fl == "*" {
-			current_source = 1
+			tags = tags.Merge(opentsdb.TagSet{"current_source": "1"})
```
kylebrandt: Conditional tags are generally a bad idea. It will trigger unknowns when you alert on metric{foo=*,current_source=*} when things switch, because that tagset will have "disappeared".

So it would be better to always have the tagkey present and have a value of "true" / "false".

Even with that change, I'm not sure we need an additional metric so would like to see what you are trying to do.

If we do go ahead with make it an additional tag, I would add a new metric most likely so the behavior doesn't change if other people are alerting on the metric.

PR #1784, opened 2016-06-15
cmd/bosun: add timedelta func

kylebrandt:
Needs docs, and a couple funcs_test.go would be nice

image


captncraig:
looks like esc updated. We should all update.

CheRuisiBesares:
Woot thanks!

PR #1786, opened 2016-06-15
cmd/scollector: Added collector for AWS Billing reports

mhenderson-so:
- This includes updating the AWS SDK as there was breaking changes in November
   https://github.com/aws/aws-sdk-go/issues/421

 - Piggybacks on the existing AWS collector and its configuration

 - Also fixed a typo in the existing AWS collector


cmd/scollector/collectors/aws.go
```diff
@@ -76,7 +86,7 @@ func c_aws(accessKey, secretKey, region string) (opentsdb.MultiDataPoint, error)
 	}
 	loadBalancers, err := awsGetLoadBalancers(*elb)
 	if err != nil {
-		slog.Info("No ELB Load Balancecrs found.")
+		slog.Info("No ELB Load Balancers found.")
```
kylebrandt: Warn?

cmd/scollector/collectors/awsBilling.go
```diff
@@ -0,0 +1,445 @@
...
+	//So, billing comes with a bunch of different manifets and whatnot. We don't really need all of that
+	//for this basic billing integration, as the first line of the CSVs we're interested in also has
+	//the breakdown of how it's all structured. We only want to process the last file that is present
+	//for each month, as each month we can get 30+ files. The most recent file for any given month should
+	//contain everything we need.
+	var monthsProcessed []string //The months that have been processed so far
+	var allBills []*billBill
+	for _, billingObject := range thisBucketContents {
+		//If the filename ends in a .gz then it's probably what we want
+		if filepath.Ext(billingObject.awsBillingItem.fileName) == ".gz" {
```
kylebrandt: A lof of nested Ifs, maybe better to assign some of these to bool values, and then just if a && b &&

cmd/scollector/collectors/awsBilling.go
```diff
@@ -0,0 +1,445 @@
...
+				}
+			}
+		}
+	}
+	//Dump into scollector format and clean up
+	for _, thisBill := range allBills {
+		thisBill.toTSDB(&md)
+	}
+	//Clean up after ourselves
+	for _, purge := range purgeObjects {
```
kylebrandt: Does this cleanup need to be done only if all the above doesn't err? If not, maybe defer it, or assign it to a func earlier and then call that func before returning the err

cmd/scollector/collectors/awsBilling.go
```diff
@@ -0,0 +1,445 @@
...
+		slog.Infoln("Need to purge object", *purge.Key, "last modified", purge.LastModified)
```
kylebrandt: I think this info will fill the logs, seems more like a debug like so I woudl remove

cmd/scollector/collectors/awsBilling.go
```diff
@@ -0,0 +1,445 @@
...
+}
+
+//ByLastModified implements sorter for S3 data by last modified date
+type ByLastModified []*s3.Object
+
+func (a ByLastModified) Len() int           { return len(a) }
+func (a ByLastModified) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
+func (a ByLastModified) Less(i, j int) bool { return a[i].LastModified.Before(*a[j].LastModified) }
+
+type billBill struct {
```
kylebrandt: billBill ...?

cmd/scollector/collectors/awsBilling.go
```diff
@@ -0,0 +1,445 @@
...
+		//For our denormalized metrics, just keep the host tag, and the required tag
+		//Some entries do not have a resource ID (e.g. access denied errors, or bucket does not exist, for obvious reasons)
+		tagsByResource := opentsdb.TagSet{"host": tags["host"]}
+		if resid, present := tags["resourceid"]; present {
+			tagsByResource["resourceid"] = resid
+		}
+		tagsByOperation := opentsdb.TagSet{"host": tags["host"], "operation": tags["operation"]}
+		descCosts := fmt.Sprintf("Usage costs for Amazon %v. Datapoints represent costs for a full hour. Data typically lags by 24 hours.", metricCode)
+		descUsage := fmt.Sprintf("Usage volume for Amazon %v. Datapoints represent a full hour of usage. Data typically lags by 24 hours.", metricCode)
+		descUsageRes := fmt.Sprintf("Usage volume for Amazon %v, denormalized with only a resource tag. Datapoints represent a full hour of usage. Data typically lags by 24 hours.", metricCode)
```
kylebrandt: generally these descriptions are declared as consts outside of the checks

cmd/scollector/collectors/awsBilling.go
```diff
@@ -0,0 +1,445 @@
...
+		descUsageOps := fmt.Sprintf("Usage volume for Amazon %v, denormalized with only an operation tag. Datapoints represent a full hour of usage. Data typically lags by 24 hours.", metricCode)
+		//Glacier bills by day, not by hour, for storage, so we need to fill in some fake datapoints
+		if line.ProductCode == "AmazonGlacier" && line.Operation == "Storage" {
+			for i := 0; i < 24; i++ {
+				AddTS(md, fmt.Sprintf("aws.billing.%v.cost", metricCode), line.UsageEndDate.Add(time.Duration(-1*i)*time.Hour).Unix(), line.UnblendedCost/24, tags, metadata.Gauge, metadata.Count, descCosts)
```
kylebrandt: When you have something like line.UsageEndDate.Add(time.Duration(-1*i)*time.Hour).Unix(), don't call it multiple times, save the result in a var and then pass that.

cmd/scollector/collectors/awsBilling.go
```diff
@@ -0,0 +1,445 @@
...
+	b.fetchR53(awsBilling)
+}
+
+//If there is Route53 data for this row, then populate the Route 53 item
+func (b *billLineItem) fetchR53(awsBilling *awsBillingConfig) {
+	if b.ProductCode == "AmazonRoute53" { //Don't do anything if we don't have any R53 info to get
+		zoneID := strings.Split(b.ResourceID, "/")[1]   //The billing ID has a huge resource ID, we only need the last part of it
+		cachedR53Zone, ok := awsBillingR53zones[zoneID] //Check if we have a copy of this zone in our local cache
+		if ok {                                         //If we have a copy of the zone, then use that
+			b.Route53Zone = cachedR53Zone.HostedZone
```
kylebrandt: return, then you you can dedent the else condition

cmd/scollector/collectors/awsBilling.go
```diff
@@ -0,0 +1,445 @@
...
+
+//Download a GZipped item from S3 and return the non-gzipped version of the item
+func (billingObject *bucketContents) downloadGzippedItem(downloader *s3manager.Downloader, awsBilling *awsBillingConfig) (*[]byte, error) {
+	//Get a temporary file to dump this into
+	tempFile, err := ioutil.TempFile("", "scollector-aws-billing-")
+	if err != nil {
+		return nil, err
+	}
+	defer tempFile.Close()
+	_, err = downloader.Download(tempFile,
```
kylebrandt: if _, err := ...; err != nil { return err }

cmd/scollector/collectors/awsBilling.go
```diff
@@ -0,0 +1,445 @@
...
+func (b *billingKeyStructure) parseFromObjectKey(key *string, awsBilling *awsBillingConfig) {
+	keyDir, keyFile := filepath.Split(*key)
+	if keyDir == "" || keyFile == "" {
+		return
+	}
+	dirParts := strings.Split(keyDir, "/")
+	if dirParts[0] != awsBilling.bucketPath {
+		return
+	}
+	dateParts := strings.Split(dirParts[2], "-")
```
kylebrandt: could panic, need to check length of dirParts

cmd/scollector/collectors/awsBilling.go
```diff
@@ -0,0 +1,445 @@
...
+}
+
+//Process the billing file and spit out a bill that contains the bill header and
+//line items
+func (awsBilling *awsBillingConfig) ReadBillingFile(contents *[]byte) *billBill {
+	//Parse the bill's CSV format
+	r := csv.NewReader(strings.NewReader(string(*contents)))
+	records, err := r.ReadAll()
+	if err != nil {
+		log.Fatal(err)
```
kylebrandt: Should be using slog not log, also don't log fatal, that would stop scollector

cmd/scollector/collectors/awsBilling.go
```diff
@@ -0,0 +1,445 @@
...
+		return nil, fmt.Errorf("unable to create S3 downloader")
+	}
+	if purgeDays == 0 {
+		slog.Infof("S3 purging of objects is disabled")
+		awsBilling.purgeOlderThan = time.Date(2999, 12, 31, 23, 59, 59, 0, time.UTC)
+	} else {
+		purgeHours := time.Duration(-1 * 24 * purgeDays)
+		awsBilling.purgeOlderThan = time.Now().Add(purgeHours * time.Hour)
+	}
+	return awsBilling.Go()
```
kylebrandt: maybe .Check() or something, don't really ever see .Go in go code

cmd/scollector/collectors/awsBilling.go
```diff
@@ -0,0 +1,445 @@
...
+	productCodes := []string{}
+	for _, line := range b.LineItems {
+		tags := make(opentsdb.TagSet)
+		//Keep track of the unique product codes we've seen so we can
+		//send metadata for them later
+		if !stringInSlice(line.ProductCode, productCodes) {
+			productCodes = append(productCodes, line.ProductCode)
+		}
+		//Add the host tag, and delete the host tag if it's blank
+		if host, present := tags["host"]; !present {
```
kylebrandt: This is done somewhere in or below Add and AddTS IIRC, so shouldn't need it

cmd/scollector/collectors/awsBilling.go
```diff
@@ -0,0 +1,445 @@
...
+			tagsByResource["resourceid"] = resid
+		}
+		tagsByOperation := opentsdb.TagSet{"host": tags["host"], "operation": tags["operation"]}
+		descCosts := fmt.Sprintf("Usage costs for Amazon %v. Datapoints represent costs for a full hour. Data typically lags by 24 hours.", metricCode)
+		descUsage := fmt.Sprintf("Usage volume for Amazon %v. Datapoints represent a full hour of usage. Data typically lags by 24 hours.", metricCode)
+		descUsageRes := fmt.Sprintf("Usage volume for Amazon %v, denormalized with only a resource tag. Datapoints represent a full hour of usage. Data typically lags by 24 hours.", metricCode)
+		descUsageOps := fmt.Sprintf("Usage volume for Amazon %v, denormalized with only an operation tag. Datapoints represent a full hour of usage. Data typically lags by 24 hours.", metricCode)
+		//Glacier bills by day, not by hour, for storage, so we need to fill in some fake datapoints
+		if line.ProductCode == "AmazonGlacier" && line.Operation == "Storage" {
+			for i := 0; i < 24; i++ {
```
kylebrandt: make 24 a const so not a magic number, or comment what it is 0-23 maybe

cmd/scollector/collectors/awsBilling.go
```diff
@@ -0,0 +1,445 @@
...
+			thisBillLine := &billLineItem{}                                //Somewhere to store our row
+			thisBillLine.parseFromBill(row, cols, awsBilling)              //Parse it
+			thisBill.LineItems = append(thisBill.LineItems, *thisBillLine) //Add it to the bill
+		}
+	}
+	return &thisBill
+}
+
+//Return the a given column's value from a row, given a field
+func val(row []string, cols map[string]int, field string) string {
```
kylebrandt: Are we sure that we could never get here and have that value of cols[strings.ToUpper(field)] be out of the len of row and cause a panic?

cmd/scollector/collectors/awsBilling.go
mhenderson-so: https://www.youtube.com/watch?v=KiuO_Z2_AD4

cmd/scollector/collectors/awsBilling.go
mhenderson-so: Yes, this should only execute if you reach the end of the function without everything catching fire

cmd/scollector/collectors/awsBilling.go
mhenderson-so: Checks properly now, but comment didn't collapse due to this specific line not changing

cmd/scollector/collectors/awsBilling.go
mhenderson-so: Returns now, but comment didn't collapse due to this specific line not changing

cmd/scollector/collectors/awsBilling.go
```diff
@@ -0,0 +1,508 @@
...
+		//Usage Type is meaningless for R53, they're all 'DNS-Queries'
+		if line.ProductCode != "AmazonRoute53" {
+			tags["usagetype"] = line.UsageType
+		}
+		//Use the zone name if present for Route 53 data
+		if line.ProductCode == "AmazonRoute53" && line.Route53Zone != nil && *line.Route53Zone.Name != "" {
+			tags["resourceid"] = strings.ToLower(*line.Route53Zone.Name)
+		} else if line.ResourceID != "" {
+			//Remove pesky invalid characters that show up. Probably should redo this as a regex
+			tags["resourceid"] = strings.ToLower(strings.Replace(strings.Replace(strings.Replace(strings.Replace(line.ResourceID, "/", "-", -1), ":", "-", -1), "[", "", -1), "]", "", -1))
```
kylebrandt: https://godoc.org/bosun.org/opentsdb#TagSet.Clean

PR #1788, opened 2016-06-16
Introduced DisableMetadata configuration property

GROpenSourceDev:
This allows you to externally configure if you would like to disable sending metadata to TSDB. We also wrapped an existing call with the internal flag which appeared to be missing.

### DIFF
--- a/cmd/scollector/conf/conf.go
+++ b/cmd/scollector/conf/conf.go
@@ -19,6 +19,8 @@ type Conf struct {
 	Hostname string
 	// DisableSelf disables sending of scollector self metrics.
 	DisableSelf bool
+	// DisableMetadata prevents scollector sending metadata to tsdbhost
+	DisableMetadata bool
 	// Freq is the default frequency in seconds for most collectors.
 	Freq int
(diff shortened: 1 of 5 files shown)


kylebrandt:
@GROpenSourceDev Side not, not ignoring your PRs. Just pushing towards our next release so want to save this and the other PR for a follow up release. Also am waiting for the person that does most of the windows+scollector work to get back from vacation/conference.

tldr: your patience and contributions are very much appreciated!

GROpenSourceDev:
@kylebrandt No problem, thanks for letting us know.

captncraig:
One thing I'd like to see: If metadata is disabled and never sending, the in-memory collections in the metadata package will still fill up and consume memory. Would be nice if AddMeta in the metadata package would no-op if the flag is set.

GROpenSourceDev:
@captncraig we've made the change as suggested. Thanks

GROpenSourceDev:
@captncraig @kylebrandt Hi, Is it possible to have an update on this PR? We're keen to move back onto Master if possible. Thanks

PR #1789, opened 2016-06-17
timedelta docs and units

CheRuisiBesares:
This adds some basic units and docs for the timedelta expression.

### DIFF
--- a/cmd/bosun/expr/funcs_test.go
+++ b/cmd/bosun/expr/funcs_test.go
@@ -160,3 +160,41 @@ func TestMerge(t *testing.T) {
 		t.Errorf("error expected due to identical groups in merge but did not get one")
 	}
 }
+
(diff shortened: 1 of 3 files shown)


CheRuisiBesares:
So in my use case it would be cool if the final point got diffed against the time now. I think that is the same as `since()` so i think i can get that last point by doing something like

```
series =  q()
merge(timedelta($series),series("foo=bar", epoch(), since($series))
```

I'll have to test that more later.

Anyway Im not sure what is failing about the exit code test.

`go test -run Timedelta` is working for me locally.

CheRuisiBesares:
So just as a note I visually verified the changes to the doc and everything seems alright.

CheRuisiBesares:
@kylebrandt thoughts about how to finalize this?

PR #1793, opened 2016-06-22
docker: update docker build files

gbrayut:
Use go 1.6.2 instead of 1.6
Include tsdbrelay in container. Use TSDBRELAY_OPTS env var to override startup settings
Include /data/scollector.toml file with settings required for ledis backed counters

The latest OpenTSDB version is downloaded at build time from https://github.com/OpenTSDB/opentsdb/releases

This also updates the bosun.conf file to use `tsdbVersion = 2.2` and `ledisDir = /data/ledis_data`

### DIFF
--- a/cmd/bosun/docker/build/Dockerfile
+++ b/cmd/bosun/docker/build/Dockerfile
@@ -24,7 +24,7 @@ RUN mkdir -p /hbase \
     | tar -xzC /hbase \
     && mv /hbase/hbase-$HBASEVER /hbase/hbase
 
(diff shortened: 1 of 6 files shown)


gbrayut:
The first commit is now live at https://hub.docker.com/r/stackexchange/bosun/

The second will have to wait until the next time we update

gbrayut:
Looks like I forgot to merge these to master... Just pushed in ebb2fa49eb2c65287c1e082bb65d3959a612623f

PR #1796, opened 2016-06-23
cmd/bosun: show quiet mode banner in ui

### LINKED ISSUE #1791: Add banner to show when Bosun is in Quiet mode
We need to display something on the webpage that lets us know when we are in quiet mode, since we seem to often forget that it was set.

Maybe even send it as a metric to bosun so we can alert on it like puppet.

kylebrandt:
also add api route to detect quiet mode to support this. Fixes #1791


kylebrandt:
image

captncraig:
LGTM

PR #1798, opened 2016-06-23
Added raid mdadm status collector.

nicollet:
Use IsDigit instead of own function.
More test coverage.

### DIFF
--- a/cmd/scollector/collectors/disk_linux.go
+++ b/cmd/scollector/collectors/disk_linux.go
@@ -1,11 +1,15 @@
 package collectors
 
 import (
+	"bufio"
 	"fmt"
+	"io"
 	"io/ioutil"
+	"os"
 	"regexp"
 	"strconv"
 	"strings"
+	"unicode/utf8"
 
 	"bosun.org/metadata"
 	"bosun.org/opentsdb"
@@ -17,6 +21,9 @@ func init() {
 	collectors = append(collectors, &IntervalCollector{F: c_iostat_linux})
 	collectors = append(collectors, &IntervalCollector{F: c_dfstat_blocks_linux})
 	collectors = append(collectors, &IntervalCollector{F: c_dfstat_inodes_linux})
+	// todo: explain what exactly is interval, looks more like a
+	// timeout to me
+	collectors = append(collectors, &IntervalCollector{F: c_mdadm_linux})
 }
 
 var diskLinuxFields = []struct {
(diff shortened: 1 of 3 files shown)


cmd/scollector/collectors/disk_linux.go
```diff
@@ -156,6 +163,56 @@ func c_iostat_linux() (opentsdb.MultiDataPoint, error) {
...
+		// there is no mdadm: not really an error
+		return opentsdb.MultiDataPoint{}, nil
+	}
+	return mdadmLinux(mdstat)
+}
+
+// mock our file for testing purpose
+func mdadmLinux(mdStat io.Reader) (opentsdb.MultiDataPoint, error) {
+	var md opentsdb.MultiDataPoint
+	var descr = "0: raid is active, 1: bad state (not active)"
```
kylebrandt: convention is to declare descriptions as constants outside funcs. Also they are full sentences.

cmd/scollector/collectors/disk_linux.go
```diff
@@ -156,6 +163,56 @@ func c_iostat_linux() (opentsdb.MultiDataPoint, error) {
...
+		raidFailed := 0 // 0: ok, 1: ko
+		if fields[2] != "active" {
+			raidFailed = 1
+		}
+
+		metric := "linux.disk.mdadm.failed"
+		tags := opentsdb.TagSet{"volume": volumeName}
+
+		// Now let's write some datapoints
+		// FIXME: do we have to write linux.disk. + os.disk.?
```
kylebrandt: For now I think it is fine to keep it in linux.disk, if we ever add software raid monitoring for another OS that would probably be the time to try to normalize it as an additional metric into the os.* namespace

cmd/scollector/collectors/disk_linux_test.go
```diff
@@ -0,0 +1,40 @@
...
+		"",
+		"md",
+		"md123 :",
+		"md125: rebuild raid1 sda2[9]",
+	}
+
+	// all those tests should return some md
+	goodTests := []string{
+		"md125 : rebuilding raid5 sda1[1]",
+		"md123 : active raid1\n" +
```
kylebrandt: can use backticks for multiline strings

cmd/scollector/collectors/disk_linux.go
```diff
@@ -156,6 +163,56 @@ func c_iostat_linux() (opentsdb.MultiDataPoint, error) {
 	return md, err
 }
 
+// Check mdadm raid arrays on linux
+// linux.disk.mdadm.failed value:
+//   0 => active
+//   1 => something else, error
+func c_mdadm_linux() (opentsdb.MultiDataPoint, error) {
```
kylebrandt: Maybe rename, first impression would be that this calls the mdadm command

cmd/scollector/collectors/disk_linux.go
```diff
@@ -156,6 +163,56 @@ func c_iostat_linux() (opentsdb.MultiDataPoint, error) {
...
+// mock our file for testing purpose
+func mdadmLinux(mdStat io.Reader) (opentsdb.MultiDataPoint, error) {
+	var md opentsdb.MultiDataPoint
+	var descr = "0: raid is active, 1: bad state (not active)"
+	scanner := bufio.NewScanner(mdStat)
+	for scanner.Scan() {
+		l := scanner.Text()
+		// md125 : active raid1 sda2[0] sdb2[1]
+		// s:0   1 2      3     disks...
+		if utf8.RuneCountInString(l) <= 2 || l[0:2] != "md" { // l =~ /^md/
```
kylebrandt: Maybe just strings.HasPrefix? But this is probably fine

PR #1802, opened 2016-06-25
Bosun function tail

CheRuisiBesares:
There are a good amount of changes here so I will break them down here.

- Added new `NewSortableSeries` function. This allows you to convert a series into a sortable series but it doesn't do the sorting for you. This is useful if you want to use the comparison operators and hold onto state but you don't care about it being ordered from the onset.

- Added new `Tail` function to builtins. This function uses a very basic quickSelect to find the latest data points and returns just the final k data points. I will be adding additional documentation for this in the next commit. If there are less data points in the series than requested for by tail the series is left untouched.

- New unit test for the `tail` function and the `quickSelect` function are now part of the `funcs_test.go`

- New `quickSelect` function is available for anything implementing `sort.Interface` A few things here. Im not sure if we want to just eat the `log(n)` and just use `sort.Sort()` from the std lib. Im also not sure if we want to move these kinds of helpers out of the `funcs.go` to someplace else and not sure if there is an outside lib we would rather use.

Use cases for tail are inspired by `tail_avg` that you see in the skyline project https://github.com/etsy/skyline/blob/master/src/analyzer/algorithms.py Having the tail function will allow us to use some of the other bosun machinery to do math on a known length set of the most recent data points. Increasing the tail length will make the metrics more resilient to changes while shortening it will make it more sensitive. This will be helpful for situations where metrics are pushed by external systems and you cannot make guarantees about when metrics will arrive and at what density.

### DIFF
--- a/cmd/bosun/expr/funcs.go
+++ b/cmd/bosun/expr/funcs.go
@@ -473,7 +473,7 @@ func Tail(e *State, T miniprofiler.Timer, series *Results, number *Results) (*Re
 		// load up new series with desired
 		// number of points
 		// we already checked len so this is safe
-		for _, item := range sorted[len(sorted)-tailLength : len(sorted)] {
+		for _, item := range sorted[len(sorted)-tailLength:] {
 			newSeries[item.T] = item.V
 		}
 		res.Results = append(res.Results, s)
--- a/cmd/bosun/expr/funcs.go
+++ b/cmd/bosun/expr/funcs.go
@@ -316,7 +316,7 @@ var builtins = map[string]parse.Func{
 		F:      TimeDelta,
 	},
 	"tail": {
-		Args:   []models.FuncType{models.TypeSeriesSet, models.TypeScalar},
+		Args:   []models.FuncType{models.TypeSeriesSet, models.TypeNumberSet},
 		Return: models.TypeSeriesSet,
 		Tags:   tagFirst,
 		F:      Tail,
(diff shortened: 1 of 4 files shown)


cmd/bosun/expr/funcs.go
```diff
@@ -315,6 +316,12 @@ var builtins = map[string]parse.Func{
 		Tags:   tagFirst,
 		F:      TimeDelta,
 	},
+	"tail": {
+		Args:   []models.FuncType{models.TypeSeriesSet, models.TypeScalar},
```
kylebrandt: You might consider a numberSet instead of a scalar and then use something like `match`. This way your argument to tail could the be the result of another reduction function. Want to convert more functions to be like this

cmd/bosun/expr/funcs.go
```diff
@@ -440,6 +447,86 @@ func Filter(e *State, T miniprofiler.Timer, series *Results, number *Results) (*
 	return series, nil
 }
 
+func quickSelect(data sort.Interface, k int, max bool) {
```
kylebrandt: My vote would be to just use sort.Sort() like you mentioned in the comments to keep it simple

cmd/bosun/expr/funcs.go
```diff
@@ -440,6 +447,86 @@ func Filter(e *State, T miniprofiler.Timer, series *Results, number *Results) (*
...
+	}
+	data.Swap(partitionIndex, high)
+	return partitionIndex
+}
+
+func Tail(e *State, T miniprofiler.Timer, series *Results, v float64) (*Results, error) {
+	for _, sr := range series.Results {
+		// if there are fewer points than the requested tail
+		// short circut and just return current series
+		if len(sr.Value.Value().(Series)) > int(v) {
```
kylebrandt: invert condition and continue so most of code is not indented - i.e. https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

kylebrandt:
@captncraig Any opinion on the sort? My vote is to keep the standard sort, but if you think differently I defer to you

cmd/bosun/expr/funcs.go
captncraig: Agree.

captncraig:
Unless we have a convincing benchmark why this is significantly better in a realistic case, I would rather not add our own sort function.

CheRuisiBesares:
Yeah that is totally reasonable. There are probably smarter ways to pick the pivot there anyway and Id feel better going with stdlib. Updating this now. Thanks for feedback

CheRuisiBesares:
@kylebrandt @captncraig knocked out some of those things you requested. This allows you to pass int numberSets now and we are using the built in sort functions for now. All of the cruft is ripped out and tests are updated.

 I think we should add some documentation for that match function as it take a little to read through what that thing is doing. Maybe in another pull later on.

CheRuisiBesares:
Not sure what is up with those failing tests. Happy to make changes but local seems to checkout.

kylebrandt:
@CheRuisiBesares Failing because you need a gofmt funcs.go

CheRuisiBesares:
Ah sorry my go is rusty. Thanks for the help.

PR #1807, opened 2016-06-28
Mssql metrics

GROpenSourceDev:
We've added some extra metrics for the MSSQL collector, relating to memory usage and page life expectancy


gbrayut:
I will merge this in https://github.com/bosun-monitor/bosun/pull/1849 but I changed the metric names and used bytes instead of kbytes to match our other metrics.

Thanks!

PR #1810, opened 2016-07-01
Added raid mdadm status collector.

nicollet:
- Use IsDigit() instead of own function.
 - Get status of each spindle.
 - Added tests that are not linux specific.

### DIFF
--- a/cmd/scollector/collectors/disk.go
+++ b/cmd/scollector/collectors/disk.go
@@ -1,0 +1,141 @@
+package collectors
+
+// things that are not OS specific.
+// ie: can compile and execute anywhere
+
+import (
+	"bufio"
+	"io"
+	"strconv"
+	"strings"
+
+	"bosun.org/metadata"
+	"bosun.org/opentsdb"
+)
+
+// Check mdadm raid arrays on linux
+// linux.disk.mdadm.state values:
+const (
+	mdadmUnknown  = iota
+	mdadmNormal   // 1 => active or clean state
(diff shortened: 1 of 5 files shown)


cmd/scollector/collectors/disk_linux.go
```diff
@@ -17,6 +20,8 @@ func init() {
 	collectors = append(collectors, &IntervalCollector{F: c_iostat_linux})
 	collectors = append(collectors, &IntervalCollector{F: c_dfstat_blocks_linux})
 	collectors = append(collectors, &IntervalCollector{F: c_dfstat_inodes_linux})
+	// todo: increase timer for Mdadm: won't change that often
```
captncraig: todo should be done or removed

captncraig:
looks good to me once you fix the fmt and make sure the collectors are running at the correct interval that you want.

cmd/scollector/collectors/disk.go
```diff
@@ -0,0 +1,141 @@
...
+
+	"bosun.org/metadata"
+	"bosun.org/opentsdb"
+)
+
+// Check mdadm raid arrays on linux
+// linux.disk.mdadm.state values:
+const (
+	mdadmUnknown  = iota
+	mdadmNormal   // 1 => active or clean state
```
tlimoncelli: If these are based on linux.disk.madm values, don't use iota.  iota could change if someone inserts an item in the middle. If this is a copy of constants used by linux.disk.mdadm.state, set them to the constant values.

Issue #705, opened 2015-02-09
warm configure


wkshare:
reread alert configure, and not restart bosun.
Emm~~, This feature would be useful：）

kylebrandt:
Might be done as part of https://github.com/bosun-monitor/bosun/issues/420

PR #1817, opened 2016-07-09
Conf int with reload

kylebrandt:
wip:

```
# kbrandt at grove in ~ on git:master x [16:19:15]
$ curl -XPOST 'http://localhost:8080/api/config/alert' -d "{
    \"Name\" : \"meowNew\", \"AlertText\" : \"alert meowNew {\n\twarn = 0 + 6\n}\", \"Hash\": \"\"
}"
reloaded%
# kbrandt at grove in ~ on git:master x [16:19:12]
$ curl -XDELETE 'http://localhost:8080/api/config/alert/meowNew'
meowNew deleted%
```


kylebrandt:
https://github.com/bosun-monitor/bosun/issues/705

kylebrandt:
example file save hook https://gist.github.com/kylebrandt/c4c9099118bea6e4ab793e6b657a1c2c

kylebrandt:
Example sys conf

```
Hostname = "bosun.example.com"
HTTPListen = ":8080"
TimeAndDate = [ 202, 75, 179, 136 ]
ShortURLKey = "aKey"
CommandHookPath = "/Users/kbrandt/src/hook/hook"
RuleFilePath = "/Users/kbrandt/src/testProdRepo/prod.conf"

[OpenTSDBConf]
    Host = "ny-tsdb01:4242"
    Version = 2.2
    ResponseLimit = 25000000

#Test comment

[ElasticConf]
    Hosts = ["http://ny-lselastic01.example.com:9200", "http://ny-lselastic02.example.com:9200","http://ny-lselastic03.example.com:9200"]

[DBConf]
    RedisHost = "localhost:6389"

[SMTPConf]
    EmailFrom = "bosun@example.com"
    Host = "mail.example.com"

```

cmd/bosun/conf/conf.go
```diff
@@ -1,120 +1,120 @@
...
-	SMTPHost        string // SMTP address: ny-mail:25
-	SMTPUsername    string // SMTP username
-	SMTPPassword    string // SMTP password
-	Ping            bool
-	PingDuration    time.Duration // Duration from now to stop pinging hosts based on time since the host tag was touched
-	EmailFrom       string
-	StateFile       string
-	LedisDir        string
-	LedisBindAddr   string
+type SystemConfProvider interface {
```
captncraig: I think the interface adds complexity here we don't really need. Why not just a struct without Getter Functions? Are we afraid some code will change it without going through the conf package? I'm not really super worried about that.

cmd/bosun/conf/conf.go
```diff
@@ -1,120 +1,120 @@
...
+	GetLogstashContext() expr.LogstashElasticHosts
+	GetElasticContext() expr.ElasticHosts
+	AnnotateEnabled() bool
+
+	MakeLink(string, *url.Values) string
+	EnabledBackends() EnabledBackends
+}
+
+func ValidateSystemConf(sc SystemConfProvider) error {
+	if sc.GetSMTPHost() == "" && sc.GetEmailFrom() == "" {
```
captncraig: I'm not sure about this check. What if someone is only using post notifications?

cmd/bosun/conf/conf.go
```diff
@@ -1,120 +1,120 @@
...
+		return fmt.Errorf("default run every must be greater than 0, is %v", sc.GetDefaultRunEvery())
+	}
+	return nil
+}
 
-	RedisHost     string
-	RedisDb       int
-	RedisPassword string
+type RuleConfProvider interface {
+	RuleConfWriter // Wrong place for now
```
captncraig: Why is this wrong? I think its pretty cool.

cmd/bosun/conf/conf.go
```diff
@@ -1,120 +1,120 @@
...
-	Squelch          Squelches `json:"-"`
-	Quiet            bool
-	SkipLast         bool
-	NoSleep          bool
-	ShortURLKey      string
-	InternetProxy    string
-	MinGroupSize     int
+	GetAlerts() map[string]*Alert
+	GetAlert(string) *Alert
+	SetAlert(string, string) (string, error)
```
captncraig: Set/Delete belong on RuleConfWriter, but you may already know that.

cmd/bosun/conf/conf.go
```diff
@@ -1,120 +1,120 @@
...
-	GraphiteHeaders      []string                  // extra http headers when querying graphite.
-	LogstashElasticHosts expr.LogstashElasticHosts // CSV Elastic Hosts (All part of the same cluster) that stores logstash documents, i.e http://ny-elastic01:9200. Only works with elastc pre-v2, and expects the schema to be logstash's default.
-	ElasticHosts         expr.ElasticHosts         // CSV Elastic Hosts (All part of the same cluster), i.e http://ny-elastic01:9200. Only works with elastic v2+, and unlike logstash it is designed to be able to use various elastic schemas.
-	InfluxConfig         client.Config
+	GetNotifications() map[string]*Notification
+	GetNotification(string) *Notification
 
-	AnnotateElasticHosts []string // CSV of Elastic Hosts, currently the only backend in annotate
-	AnnotateIndex        string   // name of index / table
+	GetMacro(string) *Macro
```
captncraig: No `GetMacros` or `GetNotifications`? Are those only looked at when an alert references them?

cmd/bosun/conf/conf.go
```diff
@@ -1,120 +1,120 @@
...
-			Host:   c.GraphiteHost,
-			Header: headers,
-		}
-	}
-	return graphite.Host(c.GraphiteHost)
+type RuleConfWriter interface {
+	BulkEdit(BulkEditRequest) error
+	GetRawText() string
+	GetHash() string
+	SaveRawText(rawConf, diff, user, message string, args ...string) error
```
captncraig: I'm ok leaving this for now, but the `rawText` abstraction may get broken if/when we add an alternate backend

cmd/bosun/conf/conf.go
```diff
@@ -283,10 +215,14 @@ func (ns *Notifications) Get(c *Conf, tags opentsdb.TagSet) map[string]*Notifica
...
-		ns, err := c.parseNotifications(val)
-		if err != nil {
-			// Should already be checked by conf parser.
-			panic(err)
+		ns := make(map[string]*Notification)
+		for _, s := range strings.Split(val, ",") {
+			s = strings.TrimSpace(s)
+			n := c.GetNotification(s)
+			if n == nil {
+				continue // TODO error here?
```
captncraig: I think this is the correct behavior for now. But we will need to think about what happens if a config provider fails. All of those methods will likely need to return an error in the future.

cmd/bosun/conf/conf.go
```diff
@@ -324,1175 +260,122 @@ func GetNotificationChains(c *Conf, n map[string]*Notification) [][]string {
...
-	case "macro":
-		c.loadMacro(s)
-	case "lookup":
-		c.loadLookup(s)
-	default:
-		c.errorf("unknown section type: %s", s.SectionType.Text)
-	}
+type Macro struct {
+	Text     string
+	Pairs    interface{} // this is BAD TODO
```
captncraig: Is this your TODO?

cmd/bosun/conf/conf.go
```diff
@@ -324,1175 +260,122 @@ func GetNotificationChains(c *Conf, n map[string]*Notification) [][]string {
...
 
-type nodePair struct {
-	node parse.Node
-	key  string
-	val  string
+	*Locator `json:"-"`
 }
 
-type sectionType int
+type LocationType int
```
captncraig: What value does `LocationType` add? Wondering if we just leave it out for now.

cmd/bosun/conf/rule/modify.go
```diff
@@ -0,0 +1,221 @@
+package rule
```
captncraig: I'm a little shaky on calling this package `rule`. Not sure what would be better though. Maybe `alerts`?

captncraig: Or is this really the `file` provider? Not sure.

cmd/bosun/main.go
```diff
@@ -137,19 +143,69 @@ func main() {
...
 			slog.Fatalf("InternetProxy error: %s", err)
 		}
 	}
-	if *flagQuiet {
-		c.Quiet = true
+	cmdHook, err := conf.MakeSaveCommandHook(sysProvider.GetCommandHookPath())
+	if err != nil {
+		slog.Fatal(err)
+	}
+	ruleProvider.SetSaveHook(cmdHook)
```
captncraig: Why does the ruleProvider need to be given the save hook. Can't the reload function below just execute it?

cmd/bosun/main.go
```diff
@@ -137,19 +143,69 @@ func main() {
...
+
+	var reload func() error
+	reloading := make(chan bool, 1)
+	reload = func() error {
+		select {
+		case reloading <- true:
+			fmt.Println("got lock")
```
captncraig: fmt

cmd/bosun/conf/conf.go
kylebrandt: I'm okay with removing the systemconf interface.

cmd/bosun/conf/conf.go
kylebrandt: bad logic, the previous was `||`, but what I mean was "if one, than the other". Will fix

cmd/bosun/conf/conf.go
kylebrandt: GetMacro isn't actually used anywhere. Will remove for now

cmd/bosun/conf/conf.go
kylebrandt: Before it was non exported, nodePair, `Pairs []nodePair`, so this was a shortcut to get it working when I moved the rule stuff to a different package

cmd/bosun/main.go
kylebrandt: The file is saved, and then the hook is called. If the hook fails, the orginal file is restored:

```
if c.saveHook != nil {
		err := c.callSaveHook(c.Name, user, message, args...)
		if err != nil {
			sErr := c.SaveConf(c)
			restore := "successful"
			if sErr != nil {
				restore = sErr.Error()
			}
			return fmt.Errorf("failed to call save hook: %v. Restoring config: %v", err, restore)
		}
	}
	err = c.reload()
```

Also the hook is passed arguments that reload doesn't need to be aware of. But maybe there is a better way?

cmd/bosun/conf/conf.go
```diff
@@ -1,128 +1,139 @@
...
+	GetElasticContext() expr.ElasticHosts
+	AnnotateEnabled() bool
+
+	MakeLink(string, *url.Values) string
+	EnabledBackends() EnabledBackends
+}
+
+// ValidateSystemConf runs sanity checks on the system configuration
+func ValidateSystemConf(sc SystemConfProvider) error {
+	if sc.GetSMTPHost() == "" && sc.GetEmailFrom() != "" || sc.GetSMTPHost() != "" && sc.GetEmailFrom() == "" {
```
captncraig: parens here please. I can never remember priorities.

captncraig: or how about:
```
hasHost := sc.GetSMTPHost() != ""
hasFrom := sc.GetEmailFrom() != ""
if hasFrom != hasHost{ERROR!}
```

cmd/bosun/conf/conf.go
```diff
@@ -324,1175 +286,135 @@ func GetNotificationChains(c *Conf, n map[string]*Notification) [][]string {
...
-		}
-		ns[s] = n
-	}
-	return ns, nil
-}
-
-type Template struct {
-	Text string
-	Vars
+// A Lookup is used to return values based on the tags of a response. It
```
captncraig: all of these comments are really helpful btw. Thanks.

cmd/bosun/conf/rule/modify.go
```diff
@@ -0,0 +1,173 @@
...
+func writeSection(l Location, orginalRaw, newText string) string {
+	var newRawConf bytes.Buffer
+	if l == nil {
+		newRawConf.WriteString(orginalRaw)
+		newRawConf.WriteString("\n")
+		newRawConf.WriteString(newText)
+		newRawConf.WriteString("\n")
+		return newRawConf.String()
+	}
+	newRawConf.WriteString(orginalRaw[:getLocationStart(l)])
```
captncraig: I love the emoji, but these incantations are repeated a few times. How about a helper on the Location type?

`textBefore(originalRaw, l)`
`textAfter(originalRaw, l)`

may not be worth it.

cmd/bosun/conf/system.go
```diff
@@ -0,0 +1,406 @@
...
+	if err := ValidateSystemConf(sc); err != nil {
+		return provider, err
+	}
+	return provider, nil
+}
+
+// NewSystemConf retruns a system conf with default values set
+func newSystemConf() *SystemConf {
+	return &SystemConf{
+		CheckFrequency:  Duration{Duration: time.Minute * 5},
```
captncraig: I still feel like some of these fields should be in the rule conf. Is there a technical reason they are here, or is it a choice? 

`CheckFrequency` `DefaultRunEvery` `MinGroupSize` `PingDuration`(maybe) and `UnknownThreshold` all seem like good candidates for rule config.

cmd/bosun/conf/system.go
```diff
@@ -0,0 +1,406 @@
...
+	}
+	if len(decodeMeta.Undecoded()) > 0 {
+		return sc, fmt.Errorf("undecoded fields in system configuration: %v", decodeMeta.Undecoded())
+	}
+	sc.md = decodeMeta
+	return sc, nil
+}
+
+// LoadSystemConfig is like LoadSystemConfigFile but loads the config from a string
+func LoadSystemConfig(conf string) (*SystemConf, error) {
```
captncraig: Can we reduce duplication here? Maybe LoadFile uses ioutil.ReadFile then calls this method?

cmd/bosun/sched/check.go
```diff
@@ -508,22 +512,42 @@ func (s *Schedule) findUnknownAlerts(now time.Time, alert string) []models.Alert
...
 	for _, ak := range s.findUnknownAlerts(r.Start, a.Name) {
 		r.Events[ak] = &models.Event{Status: models.StUnknown}
 	}
 	var warns, crits models.AlertKeys
-	d, err := s.executeExpr(T, r, a, a.Depends)
+	type res struct {
+		results *expr.Results
+		error   error
+	}
+	rc := make(chan res)
```
captncraig: Trying to reason about orphaned goroutines here. I think this chan should be buffered with size 1.

cmd/bosun/sched/check.go
```diff
@@ -609,13 +634,29 @@ func (s *Schedule) CheckExpr(T miniprofiler.Timer, rh *RunHistory, a *conf.Alert
 		collect.Add("check.errs", opentsdb.TagSet{"metric": a.Name}, 1)
 		slog.Errorln(err)
 	}()
-	results, err := s.executeExpr(T, rh, a, e)
+	type res struct {
+		results *expr.Results
+		error   error
+	}
+	rc := make(chan res)
```
captncraig: buffer of 1

cmd/bosun/sched/check.go
```diff
@@ -482,16 +482,20 @@ var bosunStartupTime = utcNow()
...
 	}
 	if !s.AlertSuccessful(alert) {
 		return keys
 	}
-	a := s.Conf.Alerts[alert]
+	a := s.RuleConf.GetAlert(alert)
 	t := a.Unknown
 	if t == 0 {
-		t = s.Conf.CheckFrequency * 2 * time.Duration(a.RunEvery)
+		runEvery := s.SystemConf.GetDefaultRunEvery()
```
captncraig: I think we should go back to precalculating runevery per alert. This also looks to me like a reason default run every and check freq should be moved to rule conf again.

cmd/bosun/sched/sched.go
```diff
@@ -496,31 +507,33 @@ func marshalTime(t time.Time) string {
...
-	DefaultSched.Close()
+func Close(reload bool) {
+	DefaultSched.Close(reload)
 }
 
-func (s *Schedule) Close() {
-	if s.Conf.SkipLast {
+func (s *Schedule) Close(reload bool) {
+	s.cancelChecks()
+	s.checksRunning.Wait()
```
captncraig: I feel like there should be a timeout here, but I'm not sure what we would do. I don't like the idea of a long running expression being unkillable. Maybe I'm worrying too much?

cmd/bosun/sched/views.go
```diff
@@ -62,9 +62,13 @@ type IncidentSummaryView struct {
 	CritNotificationChains [][]string
 }
 
-func MakeIncidentSummary(c *conf.Conf, s SilenceTester, is *models.IncidentState) IncidentSummaryView {
-	warnNotifications := c.Alerts[is.AlertKey.Name()].WarnNotification.Get(c, is.AlertKey.Group())
-	critNotifications := c.Alerts[is.AlertKey.Name()].CritNotification.Get(c, is.AlertKey.Group())
+func MakeIncidentSummary(c conf.RuleConfProvider, s SilenceTester, is *models.IncidentState) IncidentSummaryView {
+	alert := c.GetAlert(is.AlertKey.Name())
+	if alert == nil {
+		panic("should not be called on nonexistant alert, for now at least TODO?")
```
captncraig: I hate panics. Return an err maybe? Return nil maybe? Is this only called from web requests?

captncraig:
That's all I have for now.

cmd/bosun/conf/system.go
kylebrandt: That was my original thought as well. The reason I put them in system was more of a permissions. This is because editing those will change the way *all* alerts in Bosun will run. So having those in the system conf seemed safer to me.

cmd/bosun/sched/sched.go
kylebrandt: The expressions are cancelled while running. So the only thing that we wait for to complete is the run history.

kylebrandt:
@captncraig Do you see any potential issues with dispatchNotifications which is run from schedule.Run() in alertRunner.go? I'm wonder if on reload (and currently in a shutdown) notifications could potentially be lost and never sent - that would be bad.

branch-notes.md
```diff
@@ -0,0 +1,52 @@
+Thoughts:
```
captncraig: Lets go ahead and delete this at this point.

Issue #1816, opened 2016-07-08
Riak collector fail for riak kv 2.1.4


amgorb:
disabling riak collector with filters help otherwise log gets error like this and send queue grows

2016-07-08_16:37:08.62614 2016/07/08 16:37:08 error: queue.go:105: json: error calling MarshalJSON for type *opentsdb.DataPoint: 
Unparseable number undefined
2016-07-08_16:37:08.62672 2016/07/08 16:37:08 info: queue.go:124: restored 500, sleeping 5s
2016-07-08_16:37:13.63821 2016/07/08 16:37:13 info: queue.go:68: sending: 500, remaining: 6800
2016-07-08_16:37:13.65288 2016/07/08 16:37:13 error: queue.go:105: json: error calling MarshalJSON for type *opentsdb.DataPoint: 
Unparseable number undefined

gbrayut:
fixed in d3bfdea

PR #1822, opened 2016-07-11
fix 1816 issue, riak 2.x returns undefined in json instead of zeroes

amgorb:
bug fix for https://github.com/bosun-monitor/bosun/issues/1816 because riak 2.1.4 returns json with 'undefined' in some fields values instead of zeros


gbrayut:
merged in d3bfdea

Issue #1723, opened 2016-05-02
snmp.go panic: runtime error: index out of range


gbrayut:
Panic in scollector

```
panic: runtime error: index out of range

goroutine 50 [running]:
panic(0xd4df00, 0xc82000a050)
        /usr/local/go/src/runtime/panic.go:464 +0x3e6
bosun.org/snmp.check(0xc8212af2c0, 0xc821048000, 0x0, 0x0)
        /home/teamcity/buildAgent/work/2ae49d7a3bc00259/src/bosun.org/snmp/snmp.go:282 +0x444
bosun.org/snmp.walkN(0xc8201b6000, 0x1, 0x1, 0xc8212c6000, 0x0, 0x0, 0x0, 0x0, 0x0)
        /home/teamcity/buildAgent/work/2ae49d7a3bc00259/src/bosun.org/snmp/walk.go:215 +0x172
bosun.org/snmp.(*Rows).Next(0xc820be4000, 0x9)
        /home/teamcity/buildAgent/work/2ae49d7a3bc00259/src/bosun.org/snmp/walk.go:80 +0x10c
bosun.org/cmd/scollector/collectors.c_snmp_keepalived_vrrp_instances(0x0, 0x0, 0x0, 0x0, 0x0)
        /home/teamcity/buildAgent/work/2ae49d7a3bc00259/src/bosun.org/cmd/scollector/collectors/keepalived_linux.go:71 +0x25b
bosun.org/cmd/scollector/collectors.(*IntervalCollector).Run(0xc8201ab540, 0xc820092960, 0xc8200929c0)
        /home/teamcity/buildAgent/work/2ae49d7a3bc00259/src/bosun.org/cmd/scollector/collectors/interval.go:60 +0x105
created by bosun.org/cmd/scollector/collectors.Run
        /home/teamcity/buildAgent/work/2ae49d7a3bc00259/src/bosun.org/cmd/scollector/collectors/collectors.go:206 +0x141
```

gbrayut:
another similar even on co-tsdb02 a few weeks ago:

```
[root@co-tsdb03 gbray]# cat /var/log/scollector.log-20160529
panic: runtime error: index out of range

goroutine 63 [running]:
panic(0xd709a0, 0xc82000a050)
        /usr/local/go/src/runtime/panic.go:481 +0x3e6
bosun.org/snmp.check(0xc820f1e6f0, 0xc82102fb00, 0x0, 0x0)
[...]

PR #1823, opened 2016-07-11
cmd/scollector: Fix snmp.go panic: runtime error: index out of range

nicollet:
Should prevent scollector to crash.


gbrayut:
Potential fix for  #1723

PR #1826, opened 2016-07-13
cmd/scollector: google_analytics: Limit the number of sources we record.

alienth:
This is to prevent the crazy long tail we have on this metric. It's generating thousands of tags.

:eyeglasses: @kylebrandt / @gbrayut

### DIFF
--- a/cmd/scollector/collectors/google_analytics.go
+++ b/cmd/scollector/collectors/google_analytics.go
@@ -6,6 +6,7 @@ import (
 	"encoding/gob"
 	"fmt"
 	"net/http"
+	"sort"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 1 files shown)


cmd/scollector/collectors/google_analytics.go
```diff
@@ -98,8 +115,18 @@ func getActiveUsersByDimension(md *opentsdb.MultiDataPoint, svc *analytics.Servi
 		if err != nil {
 			return fmt.Errorf("Error parsing GA data: %s", err)
 		}
+		rows[i] = kv{key: key, value: value}
+	}
+	sort.Sort(sort.Reverse(rows))
+	if top != 0 && top < len(rows) {
+		fmt.Println(top, len(rows))
```
captncraig: fmt shouldn't be merged into master

alienth: woops, fixing.

cmd/scollector/collectors/google_analytics.go
```diff
@@ -58,15 +59,19 @@ func c_google_analytics(clientid string, secret string, tokenstr string, sites [
 		return nil, err
 	}
 
-	dimensions := []string{"browser", "trafficType", "source", "deviceCategory", "operatingSystem"}
+	// dimension: max records we want to fetch
+	// "source" has a very long tail so we limit it to something sane
+	// TODO: Dimensions we want and associated attributes should eventually be
+	// setup in configuration.
+	dimensions := map[string]int{"browser": 0, "trafficType": 0, "source": 10, "deviceCategory": 0, "operatingSystem": 0}
```
captncraig: does 0 mean "don't truncate: or "don't include". I'm not sure.

alienth: "Don't truncate". Think -1 would be better?

captncraig: maybe 'top' is not a descriptive name for what the value is here.

alienth: Named 'top' since I'm sorting and selecting the top X values. Recommendations on a better descriptive name?

captncraig: yeah, let's use -1 and call it `maxLength` instead of `top`.

captncraig:
lgtm

PR #1831, opened 2016-07-15
Adding additional Elasticsearch Config options

CheRuisiBesares:
This should give you access to more configuration options for the elastic search datasource. I think its possible that some things are still broken I will need to test some more. There will be no breaking changes to existsing elasticHosts config.

Things that still need to happen:
- docs need to be updated
- syntax hlighting needs to be updated
- annotations endpoint needs to have similar config options

### DIFF
--- a/cmd/bosun/conf/conf.go
+++ b/cmd/bosun/conf/conf.go
@@ -26,6 +26,7 @@ import (
 	"bosun.org/slog"
 	"github.com/MiniProfiler/go/miniprofiler"
 	"github.com/influxdata/influxdb/client"
+	elastic "gopkg.in/olivere/elastic.v3"
 )
 
 type Conf struct {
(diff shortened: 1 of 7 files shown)


CheRuisiBesares:
This is still a WIP there are still some things that need testing and additional work that needs to be done before this can land. I wanted to just get someone to check the direction on this and make sure it looks alright.

To make this work with annotations im thinking about adding a new function to https://github.com/bosun-monitor/annotate/blob/master/backend/backend.go that would have the following signature:
```
func NewCustomElastic(urls []elastic.ClientOptionFuncs, index string) (*Elastic, error) {
```
That would leave the existing api in place but also allow us to do something similar to what I stated above for the elastic configs. Additionally there wouldn't be any shared data structures outside the stuff in the elastic package.

kylebrandt:
Needs to be updated to match new system config format (toml)

unixworld:
@kylebrandt, just a heads up we need this feature urgently, so i will work on the 'toml' config format changes.

PR #1833, opened 2016-07-15
add Php cache collector

jikerseven:
Add a php_cache collector. Check the op_cache and the apcu of the web server(php).
There is a file to put at the root of the webserver. The file is given in plain text in the go file.


jikerseven:
I don't understand what is the problem? @kylebrandt

PR #1834, opened 2016-07-20
cmd/scollector: configurable elasticsearch collector

mathpl:
This adds configuration options for the Elasticsearch collector as well as a timeout on the http requests it does (they can tend to get stuck). As you can see in the doc.go this break previous scollector config since we can configure multiples.

### DIFF
--- a/cmd/scollector/collectors/elasticsearch.go
+++ b/cmd/scollector/collectors/elasticsearch.go
@@ -2,6 +2,7 @@ package collectors
 
 import (
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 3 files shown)


kylebrandt:
Looks like this also breaks auto discovery for instances listening on default ports. Maybe we can keep that while also adding the ability to specify multiple instances?

mathpl:
:+1:, will modify like described tomorrow.

mathpl:
If there's no config it will now pick http://localhost:9200/, preserving default behavior. Note that the ElasticIndexFilters config will still break and needs to be moved in a [[Elasticsearch]] section.

PR #1836, opened 2016-07-26
cmd/bosun: extend toml test, fix SearchSince type

kylebrandt:
wip: still some things to add to example / test.toml


cmd/bosun/bosun.example.toml
```diff
@@ -0,0 +1,85 @@
+# Hostname will be used when links are created in templates (i.e. acknowledge links)
+Hostname = "bosun.example.com"
+
+# The HTTP IP and Port to Listen on. Default is ":8070"
+HTTPListen = ":8080"
```
baweaver: Why 8080 here instead of the default mentioned 8070 right above?

cmd/bosun/bosun.example.toml
```diff
@@ -0,0 +1,85 @@
...
+UnknownThreshold = 5
+
+# This makes it so Bosun ping's and records a metric for every value of the "host" tag it has seen. Default is false
+Ping = true
+
+# How long before hosts stop being pinged if we haven't seen a tagset for that host.Alert. Default is 24 hours
+PingDuration = "24h"
+
+# How long certain items and metrics should be displayed in the UI if we haven't seen them. Default 3 days
+SearchSince = "72h"
```
baweaver: Loading this will cause an error as it expects an integer. Was that intended?

cmd/bosun/bosun.example.toml
```diff
@@ -0,0 +1,85 @@
...
+
+# Enable saving API endpoints and the ability to save the config via the UI. Default is false
+# EnableSave = true
+
+# Path to a command that will be executed on save of the rule configuration. This command is passed a filename, username, message, and vargs
+# If the command does not execute save operations will be canceled and the rule file will be restored
+CommandHookPath = "/Users/kbrandt/src/hook/hook"
```
baweaver: Is there an example hook file or documentation available?

cmd/bosun/bosun.example.toml
kylebrandt: @baweaver the tests also reference this file, so some things I set to non defaults.

cmd/bosun/bosun.example.toml
kylebrandt: https://gist.github.com/kylebrandt/c4c9099118bea6e4ab793e6b657a1c2c

cmd/bosun/bosun.example.toml
kylebrandt: @baweaver the tests also reference this file, so some things I set some non defaults for test.

cmd/bosun/bosun.example.toml
kylebrandt: shouldn't with this branch, if you look I changed the type of that in this branch

cmd/bosun/bosun.example.toml
jiangppp: i clicked this url,but not found

PR #1837, opened 2016-07-27
scollector: update wmi package

gbrayut:
see https://github.com/StackExchange/wmi/pull/16


gbrayut:
Testing on dev systems overnight... will merge to master on wmi repo and here tomorrow

PR #1838, opened 2016-07-27
collect: creating buffer pool for sending datapoints

captncraig:
This is a minor optimization that should reduce gc pressure in the collect package.

Tested with a simple app that sends batches of 300 data points to a local black hole server every 30ms. Monitored with [expvarmon](https://github.com/divan/expvarmon)

super unscientific data:

Before:
screen shot 2016-07-27 at 12 10 22 am

After:
screen shot 2016-07-27 at 12 12 27 am

Notice difference in HeapAlloc and Alloc. gc time is pretty much negligible at this scale, but its always good to throw away less stuff.

### DIFF
--- a/collect/queue.go
+++ b/collect/queue.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
+	"sync"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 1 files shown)


gbrayut:
tested and LGTM

PR #1839, opened 2016-07-27
opentsdb: minor optimization to Replace. No alloc on happy path.

captncraig:
Benchmarks:

Before:
```
BenchmarkReplace_Noop-8        	 1000000	      1865 ns/op	     592 B/op	      51 allocs/op
BenchmarkReplace_Something-8   	 1000000	      1442 ns/op	     336 B/op	      37 allocs/op
```

After:
```
BenchmarkReplace_Noop-8        	 5000000	       327 ns/op	       0 B/op	       0 allocs/op
BenchmarkReplace_Something-8   	 1000000	      1676 ns/op	     336 B/op	      37 allocs/op
```

Could potentially optimize the replacing case, especially since we are almost always reducing the size of the string, but I think this should relieve some of the allocs I saw.

### DIFF
--- a/opentsdb/tsdb.go
+++ b/opentsdb/tsdb.go
@@ -307,6 +307,9 @@ func Clean(s string) (string, error) {
 // tag values and replaces them.
 // See: http://opentsdb.net/docs/build/html/user_guide/writing.html#metrics-and-tags
 func Replace(s, replacement string) (string, error) {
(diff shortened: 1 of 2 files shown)


gbrayut:
Testing looks good so far. Gonna merge to master and let the dev systems run overnight

PR #1841, opened 2016-07-29
docs: Removed row / column from download.md

baweaver:
The page template already takes care of this, and it leaves trailing div tags in plaintext on the current page.

The only other difference was deindenting the remaining code.

### DIFF
--- a/docs/downloads.md
+++ b/docs/downloads.md
@@ -3,35 +3,30 @@ layout: page
 title: Download Bosun
 ---
 
-<div class="row">
-	<div class="col-md-12">
-	
-		{% if site.github != null %}
+{% if site.github != null %}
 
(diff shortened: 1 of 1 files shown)


baweaver:
As an aside and probably off-topic for here: Why are the downloads not at the top above the changelog? Seems counterintuitive to go to a download page and have to scroll to the very bottom to find the downloads.

captncraig:
Thanks. Always open to prettifying things. The download page is kinda weird now. I'm wondering if we should just have that link redirect directly to https://github.com/bosun-monitor/bosun/releases/latest

gbrayut:
I like the page since it also includes the "From Source" details, but agree it could be improved. I could see moving both those sections to the top. From source will always be newer than the binaries since it pulls from master, so maybe source at the top, then Binaries, then the change log?

PR #1845, opened 2016-08-01
Get process uptime and start time from stats file instead of pid dire…

vitorboschi:
Avoid getting process info from its directory modification time, as it can change without the process being restarted. Instead, info is retrieved from `/proc/<pid>/stats` file.

### DIFF
--- a/cmd/scollector/collectors/processes_linux.go
+++ b/cmd/scollector/collectors/processes_linux.go
@@ -62,18 +62,29 @@ func linuxProcMonitor(w *WatchedProc, md *opentsdb.MultiDataPoint) error {
 	var totalCPU int64
 	var totalVirtualMem int64
 	var totalRSSMem int64
-	for pid, id := range w.Processes {
-		file_status, e := os.Stat("/proc/" + pid)
-		if e != nil {
-			w.Remove(pid)
-			continue
+	var systemUptime int64
+	var bootTime int64
+	var uptimeIntRE = regexp.MustCompile(`(\d+)(\.+\d+)?\s(\d+)(\.+\d+)?`)
+	uptime_file, e := ioutil.ReadFile("/proc/uptime")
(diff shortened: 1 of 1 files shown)


cmd/scollector/collectors/processes_linux.go
```diff
@@ -62,18 +62,29 @@ func linuxProcMonitor(w *WatchedProc, md *opentsdb.MultiDataPoint) error {
...
-	for pid, id := range w.Processes {
-		file_status, e := os.Stat("/proc/" + pid)
-		if e != nil {
-			w.Remove(pid)
-			continue
+	var systemUptime int64
+	var bootTime int64
+	var uptimeIntRE = regexp.MustCompile(`(\d+)(\.+\d+)?\s(\d+)(\.+\d+)?`)
+	uptime_file, e := ioutil.ReadFile("/proc/uptime")
+	if e == nil {
```
kylebrandt: why not err?

cmd/scollector/collectors/processes_linux.go
```diff
@@ -62,18 +62,29 @@ func linuxProcMonitor(w *WatchedProc, md *opentsdb.MultiDataPoint) error {
...
 	var totalVirtualMem int64
 	var totalRSSMem int64
-	for pid, id := range w.Processes {
-		file_status, e := os.Stat("/proc/" + pid)
-		if e != nil {
-			w.Remove(pid)
-			continue
+	var systemUptime int64
+	var bootTime int64
+	var uptimeIntRE = regexp.MustCompile(`(\d+)(\.+\d+)?\s(\d+)(\.+\d+)?`)
```
kylebrandt: I don't think this should require a Regex. Just strings.Frields (or something like that). Then parse the float and convert to int, and return the int. This can be a function then since it is also used in procstats_linux.og

PR #1850, opened 2016-08-04
cmd/scollector: skip md volumes if mdadm binary is missing

dimamedvedev:
Rarely there are /dev/md* files but no mdadm installed.

### DIFF
--- a/cmd/scollector/collectors/disk_linux.go
+++ b/cmd/scollector/collectors/disk_linux.go
@@ -169,6 +169,9 @@ func examineMdadmVolume(volumeName string) (volumeDetail, error) {
 	// We don't use --test because it has failed us in the past.
 	// Maybe we should use it sometime in the future
 	output, err := util.Command(tmout, nil, "mdadm", "--detail", volumeName)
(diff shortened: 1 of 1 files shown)


cmd/scollector/collectors/disk_linux.go
```diff
@@ -169,7 +169,10 @@ func examineMdadmVolume(volumeName string) (volumeDetail, error) {
 	// We don't use --test because it has failed us in the past.
 	// Maybe we should use it sometime in the future
 	output, err := util.Command(tmout, nil, "mdadm", "--detail", volumeName)
-	detail := parseExamineMdadm(output)
+	detail := volumeDetail{}
```
captncraig: Just a minor thing. I'm much more used to seeing `if err != nil` blocks. How about:

```
if err != nil{
   return volumeDetail{}, err
}
return parseExamineMdadm(output), nil
```

dimamedvedev:
Thank you for a quick answer, Craig. 
Yeah, your suggestion is way prettier.

PR #1761, opened 2016-06-01
scollector: HP Enterprise Virtual Array (EVA) statistics collector

simnv:
Metrics:
> hp.eva.diskgroup.bytes
> hp.eva.diskgroup.latency
> hp.eva.diskgroup.ops
> hp.eva.diskgroup.queue
> hp.eva.hostconnection.queuedepth
> hp.eva.hostport.bytes
> hp.eva.hostport.latency
> hp.eva.hostport.ops
> hp.eva.hostport.queue
> hp.eva.storagearray.totalkb
> hp.eva.storagearray.totalrequests
> hp.eva.storagecontroller.cpu
> hp.eva.storagecontroller.transfer
> hp.eva.vdisk.bytes
> hp.eva.vdisk.latency
> hp.eva.vdisk.ops

PR #1852, opened 2016-08-05
HP EVA: Fix WMI names

simnv:
Went too far fixing go lint messages, should've fixed that before commit. This patch will make #1761 work correctly.

Also added error messages wrapping.


simnv:
@captncraig Sorry about #1761

PR #1855, opened 2016-08-08
cmd/scollector: mssql.agent.* for monitoring SQL Agent jobs

gbrayut:


gbrayut:
Ready for code review. Testing on ny-devbosun (see http://goo.gl/5GPH57 )

gbrayut:
one of the metrics had a wrong type (WMI said it was a counter but really was a gauge). Testing again overnight and should be ready to merge tomorrow if there are no issues.

cmd/scollector/collectors/sqlagent_windows.go
```diff
@@ -0,0 +1,91 @@
...
+	for _, w := range svc_dst {
+		var dst []Win32_PerfRawData_SQLSERVERAGENT_SQLAgentJobs
+		//Default Instance: Win32_PerfRawData_SQLSERVERAGENT_SQLAgentJobs                  Service: SQLSERVERAGENT
+		//Named Instance:   Win32_PerfRawData_SQLAgentOSCARMAYER_SQLAgentOSCARMAYERJobs    Service: SQLAgent$OSCARMAYER
+		//WMI Class has Alerts, Schedules, and Other instances, but for now we'll just collect the totals
+		q := wmi.CreateQuery(&dst, `WHERE Name = '_Total'`)
+		label := "mssqlserver"
+		if w.Name != `SQLSERVERAGENT` {
+			q = instanceAgentWMIQuery(w.Name, q)
+			label = strings.ToLower(w.Name[8:len(w.Name)])
```
kylebrandt: Do you know that w.Name can not be shorter than 8?

cmd/scollector/collectors/sqlagent_windows.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+type Win32_PerfRawData_SQLSERVERAGENT_SQLAgentJobs struct {
+	Activejobs     uint64
+	Failedjobs     uint64
+	Successfuljobs uint64
+}
+
+func instanceAgentWMIQuery(instancename string, wmiquery string) string {
+	var newname = strings.Replace(strings.Replace(instancename, `$`, "", 1), `_`, "", -1)
```
kylebrandt: https://godoc.org/bosun.org/opentsdb#Clean maybe?

cmd/scollector/collectors/sqlagent_windows.go
gbrayut: yes, it is either SQLSERVERAGENT or SQLAgent$+Instance Name

cmd/scollector/collectors/sqlagent_windows.go
gbrayut: These can't be cleaned here, as they are used to generate the WMI query. They should already be cleaned when they are used in the instance tag.

cmd/scollector/collectors/sqlagent_windows.go
gbrayut: ah does look like there is an off by one error though... should be 9. I'll wrap it in a length check too,  since WMI sometimes goes crazy and returns default values like 0 / ""

PR #1856, opened 2016-08-09
cmd/scollector: bugfix - datapoint validation

mathpl:
- TagSet.Clean() validates cleaned key/value are not empty.
- DataPoint.Clean() is also called when quickly dequeuing the datapoint channel.
- Data points deemed unclean increment scollector.collect.discarded instead of scollector.collect.dropped.

### DIFF
--- a/collect/collect.go
+++ b/collect/collect.go
@@ -50,6 +50,9 @@ var (
 	// Dropped is the number of dropped data points due to a full queue.
 	dropped int64
 
+	// Dropped is the number of discarded data points due to being invalid
+	discarded int64
+
 	// Sent is the number of sent data points.
 	sent int64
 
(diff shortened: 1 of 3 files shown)


collect/collect.go
```diff
@@ -131,6 +135,13 @@ func InitChan(tsdbhost *url.URL, root string, ch chan *opentsdb.DataPoint) error
 		slock.Unlock()
 		return
 	})
+	Set("collect.discarded", Tags, func() (i interface{}) {
+		slock.Lock()
```
captncraig: I don't think you even need to lock if you are using the atomic operations. You could use atomic.LoadInt64 and not need locks at all. I wish we had done all these stats like that from the start.

mathpl: want me to move all those to atomic.* ? I'd be happy to.

captncraig: Yes please. I'm going to merge this for now, but if you wanted to do that in a seperate pr it would be appreciated. Thanks!

PR #1859, opened 2016-08-11
cmd/tsdbrelay: enable version numbers, fix metadata, add error metrics

### LINKED ISSUE #1792: tsdbrelay doesn't support -version flag
We will now be including tsdbrelay in the docker container so it supports ledis backed counters. The tsdbrelay binary does not currently support the -version flag, which it should so we can see what commit it was built from.

Once this is working we will need to update https://github.com/bosun-monitor/bosun/blob/master/cmd/bosun/docker/build/Dockerfile#L35 to include a call to tsdbrelay -version

gbrayut:
Also:
removes runtime.GOMAXPROCS(runtime.NumCPU()) since that is now the default
tabs to spaces in docker.sh
updates docker image to use to go1.7.linux-amd64.tar.gz
fixes missing call to metadata.init
adds metrics for tracking errors from opentsdb/bosun
adds metrics for tracking relayed/errors from secondary relays.

~~This also changes the logic so that data points are always relayed to bosun and secondary relays instead of bailing when the primary opentsdb end point returns an error.~~ Will keep thinking on this and open a separate PR.

Fixes #1792

### DIFF
--- a/cmd/tsdbrelay/main.go
+++ b/cmd/tsdbrelay/main.go
@@ -229,7 +229,8 @@ func (rp *relayProxy) relayPut(responseWriter http.ResponseWriter, r *http.Reque
 				req, err := http.NewRequest(r.Method, relayUrl, body)
 				if err != nil {
 					verbose("relayPutUrls connect error: %v", err)
-					return
+					collect.Add("puts.additional.error", tags, 1)
+					continue
 				}
 				req.Header.Set("Content-Type", "application/json")
 				req.Header.Set("Content-Encoding", "gzip")
@@ -238,7 +239,7 @@ func (rp *relayProxy) relayPut(responseWriter http.ResponseWriter, r *http.Reque
 				if err != nil {
 					verbose("secondary relay error: %v", err)
 					collect.Add("puts.additional.error", tags, 1)
-					return
+					continue
 				}
 				resp.Body.Close()
 				verbose("secondary relay success")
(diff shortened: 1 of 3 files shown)


cmd/tsdbrelay/main.go
```diff
@@ -14,6 +15,8 @@ import (
 	"os"
 	"strings"
 
+	version "bosun.org/_version"
```
gbrayut: @captncraig I'm curious why I had to put an alias here. Locally using go 1.6.3 it can't find the reference to version package and gives an error/removes the import if I try to use a default import or even the _ import. Scollector/Bosun seem to work fine without the alias.

captncraig: I can't explain why. If it helps to remove the alias, go ahead, but it sounds like something odd may be going on with your go environment.

gbrayut:
I'm going to test this on ny-gbraylx01, which uses tsdbrelay to forward to a local docker instance and an influxdb instance and relay to the primary bosun instance.

@captncraig please review when you have time

gbrayut:
Hmm... Metadata is working now but it still doesn't always relay. I'll keep working on this.

gbrayut:
This approach isn't going to work. The reverse proxy generates a 500 response code if there is a connection issue to the primary target and immediately sends that to the client. So even if we still try forwarding the data points to the other endpoints the scollector client will have those same data points queued up to be retried, resulting in duplicate data.

I think there is a simpler way to solve the issue of HA forwarding, so I'm going to bundle up the other commits and push those to master, since they fix a bunch of non forwarding related issues.

gbrayut:
This branch now has all the non-forwarding related changes. I plan on merging it tomorrow (testing on ny-gbraylx01 overnight)

PR #1860, opened 2016-08-11
cmd/silence: allow empty tag sets

### LINKED ISSUE #1650: cmd/silence: Can't set empty tags
Due to:

https://github.com/bosun-monitor/bosun/blob/master/cmd/silence/main.go#L40

```
if *flagTags == "" {
		*flagTags = "host=" + util.Hostname
	}
```

I fixed this problem while in annoate cli by doing the following:

```
	setHostname := false
	for _, arg := range os.Args {
		if arg == "-"+flagHostFlagKey {
			setHostname = true
		}
	}
```

Kind of a bit crude, but works. I should check in `flag` and see if there is a better way to do this.

Also, while in code I should make it so it doesn't take `os/user` since that uses Cgo. Did the following in annotate and it seems to work fine:

```
	un := *userFlag
	if un == "" {
		// Since os/user requires cgo
		un = os.Getenv(USER_ENV)
		sudo := os.Getenv("SUDO_USER")
		if sudo != "" {
			un = sudo
		}
	}
```

gbrayut:
fixes #1650

### DIFF
--- a/cmd/silence/main.go
+++ b/cmd/silence/main.go
@@ -8,7 +8,7 @@ import (
 	"log"
 	"net/http"
 	"os"
-	"os/user"
+	"runtime"
 	"time"
 
 	"bytes"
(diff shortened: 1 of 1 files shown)


gbrayut:
@kylebrandt  can you review when you get a chance

PR #1861, opened 2016-08-11
cmd/scollector: prevent panic if v.Frequency_Object is ever zero

gbrayut:
This can occur in IIS if the service is disabled/removed and can probably happen in other collectors too.


gbrayut:
testing on ny-itweb01 to fix a panic on startup

PR #1866, opened 2016-08-12
cmd/scollector: add a kill switch for total memory used by scollector

gbrayut:
This should help prevent memory leaks in CGO or WMI from causing scollector to consume too much memory.

### DIFF
--- a/cmd/scollector/main.go
+++ b/cmd/scollector/main.go
@@ -245,17 +245,27 @@ func main() {
 		}
 		collect.MaxQueueLen = conf.MaxQueueLen
 	}
-	maxMemMegaBytes := uint64(500)
+	maxMemMB := uint64(500)
 	if conf.MaxMem != 0 {
-		maxMemMegaBytes = conf.MaxMem
+		maxMemMB = conf.MaxMem
 	}
 	go func() {
-		maxMemBytes := maxMemMegaBytes * 1024 * 1024
 		var m runtime.MemStats
 		for range time.Tick(time.Second * 30) {
 			runtime.ReadMemStats(&m)
-			if m.Alloc > maxMemBytes {
-				panic(fmt.Sprintf("memory max reached: (current: %v bytes, max: %v bytes)", m.Alloc, maxMemBytes))
+			allocMB := m.Alloc / 1024 / 1024
+			if allocMB > maxMemMB {
+				slog.Fatalf("memory max runtime reached: (current alloc: %v megabytes, max: %v megabytes)", allocMB, maxMemMB)
+			}
+			//See proccess_windows.go and process_linux.go for total process memory usage.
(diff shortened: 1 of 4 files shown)


cmd/scollector/main.go
```diff
@@ -245,17 +245,27 @@ func main() {
 		}
 		collect.MaxQueueLen = conf.MaxQueueLen
 	}
-	maxMemMegaBytes := uint64(500)
+	maxMemMB := uint64(500)
```
captncraig: I'm not sure I love the default of having the kill switch active. May surprise some people.

cmd/scollector/main.go
```diff
@@ -245,17 +245,27 @@ func main() {
...
 		var m runtime.MemStats
 		for range time.Tick(time.Second * 30) {
 			runtime.ReadMemStats(&m)
-			if m.Alloc > maxMemBytes {
-				panic(fmt.Sprintf("memory max reached: (current: %v bytes, max: %v bytes)", m.Alloc, maxMemBytes))
+			allocMB := m.Alloc / 1024 / 1024
+			if allocMB > maxMemMB {
+				panic(fmt.Sprintf("memory max runtime reached: (current alloc: %v megabytes, max: %v megabytes)", allocMB, maxMemMB))
+			}
+			//See proccess_windows.go and process_linux.go for total process memory usage.
```
captncraig: Is it true that this will only work if they happen to be monitoring the scollector process? Or maybe just any process monitoring enabled at all?

captncraig: Have you tried this with the wmi leaky build and seen it trigger at the appropriate time?

gbrayut: I tested by removing all the defer calls in the WMI package. It took 38 minutes for scollector to panic:

image

captncraig:
very nice. LGTM

cmd/scollector/main.go
gbrayut: It has always been active, but only for runtime memory. This adds a second check for total memory usage.

We were using over 1GB on many systems (around 80GB total on all systems) due to a WMI leak, which could have caused other systems to fail if we hadn't noticed.

gbrayut:
This only works if the process monitoring collectors are enabled, but it matches based on the os.Getpid call, so I don't think you need to be explicitly monitoring scollector.

I'm going to manually deploy this to a few systems including ny-bosun01 (largest scollector memory usage on linux in our environment) to make sure there aren't any issues. Can then merge it on Monday.

gbrayut:
So it appears we don't log panics on Windows so I changed the error from a panic to a fatal. I also added #1867 to find a way to log panics.

gbrayut:
^ was just a rebase and a missing multiplier:
image

Merging to master after travisci builds

PR #1870, opened 2016-08-13
cmd/scollector: change cadvisor to use MachineInfo DiskMap

gbrayut:


gbrayut:
Seems to be working, but needs to be reviewed and discussed so please don't merge yet.

This also implements a knownDiskMapErrors lookup so we don't spam the logs with multiple errors.

mathpl:
Sounds a lot better, I'll give it a go on my side today to compare results and let you know.

gbrayut:
hmm... looks like this approach doesn't work for more than just the loopback devices ( see a few devices failing the lookup today). Another alternative is to have an isRemote flag (or try auto detecting based on the URL containing the current scollector hostname) on the cadvisor config, indicating that it shouldn't try to do any device mapping and just use the major/minor details as the tag.

gbrayut:
I asked our IT guys and they said the device name didn't mater, so assuming there may be other inconsistencies in the future between a local vs remote collector I added an IsRemote flag to the config. Setting it to true will disable the lookup. I also kept the knownDiskMapErrors to prevent any errors from spamming the log.

captncraig:
LGTM

PR #1871, opened 2016-08-16
cmd/scollector: fix empty host= tags not working after PR #1856

gbrayut:
The add function in cmd/scollector/collectors/collectors.go uses host= to indicate that tag should be omitted

See https://github.com/bosun-monitor/bosun/blob/ac7f667fa1e649b877b77bfab7f2a643f5d8e615/cmd/scollector/collectors/collectors.go#L292

This should fix the vsphere errors on ny-bosun01:
```
error: collectors.go:256: Invalid tagset discovered: {disk=CO-EQL,host=}. Skipping datapoint. Added from: vsphere.go:138
error: collectors.go:256: Invalid tagset discovered: {disk=datastore1,host=}. Skipping datapoint. Added from: vsphere.go:105
error: collectors.go:256: Invalid tagset discovered: {disk=datastore1,host=}. Skipping datapoint. Added from: vsphere.go:136
error: collectors.go:256: Invalid tagset discovered: {disk=datastore1,host=}. Skipping datapoint. Added from: vsphere.go:138
```

### DIFF
--- a/cmd/scollector/collectors/collectors.go
+++ b/cmd/scollector/collectors/collectors.go
@@ -241,6 +241,11 @@ func AddTS(md *opentsdb.MultiDataPoint, name string, ts int64, value interface{}
 	}
 
 	tags := t.Copy()
+	if host, present := tags["host"]; !present {
+		tags["host"] = util.Hostname
+	} else if host == "" {
+		delete(tags, "host")
+	}
 	// if tags are not cleanable, log a message and skip it
 	if err := tags.Clean(); err != nil {
 		line := ""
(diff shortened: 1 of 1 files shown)


captncraig:
My only worry is that the invariant for Clean is now violated. Clean should always result in a valid datapoint (can be sent and accepted by opentsdb as is), or an error. A solution more specific to AddTS  might be to move the [host fixing](https://github.com/bosun-monitor/bosun/blob/ac7f667fa1e649b877b77bfab7f2a643f5d8e615/cmd/scollector/collectors/collectors.go#L259) up above the call to Clean.

gbrayut:
Ah... I couldn't find that (bit scatter brained). Let me try that instead.

gbrayut:
Yep... that works and is much better. I'll let this run on ny-bosun01 overnight as well but don't expect any issues

PR #1872, opened 2016-08-16
cmd/bosun: parseFloat in templates

kylebrandt:


kylebrandt:
Same thing can be done with .Eval, so this is not needed.

PR #1877, opened 2016-08-19
Change version.go to 0.6.0 and add runtime date stamp if buildTime is missing

gbrayut:


gbrayut:
I think my git core.autocrlf changed somehow, as it was set to false. That means the govendor files had /r/n, so I redid it from a linux system. I changed autocrlf to input so I should commit unix style line endings now.

kylebrandt:
I think it should be -pre, or we should increment it in every breaking
commit and release whenever.

gbrayut:
If you look at the history we previously didn't apply labels until it entered alpha or RC status. I'm fine with changing version to "0.6.0-pre" but we already append -dev if there isn't a OfficialBuild value, so dev builds would then look like:

> scollector version 0.6.0-pre-dev last modified 2016-08-19T17:02:20-06:00

I don't think we can automate increment the value unless we change the last digit to something dynamic (like a yymmdd stamp), and we definitely won't remember to manually increment it.

gbrayut:
updated to 0.6.0-pre and going to merge to master

PR #1878, opened 2016-08-20
cmd/scollector: Set custom UserAgent and use facebookgo/httpcontrol for 60s RequestTimeout

gbrayut:


gbrayut:
Not ready to merge yet, as I plan on testing on a few systems for a while first.

gbrayut:
This is working now and ready for review. I am going to test it on a few systems to make sure the new 60s timeout doesn't cause any issues.

gbrayut:
I added another commit that will set the UA for the collect package when used in tsdbrelay and bosun. I'll also test these on my VM.

collect/collect.go
```diff
@@ -47,6 +47,9 @@ var (
 	// Whether or not to use NTLM authentication
 	UseNtlm bool = false
 
+	// DefaultClient can be used to override the HTTP client that will be used to make requests.
+	DefaultClient *http.Client
```
captncraig: This is kinda wierd to me. Collect already has an odd client with a ghetto timeout thing (that looks buggy to me). Can we just have a central utility that sets http.DefaultClient?

Something like `util.SetupHttpClient(appName, msg, features...)`?

captncraig: I'd rather collect never had to know anything about http clients, and can just Get and Post to its heart's content.

cmd/scollector/conf/conf.go
```diff
@@ -49,6 +49,11 @@ type Conf struct {
 	// UseNtlm specifies if HTTP requests should authenticate with NTLM.
 	UseNtlm bool
 
+	// UserAgentMessage is an optional message that will be appended to the
+	// end of the User Agent when making HTTP requests.
+	// Example: Scollector/0.6.0-pre (Custom message here)
+	UserAgentMessage string
```
captncraig: This is a doc change. Is the primary use case here putting the hostname? Might want to note why this exists in the docs.

captncraig:
In general, it looks pretty good. I'd just like to reduce some of the duplicated code if possible.

gbrayut:
After speaking with Craig the new commits remove the client code from collect (just use collect.DefaultClient to override) and update the scollector/docs.go file. We decided against a util function, since bosun already requires a custom transport and we may need to add additional custom headers in scollector/tsdbrelay the future.

I'll test on a few systems then merge this to master.

PR #1889, opened 2016-08-25
cmd/tsdbrelay: fix http: proxy error: EOF due to mixing Content-Encoding headers

gbrayut:


gbrayut:
Confirmed this fixes the `http: proxy error: EOF` and `http: proxy error: net/http: request canceled` errors we were seeing. They were caused by collectors sending non-gzip requests to tsdbrelay, which then added the encoding header and forwarded to another tsdbrelay instance, causing an EOF error, which then canceled the request (see the recently added tsdbrelay.*.error metrics for more details). Those datapoints were only being saved in our primary cluster, but now they are flowing to the failover cluster as well.

I'll merge this after testing for a few more hours to make sure there are no other issues.

Issue #1886, opened 2016-08-24
InfluxQL parser is out of date


earwin:
I'm trying to use `moving_average` function in my `influx()` expression and getting a rather cryptic `invalid number of arguments for moving_average, expected 1, got 2` error message.
The very same expression run directly against influx works without any problem.
It then came to my attention that bosun parses influx queries before running them, using a copy of influx code embedded in your repository.
My guess is that parser is out of date?

gbrayut:
I added a PR with the new vendor libraries, but it may take some time before I can do the required testing.

gbrayut:
Thanks to krise3k `SELECT moving_average(usage_idle,6)` queries should now work on master, but moving_average with nested functions like `SELECT moving_average(min(usage_idle),6)` are still having issues with parsing/validation. See https://github.com/bosun-monitor/bosun/pull/1976#issuecomment-265023594 for more details. 

I will wait to close this this until nested functions are working.

PR #1890, opened 2016-08-25
Influxdb: Update vendor libraries

gbrayut:
Potential fix for #1886 

Created using:

```
go get -u github.com/influxdata/influxdb
go get -u github.com/gogo/protobuf/proto #New influxql library requires updated version
govendor update github.com/influxdata/influxdb/client github.com/influxdata/influxdb/influxql github.com/influxdata/influxdb/influxql/internal github.com/influxdata/influxdb/models github.com/influxdata/influxdb/pkg/escape github.com/influxdata/influxdb/pkg/slices github.com/influxdata/influxdb/influxql/neldermead github.com/gogo/protobuf/proto
#TODO: also find a way to check for missing packages. Maybe just delete GOPATH/src/github.com/influxdata/influxdb and try building?
```

This needs to be tested against multiple influxdb versions before it can be merged: 
- [ ] current release v0.12.0
- [ ] previous release v0.9.0 (this was the first version supported by Bosun)
- [ ] next release v1.0.0-rc1

It may take me a while to test all three versions, so if anyone has them already setup and is able to do some testing that would be great.


gbrayut:
```
--- FAIL: TestInfluxQueryDuration (0.00s)
	influx_test.go:48: select * from a WHERE value > 0: 
expected: SELECT * FROM a WHERE value > 0.000 AND time >= '2000-01-01T01:00:00Z' AND time <= '2000-01-01T02:00:00Z' fill(none)
_____got: SELECT * FROM a WHERE value > 0 AND time >= '2000-01-01T01:00:00Z' AND time <= '2000-01-01T02:00:00Z' fill(none)
	influx_test.go:48: select * from a WHERE value > 0: 
expected: SELECT * FROM a WHERE value > 0.000 AND time >= '2000-01-01T01:00:00Z' AND time <= '2000-01-01T02:00:00Z' GROUP BY time(15m) fill(none)
_____got: SELECT * FROM a WHERE value > 0 AND time >= '2000-01-01T01:00:00Z' AND time <= '2000-01-01T02:00:00Z' GROUP BY time(15m) fill(none)
```
I'll have to dig into why the value is different...

earwin:
I don't know, if it's worth mentioning here or no, but influx client you're using is deprecated, as per https://github.com/influxdata/influxdb/blob/master/client/README.md
They now have an incompatible v2 client.

gbrayut:
yeah saw that. It looks like it isn't backwards compatible though. This code was written by someone at InfluxDB, so I hope it will work with just a vendor update, but we'll see. I won't have time to dig for a few weeks probably, so if it needs a rewrite it could take a while.

PR #1893, opened 2016-09-02
cmd/bosun: make graph/expr more keyboard-tab friendly

kylebrandt:
also make it so shift-enter executes on the graph page on text input fields (similar to expr page)


captncraig:
It looks like go generate is out of date. Or maybe there is a js file you didn't commit?

kylebrandt:
ya will fix

cmd/bosun/web/static/js/graph.ts
```diff
@@ -466,7 +469,12 @@ bosunControllers.controller('GraphCtrl', ['$scope', '$http', '$location', '$rout
 		});
 		return request;
 	}
-	$scope.Query = function() {
+		$scope.keydown = function ($event: any) {
```
captncraig: I'm curious why only certain fields attach this event, and others do not. Would it work to attach the keydown event to the entire view and handle it no matter where they are focused?

kylebrandt: If you are over a link and hit shift-enter it opens the link a new tab. But maybe it could globally be applied to text input fields for the view if you can think of a way to do that.

captncraig:
Ok then. Go ahead when the build passes

PR #1900, opened 2016-09-07
cmd/scollector: Update SNMP timeout from 5s to 30s and allow override

gbrayut:


gbrayut:
Going to test on co-tsdb01 to see if it reduces our snmp errors

Issue #1862, opened 2016-08-11
scollector generates wrong key without FQDN


R-Sommer:
Hi,

I run a vsphere environment without DNS. This leads scollector to generate an incomplete key i.e. vsphere.guest.mem.used,guest=vsa1,host=192 (I'd assume the idea behind is to cut the domain part from FQDN).

gbrayut:
This is due to the clean function in the [bosun.org/util](https://github.com/bosun-monitor/bosun/blob/master/util/util.go#L31) package. You can add `FullHost = true` to disable the short names and always use FQDNs.

@kylebrandt  thoughts on adding an IP check to the util function? probably shouldn't string split if it is a valid IPv4 address.

R-Sommer:
@gbrayut I added  `FullHost = true` and now I have complete ip addresses. However, there is a new error:

error: collectors.go:256: Invalid tagset discovered: {disk=ESXi1-BackupDS,host=}. Skipping datapoint. Added from: vsphere.go:104

Do you want me to keep this issue open regarding IP check? Shall I open a new issue for this new error?

PR #1902, opened 2016-09-07
util: don't try to clean IP addresses. Also don't clean twice on Set

gbrayut:


gbrayut:
Addresses part of https://github.com/bosun-monitor/bosun/issues/1862

Tested locally and it works using ipv4 address as hostname. IPv6 addresses don't use .s so they won't have the same split issue.

captncraig:
lgtm

PR #1904, opened 2016-09-12
Implementing https listener. 

captncraig:
Https and http may both be specified. Removing RelayListener.

### DIFF
--- a/cmd/bosun/conf/conf.go
+++ b/cmd/bosun/conf/conf.go
@@ -30,7 +30,9 @@ import (
 // outside of the package without a setter
 type SystemConfProvider interface {
 	GetHTTPListen() string
-	GetRelayListen() string
+	GetHTTPSListen() string
+	GetTLSCertFile() string
+	GetTLSKeyFile() string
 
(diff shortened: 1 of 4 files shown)


cmd/bosun/conf/system.go
```diff
@@ -17,8 +17,11 @@ import (
 // SystemConf contains all the information that bosun needs to run. Outside of the conf package
 // usage should be through conf.SystemConfProvider
 type SystemConf struct {
-	HTTPListen    string
-	RelayListen   string
+	HTTPListen  string
```
kylebrandt: Can you add these to the example toml file?

cmd/bosun/conf/conf.go
```diff
@@ -30,7 +30,9 @@ import (
 // outside of the package without a setter
 type SystemConfProvider interface {
 	GetHTTPListen() string
-	GetRelayListen() string
+	GetHTTPSListen() string
+	GetTLSCertFile() string
```
kylebrandt: can we combine these into a pem?

PR #1906, opened 2016-09-14
cmd/bosun: POC for table type with leftjoin func

kylebrandt:
Mainly meant to be used for Grafana, possible WIP... or maybe fine as is... would at least need tests and docs

image

### DIFF
--- a/cmd/bosun/expr/funcs.go
+++ b/cmd/bosun/expr/funcs.go
@@ -310,6 +310,14 @@ var builtins = map[string]parse.Func{
 		Tags:   tagFirst,
 		F:      Shift,
 	},
+	"leftjoin": {
+		Args:     []models.FuncType{models.TypeString, models.TypeString, models.TypeNumberSet},
+		VArgs:    true,
+		VArgsPos: 2,
+		Return:   models.TypeTable,
+		Tags:     nil, // TODO
+		F:        LeftJoin,
+	},
 	"merge": {
 		Args:   []models.FuncType{models.TypeSeriesSet},
 		VArgs:  true,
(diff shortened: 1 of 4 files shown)


cmd/bosun/expr/funcs.go
```diff
@@ -595,6 +603,56 @@ func Merge(e *State, T miniprofiler.Timer, series ...*Results) (*Results, error)
...
+func LeftJoin(e *State, T miniprofiler.Timer, keysCSV, columnsCSV string, rowData ...*Results) (*Results, error) {
+	res := &Results{}
+	dataWidth := len(rowData)
+	if dataWidth == 0 {
+		return res, fmt.Errorf("leftjoin requires at least one item to populate rows")
+	}
+	keyColumns := strings.Split(keysCSV, ",")
+	dataColumns := strings.Split(columnsCSV, ",")
+	keyWidth := len(keyColumns)
+	keyIndex := make(map[string]int)
```
captncraig: if gonna use make, may as well insert the size here.

cmd/bosun/expr/funcs.go
```diff
@@ -595,6 +603,56 @@ func Merge(e *State, T miniprofiler.Timer, series ...*Results) (*Results, error)
...
+	for i, v := range keyColumns {
+		keyIndex[v] = i
+	}
+	if len(dataColumns) != dataWidth {
```
captncraig: can do this check even sooner. fail fast.

cmd/bosun/expr/funcs.go
```diff
@@ -595,6 +603,56 @@ func Merge(e *State, T miniprofiler.Timer, series ...*Results) (*Results, error)
...
+		return res, fmt.Errorf("mismatch in length of data rows and data labels")
+	}
+	t := Table{}
+	t.Columns = append(keyColumns, dataColumns...)
+	rowWidth := len(dataColumns) + len(keyColumns)
+	rowGroups := []opentsdb.TagSet{}
+	for i, r := range rowData {
+		if i == 0 {
```
captncraig: if I understand, the first dataset becomes the "authority" on which tags exist, and others join based on the key tag(s), adding their value to the appropriate column?

cmd/bosun/expr/funcs.go
```diff
@@ -595,6 +603,56 @@ func Merge(e *State, T miniprofiler.Timer, series ...*Results) (*Results, error)
...
+			for _, val := range r.Results {
+				row := make([]interface{}, rowWidth)
```
captncraig: is interface{} appropriate? Would `Value` work a bit more specifically, or does this include the string keys too?

cmd/bosun/expr/funcs.go
```diff
@@ -595,6 +603,56 @@ func Merge(e *State, T miniprofiler.Timer, series ...*Results) (*Results, error)
...
+				row[keyWidth+i] = val.Value.Value()
+				rowGroups = append(rowGroups, val.Group)
+				t.Rows = append(t.Rows, row)
+			}
+			continue
+		}
+		for rowIndex, group := range rowGroups {
+			for _, val := range r.Results {
+				if group.Subset(val.Group) {
+					t.Rows[rowIndex][keyWidth+i] = val.Value.Value()
```
captncraig: what happens if multiple items in the second set match a key from the first set? We will throw data away. Perhaps a warning if they have not properly reduced/aggregated?

captncraig:
I'm a little off on the name `leftJoin` too. It is somewhat technically correct, but a bit misleading. That may be more appropriate for something else in the future. How about `table`, `toTable` or `buildTable`?

cmd/bosun/expr/funcs.go
kylebrandt: will fix

cmd/bosun/expr/funcs.go
kylebrandt: will fix

cmd/bosun/expr/funcs.go
kylebrandt: Yes, that is why I called it left join, everything is joined to the left column.

cmd/bosun/expr/funcs.go
kylebrandt: There are string values since some of the columns are tags. If there is more complicated use of tables in the future I could add another field to the table type that specifies the type of each column

kylebrandt:
@captncraig regarding the name, I imagine there might be other things that create tables. We also might want to create tables with other style of joins.

This also behaves similarly to the leftjoin template function so it continues a pattern.

cmd/bosun/expr/funcs.go
kylebrandt: Since it is a left join, that can't happen as the column won't be joined. The requirement is that The columns are a subset of the first. So if your first column just has host, and the second has host,drive - it won't be joined because host,drive is not a subset of host. So if you do this you just end up with a column that has no values in it.

PR #1915, opened 2016-09-23
cmd/bosun: update docker container to 0.6.0-pre version

gbrayut:
This uses the new TOML configuration format and allows the rules to be reloaded from the UI.
Also update HBASE to latest release and set explicit version for OpenTSDB git clone

### DIFF
--- a/cmd/bosun/docker/build/Dockerfile
+++ b/cmd/bosun/docker/build/Dockerfile
@@ -11,11 +11,11 @@ RUN apt-get update && apt-get install -y \
     && rm -rf /var/lib/apt/lists/*
 
 ENV TSDB /tsdb
-RUN git clone --single-branch --branch v2.2.0 --depth 1 git://github.com/OpenTSDB/opentsdb.git $TSDB && \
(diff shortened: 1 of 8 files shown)


gbrayut:
testing on http://ny-gbraylx01.ds.stackexchange.com:8070/ with no issues so far... will try and merge early next week.

captncraig:
Is the intention ever that a user of this would mount their own configs as a volume, or just edit it in-app? Perhaps the readme on docker hub should be updated with some of these changes as well? 

Looks pretty good to me, although long term I would like to see a simpler bosun image with docker-compose based backends and such.

gbrayut:
Since this is designed for PoC and testing I'm less worried about that, but I think right now it should allow that so long as you replace all the files in that folder.

Future images should be more granular and designed to support that kind of setup. I plan on creating a spec for those at some point, but not sure if it will happen this month.

gbrayut:
Pushed to Docker hub as stackexchange/bosun:0.6.0-pre with version:

bosun version 0.6.0-pre-dev (49193375a30655a74df7d74f2b9e53a40f73694e) built 2016-11-08T22:24:51Z

PR #1919, opened 2016-09-27
Feature/add close on normal

ewales:


captncraig:
close this cause it is not ready, or some other reason? I'd be happy to review it if you think it is ready.

PR #1922, opened 2016-10-04
Added es SimpleClient (standalone es instance) support.

unixworld:
This change will fix connection issue("no Elasticsearch node available") with standalone elasticsearch server.  

For more deatils refer https://github.com/olivere/elastic/wiki/Client

### DIFF
--- a/cmd/scollector/collectors/dns_windows.go
+++ b/cmd/scollector/collectors/dns_windows.go
@@ -40,9 +40,9 @@ func c_dns_windows() (opentsdb.MultiDataPoint, error) {
 				}
 			} else { //Unique stat, we can put it straight away
 				if "" == p.Grouping {
-					Add(&md, "dns."+p.Category+"."+p.Metric, v.Value, opentsdb.TagSet{}, p.RateType, metadata.Count, p.Description)
+					Add(&md, "dns."+p.Category+"."+p.Metric, *v.Value, opentsdb.TagSet{}, p.RateType, metadata.Count, p.Description)
 				} else {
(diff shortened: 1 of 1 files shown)


unixworld:
Test is failing due to missing env var "SETSTATUS_TOKEN"

captncraig:
The build failure is actually that `cmd/bosun/conf/system_test.go` needs to have gofmt run on it.

unixworld:
@captncraig Thanks for the quick help as mentioned after running gofmt on cmd/bosun/conf/system_test.go, the test passed.

unixworld:
@captncraig we had the same issue with annotations backend (https://github.com/bosun-monitor/annotate/blob/master/backend/backend.go#L175) "2016/10/17 16:37:11 warning: web.go:159: could not initalize annotate backend, will try again: no Elasticsearch node available".  So do i need to change the annotations code as well to support SimpleClient?

unixworld:
@kylebrandt, as per you comment in slack do you want me to re-work on https://github.com/bosun-monitor/bosun/pull/1831 or this is expectable we just need  ES simple client support for both bosun backend and annotations. Please let me know so that I will make the changes accordingly.

kylebrandt:
Why close? Sorry was thinking about this but also dealing with the Monday morning work.

I think this might be good, just want to make sure we could have simple client as an option, or a more fine grained option as well. So as long as this doesn't block the more complex version I think this good. Solves most people's problem.

unixworld:
Opps sorry ...putting it back also working on adding support for es option

Issue #1926, opened 2016-10-07
Alerts using multiple Elastic indexes with the same timespan will have cache conflicts and return invalid results


gbrayut:
Currently it looks like the bosun cache doesn't account for which index is being used in an escount query, only the timespan, bucket size, and the field name. You can run an escount(esdaily("CreationDate", "fastly-", "2006.01.02"), ....) query on the expression page and then change the index to "trafficlogs-" or even "thisindexdefinitelydoesntexist" and you still get the same results from the previously cached query.

This causes issues when you try to use multiple indexes in an alert as they will all return the exact same results if the time span and field name is the same. There are a few more details in our [SRE Chatroom](http://chat.meta.stackexchange.com/transcript/message/5320725#5320725).

The current workaround is to vary the time span or bucket size a bit for each index queried.

kylebrandt:
I think this doesn't have an "easy" fix that is acceptable performance wise (not in terms of numbers, but just plain old decency) to fix this.

The bug is that the indices are not taking into account when generating a cache key (There is also an unused CacheKey function which is just sloppy). So if you have the same query but to different indices, bosun will incorrectly regard those as the same item, cache the results of one, and return that result for both requests. 

Some thoughts before I forget:

Currently we have various functions that return Indexers. Indexers are called at the same time of the query. 

Indexers basically:

1. Ask elastic for all the indexes
2. Use the arguments provided to Indexer creation functions to select which indices should be queried with various logic depending on the indexer func used.

This was already bad because we keep asking for all of the indices.

So I believe what we need to is to cache the indices as a separate cache item so we don't keep hitting them over and over again. The question is what level to cache at. 

Since all indexers when called to generate indices have to fetch all the indices. I could fetch those if I could think of a clean way to know when to invalidate them.

I could also maybe change the logic to see which indexes "could exist" based on the indexer, and check to see if those exist. I could then put them into the cache as I discover them. If all the indexes I might need are in the cache I'm good to go, if not then I can check for the specific indices I need.

Either way indices should probably have their own cache since they should have a longer lifetime in general than queries would.

Not sure what to do yet, need to let this soak in.

kylebrandt:
Too much data not to cache based on how indices are currently fetched:

```
# kbrandt at ny-bosun01.ds.stackexchange.com in ~ [14:57:51]
$ curl -XGET 'http://ny-lselastic01:9200/_all' > /dev/null
[...]

PR #1931, opened 2016-10-09
don't fetch indices , add indices to cache, and add esmonthly func

kylebrandt:


kylebrandt:
Currently if the time series is missing it errors inside ESDateHistrogram when it can't get the expected time series in the result. This result gets cached. Before when it would error on no indices in `getFn`, a wrapper for  `func (e ElasticHosts) Query(r *ElasticRequest) (*elastic.SearchResult, error)`, ensuring it was not cached. But now a lack of indices throws an error. Need to research inspecting the shards of the result and understanding what the means in this context.

kylebrandt:
I think counting the resulting shards and erroring when there were no shards that matched will get the result I want. 

However, it depends on the following PR https://github.com/olivere/elastic/pull/373 

Alternatively, I could go for the date math method mentioned in https://github.com/bosun-monitor/bosun/issues/1926#issuecomment-252287736 . But only if `allow_no_indices` applies to datemath and not just wildcard. According to the documentation it only applies to wildcard which I haven't tested yet. Even if does work I think the implementation will be a bit of pain.

kylebrandt:
note: forgot to fmt, force pushed the fmt.

cmd/bosun/expr/elastic.go
```diff
@@ -265,37 +269,47 @@ func (e ElasticHosts) Query(r *ElasticRequest) (*elastic.SearchResult, error) {
...
-	var ok bool
-	str, ok = s.(string)
-	if !ok {
-		return "", fmt.Errorf("failed to generate string representation of search source for cache key: %s", s)
+	b, err := json.Marshal(s)
+	if err != nil {
+		return "", fmt.Errorf("failed to generate json representation of search source for cache key: %s", s)
 	}
-	return str, nil
+	return fmt.Sprintf("%v\n%s", r.Indices, b), nil
```
captncraig: probably use `string(b)` here. Don't want `[42 31 13]` in you key.

captncraig: additionally, what does the source have that causes it to be a different key? What varies?

captncraig: I ask because it surprises me that something like getting a cache key even has a possibility of erroring.

kylebrandt:
fixes #1926

PR #1935, opened 2016-10-13
bosun: add annotation funcs to expr library

kylebrandt:


kylebrandt:
Antable in action in Grafana:

image

cmd/bosun/expr/annotate.go
```diff
@@ -0,0 +1,254 @@
...
+		F:      AnDurations,
+	},
+	"antable": {
+		Args:   []models.FuncType{models.TypeString, models.TypeString, models.TypeString, models.TypeString},
+		Return: models.TypeTable,
+		F:      AnTable,
+	},
+}
+
+func procDuration(e *State, startDuration, endDuration string) (*time.Time, *time.Time, error) {
```
captncraig: I'd like a better name for this if possible. It is a bit ambiguous.

cmd/bosun/expr/annotate.go
```diff
@@ -0,0 +1,254 @@
...
+	if err != nil {
+		return nil, err
+	}
+	filteredAnnotations := annotate.Annotations{}
+	for _, a := range annotations {
+		if filter == "" {
+			filteredAnnotations = append(filteredAnnotations, a)
+			continue
+		}
+		match, err := boolq.AskExpr(filter, a)
```
captncraig: Is there a way to use boolq so that we only parse the filter once?

cmd/bosun/expr/annotate.go
```diff
@@ -0,0 +1,254 @@
...
+		if err != nil {
+			return nil, err
+		}
+		if match {
+			filteredAnnotations = append(filteredAnnotations, a)
+		}
+	}
+	sort.Sort(sort.Reverse(annotate.AnnotationsByStartID(filteredAnnotations)))
```
captncraig: does the annotation server endpoint not sort them already?

cmd/bosun/expr/annotate.go
captncraig: also, why does this need to return pointers? If it is either an error, or valid times, just return `time.Time{}` in the error case. Saves some of our `*` budget.

cmd/bosun/expr/annotate.go
```diff
@@ -0,0 +1,254 @@
...
+	series := make(Series)
+	for i, a := range filteredAnnotations {
+		aStart := a.StartDate.Time
+		aEnd := a.EndDate.Time
+		inBounds := (aStart.After(*reqStart) || aStart == *reqStart) && (aEnd.Before(*reqEnd) || aEnd == *reqEnd)
+		entirelyOutOfBounds := aStart.Before(*reqStart) && aEnd.After(*reqEnd)
+		aDuration := aEnd.Sub(aStart)
+		if inBounds {
+			// time has no meaning here, so we just make the key an index since we don't have an array type
+			series[time.Unix(int64(i), 0).UTC()] = aDuration.Seconds()
```
captncraig: I hate this misuse of timestamps. It is pretty much useless without some aggregation (I'm assuming sum) is applied on top. So why not aggregate here and return a number?

cmd/bosun/expr/annotate.go
```diff
@@ -0,0 +1,254 @@
...
+		}
+		if entirelyOutOfBounds {
```
captncraig: I'd feel more comfortable if this was an else if chain. Just because.

cmd/bosun/expr/annotate.go
```diff
@@ -0,0 +1,254 @@
...
+			// Duration is equal to that of the full request
+			series[time.Unix(int64(i), 0).UTC()] = reqEnd.Sub(*reqStart).Seconds()
+		}
+		if aDuration == 0 {
+			// This would mean an out of bounds. Should never be here, but if we don't return an error in the case that we do end up here then we might panic on divide by zero later in the code
```
captncraig: this means the time filter function is broken, no? I say silently skip it. But I guess this works too. nvm.

cmd/bosun/expr/annotate.go
```diff
@@ -0,0 +1,254 @@
...
+	reqStart, reqEnd, err := procDuration(e, startDuration, endDuration)
+	if err != nil {
+		return nil, err
+	}
+	filteredAnnotations, err := getAndFilterAnnotations(e, reqStart, reqEnd, filter)
+	if err != nil {
+		return nil, err
+	}
+	series := make(Series)
+	for i, a := range filteredAnnotations {
```
captncraig: straight up duplication. Makes me sad. could potentially make a helper function that takes a `useFullDuration` flag.

at beginning set `value = (duration or 1)` by flag. And when adding to the series multiply that value by the percentage calculated in the window, and use that for the data point.

cmd/bosun/expr/annotate.go
```diff
@@ -0,0 +1,254 @@
...
+	return &Results{
+		Results: []*Result{
+			{Value: t},
+		},
+	}, nil
+}
+
+// hhmmss formats a duration into HHH:MM:SS (Hours, Minutes, Seconds) so it can be lexically sorted
+// up to 999 hours
+func hhmmss(d time.Duration) string {
```
captncraig: wouldn't that be `hhhmmss`?

cmd/bosun/expr/annotate.go
```diff
@@ -0,0 +1,254 @@
...
+	columns := strings.Split(fieldsCSV, ",")
+	columnLen := len(columns)
+	if columnLen == 0 {
+		return nil, fmt.Errorf("must specify at least one column")
+	}
+	columnIndex := make(map[string]int, columnLen)
+	for i, v := range columns {
+		// switch is so we fail before fetching annotations
+		switch v {
+		case "start", "end", "owner", "user", "host", "category", "url", "message", "duration", "link":
```
captncraig: kinda disliking this style. Could this be more data-driven? Thinking something like define a `map[string]func(annotation)interface{}` and all of the column checks and data fetching use that. Could kill the giant switch, and some of the duplication here. Maybe?

cmd/bosun/expr/annotate.go
kylebrandt: Same type that t() returns. Could make a new type but that is hard. The reason to return a series is that so you can:

 * See a list of the individual items
 * more importantly in the case of ancounts, len vs sum lets you distinguish between fractional outages vs the number of outages.

cmd/bosun/expr/annotate.go
kylebrandt: Thanks for catching this, this makes the rest of the code a lot cleaner.

PR #1938, opened 2016-10-17
cmd/bosun: don't render annotation width less than 8 pixels

### LINKED ISSUE #1937: 0-duration annotations are difficult to click on
In the bosun graph view, annotations that are 0-minutes long are very difficult (possibly impossible) to click on.

Steps to reproduce:

* Create an annotation with the same start and end timestamp.
* View a graph that includes the annotation
* The annotation appears, but clicking on it is very difficult.

kylebrandt:
fixes #1937


kylebrandt:
This will cause annotations to overflow onto each other when view long periods of time. But I don't really care right now because:

1. You can just zoom into those areas
2. I'm intend to add filtering to the UI in the not to distant future now that annotations have ask method can be filtered in memory using boolq

PR #1941, opened 2016-10-19
cmd/scollector: check for NaN and Inf for float64 datapoints

mathpl:
Catches Inf and NaN on float before we hit the json marshaller. Symptom before this:

```
error: queue.go:105: json: error calling MarshalJSON for type *opentsdb.DataPoint: Unparseable number NaN
info: queue.go:124: restored 2000, sleeping 5s
error: queue.go:105: json: error calling MarshalJSON for type *opentsdb.DataPoint: Unparseable number -Infinity
info: queue.go:124: restored 2000, sleeping 5s
```

### DIFF
--- a/opentsdb/tsdb.go
+++ b/opentsdb/tsdb.go
@@ -93,7 +93,7 @@ func (d *DataPoint) Valid() bool {
 		return false
 	}
 	f, err := strconv.ParseFloat(fmt.Sprint(d.Value), 64)
-	if err != nil || math.IsNaN(f) {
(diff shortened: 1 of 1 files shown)


captncraig:
Is there a better place to catch this? What collectors are inserting NaNs?

mathpl:
That is the question, I'm not sure! I know the impact though, no more metrics out of that scollector. I've seen this happen in the cassandra collector (which I've made a PR for a while back) and I'll hunt down the source of this one as well.

This however is a lot safer since we can keep sending out metrics and still get feedback that something is wrong via scollector.discarded.

PR #1942, opened 2016-10-19
cmd/scollector: configurable redis

mathpl:
Allows for configurable redis instances while retaining the current auto-config behavior based on ps when on linux. Also allows to select usec instead of msec for cmdstats and to disable reporting of redis.cmd_Xsec_pc.

Both collectors reuse the same collecting code moved to redis.go.


cmd/scollector/collectors/redis_remote.go
```diff
@@ -0,0 +1,23 @@
...
+
+	"bosun.org/cmd/scollector/conf"
+	"bosun.org/opentsdb"
+)
+
+func init() {
+	registerInit(startRedisRemoteCollector)
+}
+
+func startRedisRemoteCollector(c *conf.Conf) {
```
captncraig: not sure why this code could not be combined with the other redis init function and this file would not be needed.

cmd/scollector/conf/conf.go
```diff
@@ -209,6 +210,13 @@ type RedisCounters struct {
 	Database int
 }
 
+type Redis struct {
+	Address      string
+	UseMsec      bool
```
captncraig: I'm not sold on a flag to change units. Why is this necessary again?

cmd/scollector/collectors/redis_remote.go
captncraig: oh, because the other one is linux only. got it.

cmd/scollector/conf/conf.go
mathpl: This is to optionally prevent pushing floats to opentsdb while retaining current behavior by default. For context, storing float in opentsdb systematically uses 4 bytes [1], integer can go down to a single byte and will only use more if needed. Storing as float provide no additional precision in this case since the full precision of the data is an integer in redis. Considering the time serie using this (all redis calls stats) it will have a mesurable impact on that column size.

1. http://opentsdb.net/docs/build/html/user_guide/backends/hbase.html#data-point-columns

mathpl:
This is indeed a bit more than barebone configurable endpoints for redis collector. I needed to move out everything I could out of _linux for obvious reasons which certainly makes the diff a PITA for large chunks of the code. All the really changed in c_redis() is the handling of the instances loop which remained in redis_linux.go.

The options UseMsec and PerCallStats are about reporting the "right thing", but I agree that is subjective.

PR #1946, opened 2016-10-24
WIP: Authentication

captncraig:


cmd/bosun/conf/conf.go
```diff
@@ -31,7 +31,9 @@ import (
 // outside of the package without a setter
 type SystemConfProvider interface {
 	GetHTTPListen() string
-	GetRelayListen() string
+	GetHTTPSListen() string
+	GetTLSCertFile() string
```
kylebrandt: PEM would be nice, or a interface that will allow for it in the future?

cmd/bosun/conf/system.go
```diff
@@ -140,6 +145,43 @@ type SMTPConf struct {
...
+}
+
+type LDAPConf struct {
+	// Domain name (used to make domain/username)
+	Domain string
+	// LDAP server
+	LdapAddr string
+	// allow insecure ldap connection?
+	AllowInsecure bool
+	// default permission level for anyone who can log in. Try "Reader".
```
kylebrandt: Where does "Reader" come from - what else should I try? :P

cmd/bosun/main.go
```diff
@@ -284,6 +283,9 @@ func watch(root, pattern string, f func()) {
 				if wait.After(time.Now()) {
 					continue
 				}
+				if event.Name == "web/static.go" {
```
kylebrandt: What does this do? Add comment maybe?

cmd/bosun/web/middlewares.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/captncraig/easyauth/providers/token/redisStore"
+	"github.com/gorilla/mux"
+
+	"bosun.org/cmd/bosun/conf"
+	"bosun.org/collect"
+	"bosun.org/metadata"
+	"bosun.org/opentsdb"
+)
+
+// custom middlewares for bosun. Must match  alice.Constructor signature (func(http.Handler) http.Handler)
```
kylebrandt: whitespace in comment

cmd/bosun/web/middlewares.go
```diff
@@ -0,0 +1,132 @@
...
+
+var miniprofilerMiddleware = func(next http.Handler) http.Handler {
```
kylebrandt: miniProfiler? (casing)

cmd/bosun/web/middlewares.go
```diff
@@ -0,0 +1,132 @@
...
+		t := collect.StartTimer("http_routes", opentsdb.TagSet{"route": routeName})
+		next.ServeHTTP(w, r)
+		t()
+	})
+}
+
+type noopAuth struct{}
+
+func (n noopAuth) GetUser(r *http.Request) (*easyauth.User, error) {
+	//TODO: make sure ui sends header when possible, instead of json body
```
kylebrandt: outstanding TODO?

cmd/bosun/web/middlewares.go
```diff
@@ -0,0 +1,132 @@
...
+	name := "anonymous"
```
kylebrandt: constant?

cmd/bosun/web/middlewares.go
```diff
@@ -0,0 +1,132 @@
...
+	return &easyauth.User{
+		Access:   roleAdmin,
+		Username: name,
+		Method:   "noop",
+	}, nil
+}
+
+func buildAuth(cfg conf.AuthConf) (easyauth.AuthManager, *token.TokenProvider, error) {
+	if cfg.CookieSecret == "" {
+		cfg.CookieSecret = "CookiesAreInsecure"
```
kylebrandt: constant again?

cmd/bosun/web/middlewares.go
```diff
@@ -0,0 +1,132 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+	if cfg.AuthEnabled == false {
+		auth.AddProvider("nop", noopAuth{})
+	} else {
+		authEnabled = true
+		// in proc token store so bosun can send data to itself with an ephemeral token
+		selfStore, _ := token.NewJsonStore("")
+		selfToken := token.NewToken(easyauth.RandomString(16), selfStore)
```
kylebrandt: 16?

cmd/bosun/web/middlewares.go
```diff
@@ -0,0 +1,132 @@
...
+			return nil, nil, err
+		}
+		auth.AddProvider("ldap", l)
+	}
+	var authTokens *token.TokenProvider
+	if cfg.TokenSecret != "" {
+		tokensEnabled = true
+		data, ok := schedule.DataAccess.(redisStore.Connector)
+		if !ok {
+			return nil, nil, fmt.Errorf("web's data access does not implement correct redis connector interface")
```
kylebrandt: this an error that will bubble up to a user? If so - confusing.

cmd/bosun/web/middlewares.go
```diff
@@ -0,0 +1,132 @@
...
+func buildAuth(cfg conf.AuthConf) (easyauth.AuthManager, *token.TokenProvider, error) {
+	if cfg.CookieSecret == "" {
+		cfg.CookieSecret = "CookiesAreInsecure"
+	}
+	auth, err := easyauth.New(easyauth.CookieSecret(cfg.CookieSecret))
+	if err != nil {
+		return nil, nil, err
+	}
+	if cfg.AuthEnabled == false {
+		auth.AddProvider("nop", noopAuth{})
```
kylebrandt: log warn that there is no auth on start

cmd/bosun/web/roles.go
```diff
@@ -0,0 +1,89 @@
...
+	Bits easyauth.Role
+	Name string
+	Desc string
+}
+type roleMetadata struct {
+	Permissions []bitDesc
+	Roles       []bitDesc
+}
+
+func parseRole(s string) (easyauth.Role, error) {
```
kylebrandt: string parsing combined with auth scares me - is this our best option?

cmd/bosun/conf/conf.go
captncraig: This is to correspond to [http.ListenAndServeTLS](https://golang.org/pkg/net/http/#ListenAndServeTLS) which expects two key file locations. They are pem encoded, yes, but maybe you mean a combined keyfile of some kind. There are other ways to host a tls server, but this is the easiest.

cmd/bosun/conf/system.go
captncraig: It comes from the string parsing in roles.go. Any permission or role name should be valid, comma seperated.

cmd/bosun/main.go
captncraig: Stops the server from killing itself after esc runs. This was debugging that can probably be removed.

cmd/bosun/web/embed.go
```diff
@@ -47,22 +46,12 @@ func RunTsc() {
...
-	if err != nil {
-		log.Fatal(err)
-	}
-	if err := c.Start(); err != nil {
-		log.Fatal(err)
-	}
-	go func() { io.Copy(os.Stdout, stdout) }()
-	go func() { io.Copy(os.Stderr, stderr) }()
-	if err := c.Wait(); err != nil {
+	log.Println("Starting")
```
captncraig: also debugging to remove

cmd/bosun/web/middlewares.go
captncraig: not outstanding. todone.

cmd/bosun/web/middlewares.go
captncraig: Good idea. Should probably link to docs on bosun.org. Once I add them.

cmd/bosun/web/web.go
```diff
@@ -90,56 +103,80 @@ func Listen(listenAddr string, devMode bool, tsdbHost string, reloadFunc func()
...
-	if schedule.SystemConf.ReloadEnabled() { // Is true of save is enabled
-		router.Handle("/api/reload", JSON(Reload)).Methods(http.MethodPost)
+	handle("/api/config_test", JSON(ConfigTest), canViewConfig).Name("config_test").Methods(POST)
+	handle("/api/save_enabled", JSON(SaveEnabled), fullyOpen).Name("seve_enabled").Methods(GET)
+
+	if schedule.SystemConf.ReloadEnabled() {
+		handle("/api/reload", JSON(Reload), canSaveConfig).Name("can_save").Methods(POST)
 	}
+
 	if schedule.SystemConf.SaveEnabled() {
```
kylebrandt: I like that we are more specific about the methods. But may need to note is as a break change if people are doing odd things.

cmd/bosun/web/middlewares.go
captncraig: 16 bytes is an adequate hash secret. magic number because its not worth making a constant over.

kylebrandt: are you sure you want to justify a magic number on public record? ;-)

cmd/bosun/web/roles.go
captncraig: This is just for converting roles in the config file to our numeric values. Should never be used at runtime. Maybe this should be somewhere in the config package or something?

kylebrandt:
When I run Bosun without auth, I see my username and a log out drop down, even thought I am not logged in. This probably should not be visible in no auth mode:

image

kylebrandt:
When running with a self-signed cert, the logs get spammed repeatedly with:

```
$ go run main.go -c bosun.toml -w -r -q -skiplast -n
2016/12/08 12:57:10 info: main.go:149: readonly relay at http://127.0.0.1:63606 to http://ny-tsdb02:4242
2016/12/08 12:57:10 info: web.go:224: tsdb host: 127.0.0.1:63606
2016/12/08 12:57:10 info: web.go:234: bosun web listening https on: :8080
2016/12/08 12:57:10 info: main.go:277: watching *.go in .
2016/12/08 12:57:10 info: main.go:277: watching *.html in web/static/templates
2016/12/08 12:57:10 info: main.go:277: watching *.ts in web/static/js
2016/12/08 12:57:25 error: queue.go:117: Post https://localhost:8080/api/put: x509: certificate signed by unknown authority
```

not sure why it is doing that with the -r flag anyways?

kylebrandt:
Actually, that error may be a problem regardless:
```
2016/12/13 12:16:28 error: queue.go:117: Post https://localhost:8080/api/put: x509: certificate is valid for bosun, not localhost
```

captncraig:
Ah yes. That does suck. I'm thinking we should just call the method directly if we are doing the in-proc mode. Perhaps a func in collect/metadata that can be swapped out to replace the http call altogether with a simple function call to the appropriate handler. 

but datapoints really should be going out through tsdbrelay anyway. Holy shit this is a mess.

kylebrandt:
Mixed content due to jquery: Mixed Content: The page at 'https://bosun:8080/login/' was loaded over HTTPS, but requested an insecure script 'http://code.jquery.com/jquery-3.1.1.min.js'. This request has been blocked; the content must be served over HTTPS.

captncraig:
well that's just sloppy.

collect/queue.go
```diff
@@ -174,9 +176,15 @@ func SendDataPoints(dps []*opentsdb.DataPoint, tsdb string) (*http.Response, err
 		req.Header.Set("X-Access-Token", AuthToken)
 	}
 	Add("collect.post.total_bytes", Tags, int64(buf.Len()))
+	client := DefaultClient
+	if strings.Contains(tsdb, "localhost") {
```
nicollet: Just out of curiosity, can it also have "127.0.0.1"?

kylebrandt:
Couple things regarding the bosun self metrics issue:

 * Is there some other communication channel we could use other than the HTTP route for the self metrics so we don't have to do this auth thing against localhost?
 * Unless you use influx or OpenTSDB, these metrics can't go anywhere. Might be good to wrap them under an option anyways

metadata/metadata.go
```diff
@@ -251,10 +259,25 @@ type Metasend struct {
 	Time   *time.Time `json:",omitempty"`
 }
 
-//AuthToken to use to talk to bosun
-var AuthToken string
-
 func sendMetadata(ms []Metasend) {
+	if putFunction != nil {
```
kylebrandt: could dedent by inverting condition and then returning

collect/queue.go
```diff
@@ -156,6 +158,33 @@ var bufferPool = sync.Pool{
...
+		return rec.Result(), nil
+	}
+	client := DefaultClient
+
+	if UseNtlm {
+		resp, err := ntlm.DoNTLMRequest(client, req)
+		if err != nil {
+			return nil, err
+		}
+		if resp.StatusCode == 401 {
```
kylebrandt: `http.StatusUnauthorized` constant

metadata/metadata.go
```diff
@@ -209,6 +208,15 @@ func Init(u *url.URL, debug bool) error {
 	return nil
 }
 
+var putFunction func(k Metakey, v interface{}) error
+
+func InitF(debug bool, f func(k Metakey, v interface{}) error) error {
```
kylebrandt: Comment explaining InitF

Issue #1157, opened 2015-07-13
Missing Results from .LeftJoin


kylebrandt:
I had a LeftJoin statement that was producing no results to range over. After copying the expression to the expression page I found that I was getting a error that response size was too large. I then shrank the time duration of the expression in the template and got results.

So it seems like when the response limit is hit within a .LeftJoin the error get buried -- perhaps this is true of other .Eval like funcs as well.

Marked as critical since silent errors create a lot of confusion and eat up time debugging.

PR #1950, opened 2016-10-25
bosun: make it so errors can be handled in templates

kylebrandt:


kylebrandt:
Golang templates don't let you catch an error with multiple value assignment in as you would in Go. For example, `$foo, $err := func(..)` is not something you can do in templates.

The problem this creates is that if you want to have templates that include information that could fail, but isn't vital to the template, the error can not be handled and will cause the entire template to fail. As example, you might want to query elastic for recent logs all host based alerts, but not have the entire host based alert fail just because you could not fetch the logs from elastic.

This takes the following approach to handling errors in templates:

1.  In cases where an error is created in a context-bound function (see the sixDocs branch for what this means) the template variable `.Errors` has the error appended to it (it is a slice of strings).
2. Global functions always returns strings except for parseDuration. In this case c.Errors is not appended to, but the string that would have been returned with an error is show in the template. parseDuration returns nil when it errors, and in this one exception you can't see what the error is.
3. In all other cases (most cases the context-bound funcs) .Error is always appended to with the error. If the func returns a string or an image (technically an interface) the error message will be displayed in the template. If an object is returned (i.e. a result sets, a slice, etc) nil is returned and the user can check for that in the template.

This shouldn't break any existing template definitions. Template errors are just less likely to raise a template error, if you try work with an returned object that errored, errRecover will catch the panic, and a template rendering error would occur as before.

kylebrandt:
Oh, also it should fix https://github.com/bosun-monitor/bosun/issues/1157 , a variable was mistaken shadow'd in eval @ https://github.com/bosun-monitor/bosun/pull/1950/files#diff-086fe564522de5d9992d50fecb7c2ecaL221

cmd/bosun/conf/rule/rule.go
```diff
@@ -379,22 +379,29 @@ func (c *Conf) loadMacro(s *parse.SectionNode) {
...
+			return conf.ByteSize(v).String()
 		case expr.Number:
-			return conf.ByteSize(v), nil
+			return conf.ByteSize(v).String()
 		case expr.Scalar:
-			return conf.ByteSize(v), nil
+			return conf.ByteSize(v).String()
 		}
-		return conf.ByteSize(0), fmt.Errorf("unexpected type: %T (%v)", v, v)
+		return fmt.Errorf("unexpected type passed to bytes function: %T (%v)", v, v).Error()
```
captncraig: how can the template identify succes / failure here? Is our argument that it doesn't matter? Showing `Data was error parsing float bytes` in a template may be odd.

cmd/bosun/sched/template.go
```diff
@@ -230,78 +232,96 @@ func (c *Context) eval(v interface{}, filter bool, series bool, autods int) (res
 	default:
 		return nil, "", fmt.Errorf("expected string, expression or resultslice, got %T (%v)", v, v)
 	}
-	if filter {
+	if filter && err != nil {
```
captncraig: is this logic wrong? Do you mean `== nil`? Maybe return early if error?

cmd/bosun/sched/template.go
```diff
@@ -230,78 +232,96 @@ func (c *Context) eval(v interface{}, filter bool, series bool, autods int) (res
...
 		}
 	case opentsdb.TagSet:
 		t = v
 	}
 	l := c.schedule.RuleConf.GetLookup(table)
 	if l == nil {
-		return "", fmt.Errorf("unknown lookup table %v", table)
+		err := fmt.Errorf("unknown lookup table %v", table)
+		c.addError(err)
+		return err.Error()
```
captncraig: I feel like returning empty string here would make string functions more consistent. Still evaluates falsey in templates.

cmd/bosun/sched/template.go
```diff
@@ -475,78 +510,92 @@ func (c *Context) HTTPPost(url, bodyType, data string) string {
...
 
-func (c *Context) LSQuery(index_root, filter, sduration, eduration string, size int) (interface{}, error) {
+func (c *Context) LSQuery(index_root, filter, sduration, eduration string, size int) interface{} {
 	var ks []string
 	for k, v := range c.AlertKey.Group() {
 		ks = append(ks, k+":"+v)
 	}
-	return c.LSQueryAll(index_root, strings.Join(ks, ","), filter, sduration, eduration, size)
+	res := c.LSQueryAll(index_root, strings.Join(ks, ","), filter, sduration, eduration, size)
+	if res != nil {
```
captncraig: Is `!=` correct here?

captncraig: why not just return LSQueryAll ?

cmd/bosun/conf/rule/rule.go
kylebrandt: limitation of global funcs, not really sure any way around that.

cmd/bosun/sched/template.go
kylebrandt: will fix, the shared err return is confusing anyways. As proved by the existing flaw in logic that was already there with shadowing lin 233.

cmd/bosun/sched/template.go
kylebrandt: Unless there are cases where "" is a valid result, not sure if there are any, but not fond of the precedent. I think I would rather see errors in the templates for items that lose the entire template in case of strings and images.

cmd/bosun/sched/template.go
kylebrandt: agreed changing

PR #1964, opened 2016-11-18
cmd/scollector: Track process PIDs.

alienth:
As requested by @kylebrandt .

### DIFF
--- a/cmd/scollector/collectors/collectors.go
+++ b/cmd/scollector/collectors/collectors.go
@@ -99,6 +99,7 @@ const (
 	osProcCPUDesc        = "The summed percentage of CPU time used by processes with this name (0-100)."
 	osProcMemRealDesc    = "The total amount of real memory used by the processes with this name. For Linux this is RSS and in Windows it is the private working set."
 	osProcMemVirtualDesc = "The total amount of virtual memory used by the processes with this name."
+	osProcPID            = "The PID of the process being tracked by a given ID tag. As this metric value represents the actual PID, it is not suitable for any form of aggregation."
 	osServiceRunningDesc = "1: active, 0: inactive"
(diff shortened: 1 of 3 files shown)


kylebrandt:
From Bosun chat for reference:


stianovrevage	[6:28 AM]  
Hi guys, I'm looking at https://github.com/bosun-monitor/bosun/blob/master/cmd/scollector/collectors/processes_windows.go but cannot figure out why the ID reported is not the correct PID...

kylebrandt	[7:48 AM]  
PID is not submitted, just ids are generated from a bucket for numbers

stianovrevage	[8:58 AM]  
Why?

dima.medvedev	[9:14 AM]  
Mainly because opentsdb has a hard limit on tag values

[9:15]  
I don't remember how it's done in windows, but linux process collector tries to make process ids unique for one running scollector instance.

[9:16]  
Also, it's still possible to identify process via their process start time.

stianovrevage	[9:24 AM]  
Ok, thanks for a good answer :smile:

kylebrandt	[9:24 AM]  
Also some services might recycle processes frequently

stianovrevage	[9:36 AM]  
True, but would be nice to have the option of including PID, for us who don't run into high cardinality issues

[9:36]  
image-name + pid should be pretty safe I would think

kylebrandt	[9:38 AM]  
Could send it as a metric

[9:38]  
Never really thought about that before, it would be strange, but functional

[9:40]  
linux.proc.id_pid or something. Where you host,id,name are tags and the value is the corresponding pid at the time (edited)

[9:43]  
I think perfmon has that actually (WMI)

[9:43]  
A metric like that

stianovrevage	[9:43 AM]  
Yep, and just reference that when you actually need to find the process

[9:43]  
Yes, it has CurrentID I think

[9:43]  
Which is the current PID

gbrayut:
Tested locally... LGTM

PR #1976, opened 2016-12-02
Update InfluxDB client to v2

krise3k:
Update InfluxDB client to v2.
Fix https://github.com/bosun-monitor/bosun/issues/1886
Relates to https://github.com/bosun-monitor/bosun/pull/1890

Run tests on InfluxDB 1.1.0:
- [x] single argument functions (max, min, sum, ...)
- [ ] multiple arguments functions (moving_average)

Unfortunately InfluxDB client v2 is not compatible with 0.9.x InfluxDB


gbrayut:
Did some testing on this using Influx 1.1 and Telegraf 1.1. I know you marked it WIP but I think it is ready to merge. There are some issues with nested functions, but those can be addressed in another PR (they are bosun issues not issues with the InfluxDBClient)

The following [moving average]( https://docs.influxdata.com/influxdb/v1.1/query_language/functions/#moving-average) expressions work as expected:
```
# Using date 2016-12-05 time 22:05 on the expression page
influx("telegraf",'''SELECT mean(usage_idle) FROM cpu WHERE cpu = 'cpu-total' GROUP BY host''',"5m","0m","1m")

# Matches the following influxql query
SELECT mean(usage_idle) FROM cpu WHERE cpu = 'cpu-total' AND time >= '2016-12-05 22:00:00' AND time <= '2016-12-05 22:05:00' GROUP BY host,time(1m)

# 60s Moving average of idle cpu (assuming 10s samples)
influx("telegraf",'''SELECT moving_average(usage_idle,6) FROM cpu WHERE cpu = 'cpu-total' GROUP BY host''',"5m","","")
#Works as expected
```

I ran into some issues trying to test moving average with a nested function, as the time manipulation in the [influxQueryDuration function](https://github.com/bosun-monitor/bosun/blob/cd474c3511699299f5ca76150cb60218afc91148/cmd/bosun/expr/influx.go#L95) keeps getting in the way. We currently try to parse the Influx expression before we add the "time >= ... time <=" section to the where clause and before we add the time(...) to the group by. This causes the call to influxql.ParseStatement to fail even though the final query that would be generated works as expected:

```
# This should work but fails with: moving_average aggregate requires a GROUP BY interval (since that expression isn't added until after the call to influxql.ParseStatement)
influx("telegraf",'''SELECT moving_average(min(usage_idle),6) FROM cpu WHERE cpu = 'cpu-total' GROUP BY host''',"5m","0m","1m")

# Adding ,time(1m) to the group by causes: aggregate functions with GROUP BY time require a WHERE time clause (same issue as above)
influx("telegraf",'''SELECT moving_average(min(usage_idle),6) FROM cpu WHERE cpu = 'cpu-total' GROUP BY host,time(1m)''',"5m","0m","1m")

#If I rip out the time logic from influxQueryDuration and specify the full query it works as expected
influx("telegraf",'''SELECT moving_average(min(usage_idle),6) FROM cpu WHERE time >= '2016-12-05 22:00:00' AND time <= '2016-12-05 22:05:00' AND cpu = 'cpu-total' GROUP BY host,time(1m)''',"5m","0m","")
```

Let me know if there are any other changes you wanted to make, but I think this is ready to merge. Then we can start a new PR with `SELECT moving_average(min(usage_idle),6) FROM cpu WHERE cpu = 'cpu-total' GROUP BY host,time(1m)` as a new test cases to influx_test.go and rework the influxQueryDuration to validate that correctly.

krise3k:
@gbrayut, thank you for reviewing changes. 

When I dig problems with nested functions, I found that we using full AST from InfluxDB, what causes validation problem, which you mentioned. I have started working on simpler query parser.

I found one inconsistency in [docs](https://github.com/bosun-monitor/bosun/blob/master/docs/configuration.md) and [example config](https://github.com/bosun-monitor/bosun/blob/master/cmd/bosun/bosun.example.toml#L85).
Docs saying that we have to set
> * influxHost: InfluxDB host address ip:port pair.
> * influxUsername: InfluxDB username. If empty will attempt to connect without authentication.
> * influxPassword: InfluxDB password. If empty will attempt to connect without authentication.
> * influxTLS: Whether to use TLS when connecting to InfluxDB. Default is false.
> * influxTimeout: Timeout duration for connections to InfluxDB.

and example config:
```
URL = "https://myInfluxServer:1234"
Timeout = "5m"
UnsafeSSL = true`
```
Which version should I use?

gbrayut:
Awesome! Let me know when you have something ready to review.

It looks like the docs are incorrect (even in the [sixDocs](https://github.com/bosun-monitor/bosun/blob/sixDocs/docs/configuration.md) documentation staging branch for 6.0). The config should parse into the [InfluxConf struct](https://github.com/bosun-monitor/bosun/blob/74b3c7f0948b2a9f50813a9f15e4c23147f2645c/cmd/bosun/conf/system.go#L114), so the example config is correct:

```
URL = "https://myInfluxServer:1234"
Username = "user"
Password = "password"
Timeout = "5m"
UnsafeSSL = true`
```

I'll make sure the docs get updated. Thanks!

krise3k:
So it's ready to merge.

gbrayut:
Cool... Merged via https://github.com/bosun-monitor/bosun/commit/1acb03a83d88ab67829c47fbed7a61bbba336d19

PR #1983, opened 2016-12-12
Introduce some tests in ifstat_linux

nicollet:
One little test in the ifstat_linux collector.
Will just test getting the speed of an interface.

More to come if this is welcomed.


nicollet:
I can do better.

Will repost with more interesting stuff, and well formatted.

Issue #1912, opened 2016-09-16
ifstat metrics not reported for some interfaces based on their name


ncabatoff:
My desktop's NIC is named enp0s25.  A virtualbox guest of it has enp0s3.  Neither of them produce ifstat metrics like linux_net_packets, linux_net_errs, etc.  The code seems to require that the second line of /proc/net/dev must match this RE:

`\s(eth\d+|em\d+_\d+/\d+|em\d+_\d+|em\d+|bond\d+|team\d+|p\d+p\d+_\d+/\d+|p\d+p\d+_\d+|p\d+p\d+):(.*)`

It looks like my NICs are being rejected on two counts.  First, their names are not in the whitelisted list of patterns.  Second, their names are too long, so Linux doesn't put a leading space (the \s at the start of the pattern.)  For example:

```
Inter-|   Receive                                                |  Transmit
 face |bytes    packets errs drop fifo frame compressed multicast|bytes    packets errs drop fifo colls carrier compressed
enp0s3: 1006267169 6639248    0    0    0     0          0    256352 5176892906 3667957    0    0    0     0       0          0
    lo: 169963776847 304993459    0    0    0     0          0         0 169963776847 304993459    0    0    0     0       0          0
```

```
Inter-|   Receive                                                |  Transmit
 face |bytes    packets errs drop fifo frame compressed multicast|bytes    packets errs drop fifo colls carrier compressed
enp0s25: 53576888423 79602006    0    0    0     0          0   4945828 16893672173 44642474    0    0    0     0       0          0
    lo: 1138405080 1845942    0    0    0     0          0         0 1138405080 1845942    0    0    0     0       0          0
```

By comparison, here's one that does work:

```
Inter-|   Receive                                                |  Transmit
 face |bytes    packets errs drop fifo frame compressed multicast|bytes    packets errs drop fifo colls carrier compressed
    lo:3038824498023 3525462214    0    0    0     0          0         0 3038824498023 3525462214    0    0    0     0       0          0
  eth0:96763269978 225152313    0    0    0     0          0  16490262 142000936142 190268875    0    0    0     0       0          0
```

kylebrandt:
https://github.com/bosun-monitor/bosun/pull/1756#issuecomment-228054796

PR #1985, opened 2016-12-14
cmd/scollector: Update linux interface collector names.

sysadmind:
Add systemd predictable network interface names into linux network name regex. The systemd mention of these prefixes exists here: https://github.com/systemd/systemd/blob/master/src/udev/udev-builtin-net_id.c#L29

This adds `ww`, `wl`, and `en` to the possible prefixes for interface names and also changes the space up front to be optional. When the names get too long for the column, I have noticed that the text aligns to the start of the line instead of the colon.

This should fix issues others have been having with network interfaces not showing up. 
Related:
https://github.com/bosun-monitor/bosun/pull/1756
https://github.com/bosun-monitor/bosun/issues/1912

### DIFF
--- a/cmd/scollector/collectors/ifstat_linux.go
+++ b/cmd/scollector/collectors/ifstat_linux.go
@@ -51,8 +51,11 @@ var netFields = []struct {
 	{"compressed", metadata.Counter, metadata.Count},
 }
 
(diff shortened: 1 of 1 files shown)


peterfroehlich:
@kylebrandt could you please review and approve this PR? 
Sadly systemd is not really something that can be ignored anymore...

peterfroehlich:
Thanks! ❤️

PR #2004, opened 2017-01-06
cmd/scollector: Added Google Cloud Billing collector

mhenderson-so:
Added Google Cloud billing collector. Documentation on configuring the collector is included.


cmd/scollector/doc.go
```diff
@@ -362,6 +362,27 @@ the group number for this group and put it in here.
...
+
+  [[GoogleCloudBilling]]
+    BucketName = "example-billing-bucket"
+    AuthBase64 = "ew0KICAidHlwZSI6ICJzZXJ2aWNlX2FjY291bnQiLA0KICAicHJvamVjdF9pZCI6ICJleGFtcGxlLXByb2plY3QiLA0KICAicHJpdmF0ZV9rZXlfaWQiOiAiMDhkZmY4ZDRhZjAyZTUzOTA5YzAyMGViM2MyNjg0ZWU2ZDc5NzFlNyIsDQogICJwcml2YXRlX2tleSI6ICItLS0tLUJFR0lOIFBSSVZBVEUgS0VZLS0tLS1cNG5jUldHbzBQMldhVVAxV01DUHJIWHk2eE1EU2dhTXNHSlpMb2hiQVxuUEo5MXc1SDFKbkRNY21SWFg4RUp2VEZGamZTb0R0d2ZwRmswTVYwUEFvR0JBSTZQR3BEVWdXMWRRSzVGQ2hQd1xuTHZkZlRXUGE5NmpadEpcbi0tLS0tRU5EIFBSSVZBVEUgS0VZLS0tLS1cbiIsDQogICJjbGllbnRfZW1haWwiOiAic2NvbGxlY3Rvci1iaWxsaW5nQGV4YW1wbGUtcHJvamVjdC5pYW0uZ3NlcnZpY2VhY2NvdW50LmNvbSIsDQogICJjbGllbnRfaWQiOiAiNTg3NDU4NDEyNDUyNTYwMTUwOCIsDQogICJhdXRoX3VyaSI6ICJodHRwczovL2FjY291bnRzLmdvb2dsZS5jb20vby9vYXV0aDIvYXV0aCIsDQogICJ0b2tlbl91cmkiOiAiaHR0cHM6Ly9hY2NvdW50cy5nb29nbGUuY29tL28vb2F1dGgyL3Rva2VuIiwNCiAgImF1dGhfcHJvdmlkZXJfeDUwOV9jZXJ0X3VybCI6ICJodHRwczovL3d3dy5nb29nbGVhcGlzLmNvbS9vYXV0aDIvdjEvY2VydHMiLA0KICAiY2xpZW50X3g1MDlfY2VydF91cmwiOiAiaHR0cHM6Ly93d3cuZ29vZ2xlYXBpcy5jb20vcm9ib3QvdjEvbWV0YWRhdGEveDUwOS9zY29sbGVjdG9yLWJpbGxpbmclNDBleGFtcGxlLXByb2plY3QuaWFtLmdzZXJ2aWNlYWNjb3VudC5jb20iDQp9DQo="
```
captncraig: would it be easier to just put the json file on disk and read it? Copy 3 or 4 fields from it into toml? All solutions are weird.

cmd/scollector/collectors/google_cloudbilling.go
```diff
@@ -0,0 +1,234 @@
+package collectors
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"strconv"
+	"time"
+
+	"golang.org/x/oauth2/google"
```
captncraig: organize these better. We usually use three blocks: std lib, third party, bosun.

cmd/scollector/collectors/google_cloudbilling.go
```diff
@@ -0,0 +1,234 @@
...
+
+const (
+	gcbHoursInDay = 24
+	gcbUsageDesc  = "Usage of consuming given Google Cloud services"
+	gcbCostDesc   = "Cost of consuming given Google Cloud services"
+)
+
+// GoogleCloudBilling instantiates the GoogleCloudBilling collector
+func GoogleCloudBilling(bucketName, authBase64 string) error {
+	if bucketName != "" {
```
captncraig: perhaps print an error if the bucket is not given?

cmd/scollector/collectors/google_cloudbilling.go
```diff
@@ -0,0 +1,234 @@
...
+		})
+	}
+
+	return nil
+}
+
+func c_googlecloudbilling() (opentsdb.MultiDataPoint, error) {
+	var md opentsdb.MultiDataPoint
+	//Create a connection to the Google Cloud services
+	gcClient, err := createGCStorageClient()
```
captncraig: can we create the client once and reuse it?

cmd/scollector/collectors/google_cloudbilling.go
```diff
@@ -0,0 +1,234 @@
...
+	//Get the list of available files from the storage bucket
+	reports, err := gcClient.Objects.List(gcbConf.BucketName).Do()
+	if err != nil {
+		return nil, err
+	}
+	//We want to process bills from the last 4 days. This is because Google Cloud seems to have significant
+	//lag when generating the bill exports. They can be 2 or more days old by the time the bill is put into the
+	//storage bucket.
+	dateStart := time.Now().Add(-96 * time.Hour)
+	dateRangeStart := time.Date(dateStart.Year(), dateStart.Month(), dateStart.Day(), 0, 0, 0, 0, time.UTC)
```
captncraig: this might be cleaner with [time.Truncate](https://golang.org/pkg/time/#Time.Truncate).

cmd/scollector/collectors/google_cloudbilling.go
```diff
@@ -0,0 +1,234 @@
...
+		//If the last part of the filename does not have 10 or more characters, then it's not datestamped
+		if len(fileName) < 10 {
+			continue
+		}
+		dateString := fileName[len(fileName)-10:]
+		dateDate, err := time.Parse(dateLayout, dateString)
+		if err != nil {
+			continue
+		}
+		if !(dateDate.Before(dateRangeEnd) && dateDate.After(dateRangeStart)) {
```
captncraig: this is a bit confusing. `before(start) || after(end)` might be simpler.

cmd/scollector/collectors/google_cloudbilling.go
```diff
@@ -0,0 +1,234 @@
...
+			continue
+		}
+		//Ok, so we want to process this. Now we need to download the file, convert it to a string containing the JSON,
+		//unmarshall that into the bill struct and send it off for processing
+		thisItemResponse, err := gcClient.Objects.Get(item.Bucket, item.Name).Download()
+		if err != nil {
+			return md, fmt.Errorf("gcloud: Could not download %s: %s", item.Name, err)
+		}
+		buf := bytes.NewBuffer(make([]byte, 0, thisItemResponse.ContentLength))
```
captncraig: just make a json.NewDecoder(thisItemResponse.Body) and decode from that directly. No need to read the content.

cmd/scollector/collectors/google_cloudbilling.go
```diff
@@ -0,0 +1,234 @@
...
+		if err != nil {
+			slog.Errorf("gcloud: could not convert %s to time.Time: %s", item.StartTime, err)
+			continue
+		}
+		endTime, err := time.Parse(billTimeFormat, item.EndTime)
+		if err != nil {
+			slog.Errorf("gcloud: could not convert %s to time.Time: %s", item.EndTime, err)
+			continue
+		}
+		startTime = startTime.In(time.UTC) //We want to send UTC data
```
captncraig: We usually use startTime.UTC() for this, rather than In

cmd/scollector/doc.go
mhenderson-so: I don't like either way, but I like the idea of having an extra file on disk that needs securing less than this, given I would hope that this file is secure on installations anyway. That said, if people have set their permissions sensibly (by giving read-only access to only the billing bucket), there is limited damage you can do with these creds.

cmd/scollector/collectors/google_cloudbilling.go
mhenderson-so: I couldn't get time.Truncate to do what I wanted here, because I don't want to actually round the time to an interval of 4 days, I just want "now minus 4 days" (or 96 hours, because screw timezone changes)

cmd/scollector/doc.go
captncraig: We just re-did the google auth stuff to do json key stuff. We should re-use the code from the other collector (analytics I think)

PR #2009, opened 2017-01-28
Better support of REST API notifications.

akharlamov:
I have introduced additional subtemplate for JSON payload/markdown generation. Now the subtemplates can be referred in the Notification.Body.

+ Additional template.payload section  processed as text/plain, instead of text/html. So, other markup languages can be used.
+ Processing of Notification.body template has additional Vars available (Alert, Incident, Notification).


gbrayut:
Some of the files need gofmt run. See https://travis-ci.org/bosun-monitor/bosun/builds/196181334#L192

PR #2011, opened 2017-02-02
cmd/scollector: Updated Azure EA collector to add new tags.

mhenderson-so:
Also converted to better invocation style.

### DIFF
--- a/cmd/scollector/doc.go
+++ b/cmd/scollector/doc.go
@@ -252,12 +252,23 @@ LogBillingDetails tells scollector to add the following tags to your metrics:
    - costcenter
 	 - accountname
 	 - subscription
+
+LogResourceDetails tell scollector to add the following tags to your metrics:
+   - resoucegroup
+	 - resourcelocation
+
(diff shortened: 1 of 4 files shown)


mhenderson-so:
I re-pushed this PR, as it never got merged when I originally submitted it, because the TravisCI job was borked back then. I haven't actually changed anything between now and then - but now the checks pass.

cmd/scollector/doc.go
```diff
@@ -252,12 +252,23 @@ LogBillingDetails tells scollector to add the following tags to your metrics:
...
+are not careful. It will not process any tags that begin with "hidden".
+
 If you are a heavy Azure EA user, then these additional tags may be useful for breaking down costs.
 
 	[[AzureEA]]
 	  EANumber = "123456"
 	  APIKey = "joiIiwiaXNzIjoiZWEubWljcm9zb2Z0YXp1cmUuY29tIiwiYXVkIjoiY2xpZW50LmVhLm1"
 	  LogBillingDetails = false
+	  LogResourceDetails = false
+    LogExtraTags = false
```
captncraig: I think spacing is off just a tad here.

Issue #1662, opened 2016-03-11
Dashboard loads bodies of all alerts, and then lazy loads them


kylebrandt:
In order to speed up the dashboard, the body of alerts are lazy loaded via the `/api/status`. I believe this does help dashboard loading time since all of this isn't put into the DOM on load.

However, the initial get to /api/alerts includes the bodys and attachments for all alerts (which right now is about 3MByte compressed). This makes dashboard loads take several seconds.

kylebrandt:
This was changed so at least we don't marshal the data and send it to the client. So the file size is around 50k now. However, we still read all this from redis which takes about 1.5 seconds.

So this is *better*, but not "fixed" yet.

Issue #1952, opened 2016-11-04
WIP: Improving HTTP Post notifications in Bosun


kylebrandt:
Currently working on spec'ing this. Will probably end up in 0.7 (vNextNext - after auth). The Google doc is view only:

https://docs.google.com/document/d/1e9mhfD1aLR78sxxJXL2X7T3ujGPFX4-GzAyCdOVwXzY/edit?pli=1# 

But feedback and thoughts can be left here if you like.

ayashjorden:
Testing notifications feature is very important and can save a lot of dev time.
So sending the POST/GET notification to the configured endpoint and not just display the payload will serve us better

See the [Slack](https://bosun.slack.com/archives/general/p1478645232002792) message.

angadsingh:
just went through this doc after already having commented here: https://github.com/bosun-monitor/bosun/pull/1737#issuecomment-303719268

suggestions/ideas:

* instead of opening up the template definition completely with arbitrary variables, how about having "template types" for each kind of notification. please read the complete suggestion here: https://github.com/bosun-monitor/bosun/pull/1737#issuecomment-303719268
* why can't bosun "resolve" the pagerduty incident when the bosun incident's severity decreases to normal (i.e. recovery notifications). This would go a long way in reducing alerts and alert fatigue - right now we need to set a low (1 hour) timeout in the notification chain (self) so that it retriggers. That's because we are using the repeated trigger as a way to come to know when the incident resolves, causing high number of repeats for the same incident.
* we do have a reverse integration written for PD to bosun (i.e. ack/resolve/close bosun incident when a human acks/resolves it in PD). if its alright we can share it with the bosun project. it has both incoming webhook and polling based (if you're behind a firewall) integration. it currently works with sending the `subject` of the alert as the PD `incident_key` which is then parsed back to get the bosun incident ID.
* if we're worried too much about template repetition, apart from simple template inclusion using {{template "name"}}, golang templates also supports template inheritance via "blocks" - and it does work in bosun, albeit with a catch. e.g. 

`template base {
     subject = ```This is a template which can support inheritance. {{block "alpha"}}This is the default text{{end}}. My other text.```
}

template child {
    subject = ```{{define "alpha"}}This is the replaced text{{end}}{{template "base" .}}```
}
`

The catch is that bosun seems to be using the same template context across alerts, so when there are multiple alerts using the same "base" template, the child override overrides for them too

kylebrandt:
[...]

PR #2016, opened 2017-02-11
wip: start to attempt to migrate rendered templates out of data structure

kylebrandt:
Lots in these changes so need to rest and look at it with a fresh page of eyes.

Problem right now is that when expanding things on the dashboard, sometimes there are templates, and other times there are not.

The following seems not right:

```
$ redis-cli -p 6389 LRANGE allIncidents 0 -1 | wc -l
   71077

# kbrandt at grove in ~/src/bosun.org/cmd/bosun on git:rendTempStore x [16:03:15]
$ redis-cli -p 6389 KEYS incidentById:\* | wc -l
   64492
```

Also maybe something to do with not storing when emailbody == body? 

Anyways, need a break. The dashboard loads much faster though. Even with a handful of alerts the "Setup" phase in marshal groups goes down by a factor of 10 (30 vs 300ms)

### DIFF
--- a/cmd/bosun/database/migrate.go
+++ b/cmd/bosun/database/migrate.go
@@ -112,7 +112,11 @@ func migrateRenderedTemplates(d *dataAccess) error {
 			LastAbnormalStatus: oldState.LastAbnormalStatus,
 			LastAbnormalTime:   oldState.LastAbnormalTime,
 		}
-		if _, err := d.State().UpdateIncidentState(newState); err != nil {
+		data, err := json.Marshal(newState)
+		if err != nil {
+			return slog.Wrap(err)
+		}
+		if _, err := conn.Do("SET", incidentStateKey(newState.Id), data); err != nil {
 			return slog.Wrap(err)
 		}
 	}
--- a/cmd/bosun/database/migrate.go
+++ b/cmd/bosun/database/migrate.go
@@ -21,7 +21,7 @@ type Migration struct {
 }
 
 var tasks = []Migration{
-	Migration{
+	{
 		UID:     "Migrate Rendered Templates",
 		Task:    migrateRenderedTemplates,
 		Version: 1,
--- a/cmd/bosun/database/migrate.go
+++ b/cmd/bosun/database/migrate.go
@@ -6,7 +6,7 @@ import (
 
 	"bosun.org/models"
 	"bosun.org/slog"
-	"github.com/beego/redigo/redis"
+	"github.com/garyburd/redigo/redis"
 )
 
 // Version 0 is the schema that was never verisoned
--- a/cmd/bosun/database/migrate.go
+++ b/cmd/bosun/database/migrate.go
@@ -104,16 +104,17 @@ func (d *dataAccess) Migrate() error {
 	}
 
 	for _, task := range tasks {
-		if task.Version > version { // Not sure logic is correct don't care right now.
+		if task.Version > version {
 			// Check if migration has been run if not that run
 			err := task.Task(d)
 			if err != nil {
 				return slog.Wrap(err)
 			}
+			if _, err := conn.Do("SET", schemaKey, task.Version); err != nil {
+				return slog.Wrap(err)
+			}
 		}
-		if _, err := conn.Do("SET", schemaKey, SchemaVersion); err != nil {
-			return slog.Wrap(err)
-		}
+
 	}
 	return nil
 }
--- a/cmd/bosun/database/migrate.go
+++ b/cmd/bosun/database/migrate.go
@@ -1,0 +1,154 @@
+package database
+
+import (
+	"encoding/json"
+	"time"
+
+	"bosun.org/models"
+	"bosun.org/slog"
+	"github.com/beego/redigo/redis"
+)
+
+// Version 0 is the schema that was never verisoned
+// Version 1 migrates rendered templates from
+
+var schemaKey = "schemaVersion"
+
+type Migration struct {
+	UID     string
+	Task    func(d *dataAccess) error
+	Version int64
+}
+
+var tasks = []Migration{
+	Migration{
+		UID:     "Migrate Rendered Templates",
+		Task:    migrateRenderedTemplates,
+		Version: 1,
+	},
+}
+
+type oldIncidentState struct {
+	Id       int64
+	Start    time.Time
+	End      *time.Time
+	AlertKey models.AlertKey
+	Alert    string // helper data since AlertKeys don't serialize to JSON well
+	Tags     string // string representation of Group
+
+	*models.Result
+
+	// Most recent last.
+	Events  []models.Event  `json:",omitempty"`
+	Actions []models.Action `json:",omitempty"`
+
+	Subject      string
+	Body         string
+	EmailBody    []byte
+	EmailSubject []byte
+	Attachments  []*models.Attachment
+
+	NeedAck bool
+	Open    bool
+
+	Unevaluated bool
+
+	CurrentStatus models.Status
+	WorstStatus   models.Status
+
+	LastAbnormalStatus models.Status
+	LastAbnormalTime   int64
+}
+
+func migrateRenderedTemplates(d *dataAccess) error {
+	slog.Infoln("Running rendered template migration")
+
+	ids, err := d.getAllIncidentIds()
+	if err != nil {
+		return err
+	}
+
+	conn := d.Get()
+	defer conn.Close()
+
+	for _, id := range ids {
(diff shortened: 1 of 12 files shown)


kylebrandt:
Or 290ms vs 4.5sec for a hundreds of alerts. So if the dashboard is flood this reduces the load time a lot. Also that several seconds that redis isn't being blocked since the bottleneck here was fetching data from redis that did not need to be fetched.

kylebrandt:
should eventually fix #1662

kylebrandt:
Also want to get this done before https://github.com/bosun-monitor/bosun/issues/1952 , as that will only increase the size of the rendered templates.

cmd/bosun/database/state_data.go
```diff
@@ -51,16 +59,51 @@ type StateDataAccess interface {
 	GetLatestIncident(ak models.AlertKey) (*models.IncidentState, error)
 	GetAllOpenIncidents() ([]*models.IncidentState, error)
 	GetIncidentState(incidentId int64) (*models.IncidentState, error)
-	GetAllIncidents(ak models.AlertKey) ([]*models.IncidentState, error)
+	//GetAllIncidentIds() ([]int64, error)
```
kylebrandt: remove commented out func

cmd/bosun/database/state_data.go
```diff
@@ -150,6 +193,45 @@ func (d *dataAccess) GetAllIncidents(ak models.AlertKey) ([]*models.IncidentStat
...
+		var err error
+		ids[i], err = strconv.ParseInt(strings.Split(sum, ":")[1], 0, 64)
+		if err != nil {
+			return nil, slog.Wrap(err)
+		}
+	}
+	return ids, nil
+
+}
+
```
kylebrandt: resolve descrepency between getAllIncidentIds and getAllIncideIdsByKeys ?

cmd/bosun/sched/check.go
```diff
@@ -324,15 +336,16 @@ func (s *Schedule) executeTemplates(state *models.IncidentState, event *models.E
 			attachments = nil
 		}
 		state.Subject = string(subject)
-		state.Body = string(body)
+		//state.Body = string(body)
```
kylebrandt: remove comment

cmd/bosun/database/migrate.go
```diff
@@ -0,0 +1,165 @@
...
+	LastAbnormalStatus models.Status
+	LastAbnormalTime   int64
+}
+
+func migrateRenderedTemplates(d *dataAccess) error {
+	slog.Infoln("Running rendered template migration")
+
+	// Hacky Work better?
+	ids, err := d.getAllIncidentIdsByKeys()
+	slog.Infoln("migrating %v incidents", len(ids))
```
kylebrandt: change to format func

kylebrandt: add comment to long entry that this can take several minutes.

cmd/bosun/sched/sched.go
```diff
@@ -100,6 +100,9 @@ func (s *Schedule) Init(systemConf conf.SystemConfProvider, ruleConf conf.RuleCo
 	if s.Search == nil {
 		s.Search = search.NewSearch(s.DataAccess, skipLast)
 	}
+	if err := s.DataAccess.Migrate(); err != nil {
```
kylebrandt: this might be the wrong place to call this.

kylebrandt:
seems incidents are disappearing after the migration... hrm...

kylebrandt:
not good to have backing up Last redis data structure and migration going at same time I think.

captncraig:
Looks like the 'allIncidents' key is only ever appended to. The count gets off when we purge/forget.

cmd/bosun/database/migrate.go
```diff
@@ -0,0 +1,164 @@
...
+var tasks = []Migration{
+	{
+		UID:     "Migrate Rendered Templates",
+		Task:    migrateRenderedTemplates,
+		Version: 1,
+	},
+}
+
+type oldIncidentState struct {
+	Id       int64
```
captncraig:
```
type oldIncidentState struct{
 *models.IncidentState
 Body string 
 ...
}
```

captncraig: Then marshall `i.IncidentState`

cmd/bosun/database/state_data.go
captncraig: Really we should just remove allIncidents altogether. It is not updated when you remove incidents. It does not look like we ever read from it. And you have shown we can get the incident data the hard way (with KEYS) if we need to.

cmd/bosun/sched/check.go
```diff
@@ -125,8 +126,18 @@ func (s *Schedule) runHistory(r *RunHistory, ak models.AlertKey, event *models.E
...
+			if err != nil {
+				return
+			}
+			err = data.SetRenderedTemplates(incident.Id, rt)
+			if err != nil {
+				return
+			}
 		} else {
 			err = data.SetUnevaluated(ak, event.Unevaluated) // if nothing to save, at least store the unevaluated state
+			if err != nil {
```
captncraig: This error checking block is a noop if there is no code following. Same thing directly above this block too.

cmd/bosun/sched/check.go
```diff
@@ -183,7 +194,8 @@ func (s *Schedule) runHistory(r *RunHistory, ak models.AlertKey, event *models.E
 
 	//render templates and open alert key if abnormal
 	if event.Status > models.StNormal {
-		s.executeTemplates(incident, event, a, r)
+		s.executeTemplates(incident, rt, event, a, r)
+		s.DataAccess.State().SetRenderedTemplates(incident.Id, rt)
```
captncraig: This happens in the defer as well. No need to set it twice.

cmd/bosun/sched/sched.go
captncraig: Yes, I would do it from main if possible.

models/incidents.go
```diff
@@ -22,11 +22,9 @@ type IncidentState struct {
...
 	Actions []Action `json:",omitempty"`
 
-	Subject      string
-	Body         string
-	EmailBody    []byte
-	EmailSubject []byte
-	Attachments  []*Attachment
+	Subject string
+
+	rt *RenderedTemplates
```
captncraig: remove this field

cmd/bosun/database/migrate.go
```diff
@@ -0,0 +1,112 @@
...
+				version = SchemaVersion
+				return nil
+			}
+		} else {
+			return slog.Wrap(err)
+		}
+	}
+
+	for _, task := range tasks {
+		if task.Version > version { // Not sure logic is correct don't care right now.
```
kylebrandt: @captncraig Can you think through my logic for future versions and see if this is sane?

kylebrandt:
Note: Migration for our (Stack Overflow) redis instance of 65436 incidents takes 2 min and 34 seconds on my workstation. Bosun is not accessible during the migration.

kylebrandt:
Hrmm.. Still bad, seeing incidents with no body.

cmd/bosun/database/migrate.go
```diff
@@ -0,0 +1,112 @@
...
+			// Check if migration has been run if not that run
+			err := task.Task(d)
+			if err != nil {
+				return slog.Wrap(err)
+			}
+		}
+		if _, err := conn.Do("SET", schemaKey, SchemaVersion); err != nil {
```
captncraig: should set to task.Version, not schemaVersion.

cmd/bosun/database/state_data.go
```diff
@@ -13,6 +16,8 @@ import (
 /*
 incidentById:{id} - json encoded state. Authoritative source.
 
+renderedTemplatesById:{id} - json encoded RenderedTemplates ny Incident Id
```
captncraig: typo

cmd/bosun/database/state_data.go
```diff
@@ -150,6 +192,45 @@ func (d *dataAccess) GetAllIncidents(ak models.AlertKey) ([]*models.IncidentStat
 	return d.incidentMultiGet(conn, ids)
 }
 
+func (d *dataAccess) getAllIncidentIds() ([]int64, error) {
```
captncraig: Why are these two functions both needed? Didn't you say you stopped using on or the other?

cmd/bosun/database/migrate.go
kylebrandt: In which case I should set it in the loop after each task is successful then?

captncraig: yeah

PR #2025, opened 2017-02-17
Check to ensure no two consecutive check runs share a timestamp.

### LINKED ISSUE #2023: Checks that run at CheckFrequency*1 may run twice with same now
As discovered in https://github.com/bosun-monitor/bosun/issues/2019#issuecomment-280719214

The following race may cause checks to run with the "earlier" timestamp:

```
==================
WARNING: DATA RACE
Read at 0x000001a4d288 by goroutine 97:
  bosun.org/cmd/bosun/sched.(*Schedule).checkAlert()
      /Users/kbrandt/src/bosun.org/cmd/bosun/sched/alertRunner.go:68 +0x5f
  bosun.org/cmd/bosun/sched.(*Schedule).RunAlert()
      /Users/kbrandt/src/bosun.org/cmd/bosun/sched/alertRunner.go:55 +0x17e

Previous write at 0x000001a4d288 by goroutine 94:
  bosun.org/cmd/bosun/sched.(*Schedule).updateCheckContext()
      /Users/kbrandt/src/bosun.org/cmd/bosun/sched/alertRunner.go:32 +0x32d

Goroutine 97 (running) created at:
  bosun.org/cmd/bosun/sched.(*Schedule).Run()
      /Users/kbrandt/src/bosun.org/cmd/bosun/sched/alertRunner.go:21 +0x208
  bosun.org/cmd/bosun/sched.Run()
      /Users/kbrandt/src/bosun.org/cmd/bosun/sched/sched.go:493 +0x4a
  main.main.func5()
      /Users/kbrandt/src/bosun.org/cmd/bosun/main.go:230 +0x6c

Goroutine 94 (running) created at:
  bosun.org/cmd/bosun/sched.(*Schedule).Run()
      /Users/kbrandt/src/bosun.org/cmd/bosun/sched/alertRunner.go:19 +0x134
  bosun.org/cmd/bosun/sched.Run()
      /Users/kbrandt/src/bosun.org/cmd/bosun/sched/sched.go:493 +0x4a
  main.main.func5()
      /Users/kbrandt/src/bosun.org/cmd/bosun/main.go:230 +0x6c
==================
```

As demonstrated in the following log:

```
2017/02/17 17:43:44 info: check.go:529: check alert sites.readonly.mode start with now set to 2017-02-17 17:43:44.690531283
2017/02/17 17:44:44 info: check.go:529: check alert sites.readonly.mode start with now set to 2017-02-17 17:43:44.690531283
2017/02/17 17:45:44 info: check.go:529: check alert sites.readonly.mode start with now set to 2017-02-17 17:44:44.789880559
2017/02/17 17:46:44 info: check.go:529:
```

captncraig:
Uses a simple counter that gets incremented every time we make a new check context. When a check runs, if the context has not changed since last time, we wait a second and look again. 

Also fixes race condition.

Fixes #2023

### DIFF
--- a/cmd/bosun/sched/alertRunner.go
+++ b/cmd/bosun/sched/alertRunner.go
@@ -66,6 +66,8 @@ func (s *Schedule) RunAlert(a *conf.Alert) {
 				lastCheckID = ctx.id
 				break
 			}
+			//if the timing is just wrong, we could hit this twice in the same interval. One second wait should be enough
+			//to desync us from the context changing routine.
 			collect.Add("check.context_not_changed", opentsdb.TagSet{"alert": a.Name}, 1)
 			time.Sleep(time.Second)
 		}
--- a/cmd/bosun/sched/alertRunner.go
+++ b/cmd/bosun/sched/alertRunner.go
@@ -2,10 +2,13 @@ package sched
 
 import (
 	"fmt"
+	"sync"
 	"time"
 
 	"bosun.org/cmd/bosun/cache"
 	"bosun.org/cmd/bosun/conf"
+	"bosun.org/collect"
+	"bosun.org/opentsdb"
 	"bosun.org/slog"
 )
 
@@ -23,10 +26,16 @@ func (s *Schedule) Run() error {
 	return nil
 }
 
+var checkContextLock = sync.RWMutex{}
+
 func (s *Schedule) updateCheckContext() {
+	var checkID int64
 	for {
-		ctx := &checkContext{utcNow(), cache.New(0)}
+		ctx := &checkContext{utcNow(), cache.New(0), checkID}
+		checkID++
+		checkContextLock.Lock()
 		s.ctx = ctx
+		checkContextLock.Unlock()
 		time.Sleep(s.SystemConf.GetCheckFrequency())
 		s.Lock("CollectStates")
 		s.CollectStates()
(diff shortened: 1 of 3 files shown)


cmd/bosun/sched/alertRunner.go
```diff
@@ -39,15 +48,30 @@ func (s *Schedule) RunAlert(a *conf.Alert) {
...
+			checkContextLock.RLock()
+			ctx = s.ctx
+			checkContextLock.RUnlock()
+			//make sure the context has actually changed
+			if ctx.id != lastCheckID {
+				lastCheckID = ctx.id
+				break
+			}
+			collect.Add("check.context_not_changed", opentsdb.TagSet{"alert": a.Name}, 1)
+			time.Sleep(time.Second)
```
kylebrandt: maybe this should be some fraction of the checkFrequency?

captncraig: I initially chose a millisecond. If this occurs at all, it is because the alert and the global context are perfectly lock stepped, and this one the race. A second should be more than adequate to "unlock" it in a timely manner. 

If the counter explodes, we have a bigger problem, and some bad assumptions.

kylebrandt:
What about something like instead of the routines sleeping on time, a timer tells them when they can run. That way the timer can always set the context before sending the signal

captncraig:
I thought about something like that, maybe using [sync.Cond](https://golang.org/pkg/sync/#Cond) as a signal. But I think there is an additional race there. 

If the context is updated and signalled between when the alert detects no-change, and when it subscribes to the signal, you miss a whole check interval. My way you only lose a second.

cmd/bosun/sched/alertRunner.go
kylebrandt: should comment in the code on why were are doing the sleep

cmd/bosun/sched/alertRunner.go
```diff
@@ -39,15 +48,32 @@ func (s *Schedule) RunAlert(a *conf.Alert) {
 	s.checksRunning.Add(1)
 	// ensure when an alert is done it is removed from the wait group
 	defer s.checksRunning.Done()
+	var lastCheckID int64 = -1
```
kylebrandt: The checkId and lastCheckID thing is confusing me

captncraig: each alert tracks the last id that it ran a check on. we actually could dispense with the field and just track lastCheckTime in the same way. Good point.

captncraig: how's that look now?

captncraig:
Kyle and I had a better idea in hangouts. will implement that instead.

PR #2027, opened 2017-02-21
Empty set depends

kylebrandt:


cmd/bosun/conf/rule/rule.go
```diff
@@ -630,7 +630,7 @@ func (c *Conf) loadAlert(s *parse.SectionNode) {
 		if err != nil {
 			c.error(err)
 		}
-		if len(depTags.Intersection(tags)) < 1 {
+		if len(depTags.Intersection(tags)) < 1 && len(depTags) != 0 {
```
captncraig: I'd switch the order here.

cmd/bosun/sched/check.go
```diff
@@ -614,7 +614,7 @@ func markDependenciesUnevaluated(events map[models.AlertKey]*models.Event, deps
 			continue
 		}
 		for _, dep := range deps {
-			if dep.Group.Overlaps(ak.Group()) {
+			if dep.Group.Overlaps(ak.Group()) || len(dep.Group) == 0 {
```
captncraig: Switch conditions again.

PR #2028, opened 2017-02-21
cmd/scollector: enable SWbemServices worker for better WMI performance in Windows

gbrayut:
This updates the WMI package to include the SWbemServices updates from https://github.com/StackExchange/wmi/pull/23

I plan on testing this in our environment to make sure things are stable. Shouldn't cause any issues as everything is behind the UseSWbemServicesClient = true value in scollector.toml


gbrayut:
We didn't find any issues when testing, and this has been rolled out to all our Windows servers now. The default is still using the old code, you still have to specify UseSWbemServicesClient = true to use the new code. In the future we may change that to be the default.

PR #2029, opened 2017-02-24
Letting scheduler trigger alerts from central place.

captncraig:
I hate this much less


kylebrandt:
sched test needs fixing

PR #2033, opened 2017-03-08
Add option to store SNMP IfAlias as a tag

rwolfson:
This PR adds an option to scollector SNMP interface statistics, to add an additional tag for ifAlias. This can be useful for OpenTSDB installations that don't have metadata enabled, to allow interface "descriptions" to still be utilized in charts.

This also changes ifAlias sanitization by replacing (invalid) spaces with (valid) underscores, to improve readability.

### DIFF
--- a/cmd/scollector/collectors/snmp_ifaces.go
+++ b/cmd/scollector/collectors/snmp_ifaces.go
@@ -40,7 +40,7 @@ const (
 func SNMPIfaces(cfg conf.SNMP) {
 	collectors = append(collectors, &IntervalCollector{
 		F: func() (opentsdb.MultiDataPoint, error) {
-			return c_snmp_ifaces(cfg.Community, cfg.Host)
+			return c_snmp_ifaces(cfg.Community, cfg.Host, cfg.TagIfAlias)
 		},
 		Interval: time.Second * 30,
 		name:     fmt.Sprintf("snmp-ifaces-%s", cfg.Host),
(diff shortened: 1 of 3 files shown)


cmd/scollector/collectors/snmp_ifaces.go
```diff
@@ -103,9 +103,11 @@ func c_snmp_ifaces(community, host string) (opentsdb.MultiDataPoint, error) {
 		ifTypes[k] = val
 	}
 	for k, v := range ifAliasesRaw {
+		// Replace spaces with underscores for readability, then clean bad characters
+		newAlias := strings.Replace(fmt.Sprintf("%s", v), " ", "_", -1)
```
captncraig: Were these aliases ever being used in tags before? If so, I would be concerned we are changing tagsets currently in use. If not, this looks pretty good to me.

rwolfson: It appears that they used to be, up until they were switched to metadata in 2014 in 665993dcbab0800dc0a5aef8fea47a30a613b651. In the current version of scollector, aliases are only present in metadata. I could add an additional toggle to run this space-to-underscore transform only when set, but anyone upgrading past the older version that had them as tags would see a change in tag behavior/values regardless given that they were later taken away. The immediate impact this change would have on live deployments running an incremental upgrade would be a change in metadata string formatting. I'm happy to support whatever direction you prefer here, just let me know.

rwolfson:
Hi,
Just wanted to check in on this to see where things stand with this PR. Thanks!

Issue #2018, opened 2017-02-14
tsdbrelay: Rapid socket use results in trouble


alienth:
tsdbrelay creates a new socket connection whenever it is chatting with bosun, and closes it immediately. This results in a large number of sockets stuck in TIME_WAIT. On our production systems, this often causes issues where the ports which hbase or opentsdb typically use are very frequently being consumed by ephemeral ports from sockets in TIME_WAITs, which causes issues on server startup.

Would be nice if tsdbrelay would set `SO_REUSEADDR` on socket creation so that it reuses some of those TIME_WAITs rather than eating up a new ephemeral port.

alienth:
Main issue here is probably not socket creation, but lack of http connection reuse. This will be partially addressed in #2034.

kylebrandt:
Assuming fixed enough by the issues unless we run into future problems.

PR #2034, opened 2017-03-12
cmd/tsdbrelay: Drain relay response so that connection is eligible for reuse.

alienth:
tsdbrelay is not reusing connections for its relay calls. `DefaultTransport` will not allow a connection to be reused unless the [body has been read](https://golang.org/src/net/http/transport.go?h=bodyEOFSignal#L2002). This change will use the common pattern of reading 512 bytes from the `resp.Body` to try and ensure the connection is available for reuse. This is intentionally capped at 512 bytes to avoid waiting for cases of large responses.

Some discussion on this if curious: https://groups.google.com/forum/#!search/golang$20json$20decode$20$20io.Copy/golang-nuts/4Rr8BYVKrAI/ZrJJFTNleekJ

Confirmed in testing that this addresses lack of connection reuse.

Partially addresses #2018.

@kylebrandt @captncraig @gbrayut

### DIFF
--- a/cmd/tsdbrelay/main.go
+++ b/cmd/tsdbrelay/main.go
@@ -8,6 +8,7 @@ import (
 	"flag"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
 	"net/http/httputil"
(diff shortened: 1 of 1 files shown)


gbrayut:
The code has separate paths for bosun, relays, and metadata... do we need the same io.CopyN added to these:

https://github.com/alienth/bosun/blob/8b2135b1a85f9dd075ae222482ca3cbf0cd7cb8f/cmd/tsdbrelay/main.go#L253

https://github.com/alienth/bosun/blob/8b2135b1a85f9dd075ae222482ca3cbf0cd7cb8f/cmd/tsdbrelay/main.go#L389

alienth:
@gbrayut updated.

gbrayut:
Found some more cases in metadata/collect/etc where this occurs. This PR is superseded by #2036

PR #2036, opened 2017-03-13
misc: Drain body responses so that connection is eligible for reuse

gbrayut:
After Jason updated tsdbrelay I did an audit of the various places we use Body.Close but don't read any bytes. This should allow connection reuse in those cases. Testing on ny-gbraylx01 but would like someone to review @captncraig @alienth @kylebrandt

### DIFF
--- a/cmd/scollector/collectors/interval.go
+++ b/cmd/scollector/collectors/interval.go
@@ -1,6 +1,7 @@
 package collectors
 
 import (
+	"io"
 	"io/ioutil"
 	"net/http"
 	"os/exec"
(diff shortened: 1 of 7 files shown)


cmd/scollector/collectors/interval.go
```diff
@@ -111,6 +112,10 @@ func enableURL(url string, regexes ...string) func() bool {
 			return false
 		}
 		defer resp.Body.Close()
+		if resp.StatusCode != 200 || len(res) == 0 {
```
alienth: Could just batch this up in an anonymous defer func rather than having this awkward condition here that mirrors future conditions. Since we ignore errors intentionally from `CopyN` in this case, doesn't matter if a `ReadAll` has already occurred.

Just a nitpick, tho. Fine as is if you'd prefer.

collect/queue.go
```diff
@@ -138,6 +139,8 @@ func sendBatch(batch []*opentsdb.DataPoint) {
 		time.Sleep(d)
 		return
 	}
+	// Drain up to 512 bytes and close the body to let the Transport reuse the connection
```
alienth: Comment here is semi-erroneous since we don't exactly close the body here.

alienth:
@gbrayut A couple nitpick comments, otherwise lgtm.

PR #2041, opened 2017-03-22
Switch to font-awesome and add more help links

kylebrandt:


kylebrandt:
@captncraig some of the libraries references them so I figured I would just leave them.

PR #2046, opened 2017-03-27
cmd/bosun: call annotate backend code directly

kylebrandt:


cmd/bosun/main.go
```diff
@@ -140,8 +141,26 @@ func main() {
...
+			for {
+				err := annotateBackend.InitBackend()
+				if err == nil {
+					return
+				}
+				slog.Warningf("could not initialize annotate backend, will try again: %v", err)
+				time.Sleep(time.Second * 30)
+			}
+		}()
+		web.AnnotateBackend = annotateBackend
```
kylebrandt: Yes, all the calls return an error:

```
	if !e.initialized {
		return nil, unInitErr
	}
```

in the backend code.

PR #2053, opened 2017-04-03
cmd/scollector: support auth token

kylebrandt:


cmd/scollector/main.go
```diff
@@ -50,14 +50,18 @@ var (
...
+	XAccessToken string
 	http.RoundTripper
 }
 
 func (t *scollectorHTTPTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 	if req.Header.Get("User-Agent") == "" {
 		req.Header.Add("User-Agent", t.UserAgent)
 	}
+	if t.XAccessToken != "" {
+		req.Header.Add("X-Access-Token", t.XAccessToken)
```
captncraig: We already have an AuthToken in the collect package. Should be as easy as setting that from the scollector config.

https://github.com/bosun-monitor/bosun/blob/master/collect/collect.go#L63

cmd/scollector/conf/conf.go
```diff
@@ -50,6 +50,10 @@ type Conf struct {
 	// UseNtlm specifies if HTTP requests should authenticate with NTLM.
 	UseNtlm bool
 
+	// XAccessToken is an optional string that sets the X-Access-Token HTTP header
+	// which is used to authenticate against Bosun
+	XAccessToken string
```
captncraig: Lets call this `AuthToken` perhaps? Looks like maybe we've been a tad inconsistent.

cmd/scollector/main.go
kylebrandt: Should I follow the same pattern in the metadata package? Create a global var?

captncraig: yeah. It may be there already too.

PR #2055, opened 2017-04-04
cmd/bosun: (wip) define rule vars in sysconfig

kylebrandt:


cmd/bosun/web/web.go
```diff
@@ -840,7 +840,7 @@ func ConfigTest(t miniprofiler.Timer, w http.ResponseWriter, r *http.Request) (i
 	if len(b) == 0 {
 		return nil, fmt.Errorf("empty config")
 	}
-	_, err = rule.NewConf("test", schedule.SystemConf.EnabledBackends(), string(b))
+	_, err = rule.NewConf("test", schedule.SystemConf.EnabledBackends(), nil, string(b))
```
captncraig: This could potentially cause problems with the config not validating because it is missing the system portion?

kylebrandt: Good catch

PR #2059, opened 2017-04-11
cmd/tsdbrelay: parse -t -b and -r parameters as URLs and allow filtering of relay data

gbrayut:
This is a first try at adding support for HTTPS in -t, -b, and -r parameters in tsdbrelay. I also rigged up the url fragment to allow filtering of what data gets sent to a relay. If you add `#data-only` or `#metadata-only` to the end of a host in the -r parameter section it should filter the relay results sent to that host. This is useful when the remote endpoint is NOT a load balancer or tsdbrelay instance, but instead a native OpenTSDB compatible endpoint

I haven't started testing yet, but let me know if there are any objections/recommended changes

### DIFF
--- a/cmd/tsdbrelay/doc.go
+++ b/cmd/tsdbrelay/doc.go
@@ -22,15 +22,16 @@ Usage:
 
 The flags are:
 	-b="bosun"
-		Target Bosun server. Can specify port with host:port.
+		Target Bosun server. Can specify as host, host:port, or https://host:port.
 	-t=""
-		Target OpenTSDB server. Can specify port with host:port.
+		Target OpenTSDB server. Can specify as host, host:port or https://host:port.
 	-l=":4242"
 		Listen address.
 	-v=false
 	    Enable verbose logging
 	-r=""
 		Additional relays to send data to, comma seperated. Intended for secondary data center replication. Only response from primary tsdb server wil be relayed to clients.
+		Examples: hostA:port,https://hostB:port,hostC#data-only,https://hostD:8080#metadata-only
 	-redis=""
 		Redis host to store external counter data in
 	-db=0
--- a/cmd/tsdbrelay/main.go
+++ b/cmd/tsdbrelay/main.go
@@ -48,7 +48,8 @@ var (
 
 	denormalizationRules map[string]*denormalize.DenormalizationRule
 
-	relayPutUrls []string
+	relayDataUrls     []string
+	relayMetadataUrls []string
 
 	tags = opentsdb.TagSet{}
 )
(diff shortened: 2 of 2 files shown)


gbrayut:
@kylebrandt does tsdbrelay need to support authentication tokens? If so I can try adding that in here as well.

captncraig:
We've talked about that. To really do auth properly it would need to read and validate the tokens. Currently that would mean reading from bosun's redis db, and we really don't want to make that dependency.

We may wish to ensure it is correctly **passing** the tokens given to it though.

cmd/tsdbrelay/doc.go
```diff
@@ -22,15 +22,16 @@ Usage:
 
 The flags are:
 	-b="bosun"
-		Target Bosun server. Can specify port with host:port.
+		Target Bosun server. Can specify as host, host:port, or https://host:port.
 	-t=""
-		Target OpenTSDB server. Can specify port with host:port.
+		Target OpenTSDB server. Can specify as host, host:port or https://host:port.
```
captncraig: do we allow `http://host:port` as well? Should we?

gbrayut:
ah... right. It should already be passing existing headers (grep for accessHeader)

cmd/tsdbrelay/doc.go
```diff
@@ -22,15 +22,16 @@ Usage:
...
 	-l=":4242"
 		Listen address.
 	-v=false
 	    Enable verbose logging
 	-r=""
 		Additional relays to send data to, comma seperated. Intended for secondary data center replication. Only response from primary tsdb server wil be relayed to clients.
+		Examples: hostA:port,https://hostB:port,hostC#data-only,https://hostD:8080#metadata-only
```
captncraig: The fragment thing feels super dirty. But I can't think of a better way immediately. Are we outgrowing the capacity to configure this thing without a more complicated config?

```
{
"primaryTsdb": "ny-tsdb03",
"primaryBosun": "ny-bosun01",
"relays":[
  {"type":"tsdb", "host":"ny-tsdbcluster2"},
  {"type":"tsdbrelay","host":"co-tsdb01"},
  {"type":"bosun","host":"ny-bosunxy42"}
 ]
}
```

Is starting to be more appealing to me than super-dense flags. Or whatever the toml equivalent is.

Something to think about.

cmd/tsdbrelay/doc.go
gbrayut: we do... if protocol is omitted it defaults to http. I yanked the parsing function we use in Scollector/bosun

cmd/tsdbrelay/doc.go
gbrayut: This is currently driven by command line flags. It is a bit of a hack, but one that only takes a few lines and still looks pretty clean. Moving to a full config file is another option, but would be a bigger refactor and not be backwards compatible.

captncraig: Yeah, it passes the "gets it done" test, but I worry we are running out of places to hide more metadata if we have more things to toggle.

gbrayut: my vote is hack it in via fragments now, but +1 on a better long term solution if we have a need for a "v2" breaking-changes release in the future.

Do we currently know of any other per host settings that need toggling?

Issue #2061, opened 2017-04-12
When saving bosun config file, leaving message field blank causes "exit status 1"


PeterGrace:
In the "Save" dialog in the rule editor, if you do not specify a message in the message box, the save operation will fail with the cryptic error: `failed to call save hook: exit status 1. Restoring config: successful`

Three options:
 - the save dialog should not permit the user to click save without a message, 
 - the save dialog should produce a default message in that field if none is specified,
 - the error should be more specific that message is malformed or missing.

kylebrandt:
This is part bug and part enhancement. Reading the code for SaveRawText on Conf, my intent was to pass a message from the savehook command to the user via stderr, but somewhere that is broken so I'm trying to trace that down.

As far as changing the UI before you hit the button, that is a new thing I can added with some config options.

kylebrandt:
Will be fixed in #2183

PR #2063, opened 2017-04-18
cmd/bosun: pass stderr on error to user for save command

kylebrandt:


kylebrandt:
This takes care of the bug part of https://github.com/bosun-monitor/bosun/issues/2061

PR #2094, opened 2017-05-31
cmd/bosun: expr func to add tags based on metadata

kylebrandt:


docs/expressions.md
```diff
@@ -510,7 +510,20 @@ Group functions modify the OpenTSDB groups.
 ## addtags(set variantSet, group string) (seriesSet|numberSet)
 {: .exprFunc}
 
-Accepts a series and a set of tags to add to the set in `Key1=NewK1,Key2=NewK2` format. This is useful when you want to add series to set with merge and have tag collisions.
+Accepts a set and a set of tags to add to the set in `Key1=NewK1,Key2=NewK2` format. This is useful when you want to add series to set with merge and have tag collisions.
+
+## metatag(set variantSet, tagsCSV string, keyCSV string) (seriesSet|numberSet)
```
captncraig: What if this accepted a variable number of arguments, 1 tag each? And make encoding the new key optional perhaps (if the keys are the same). Thinking `metatag($m, host=alias, iface=desc, somethingElse)`

docs/expressions.md
```diff
@@ -510,7 +510,20 @@ Group functions modify the OpenTSDB groups.
...
+
+metatag adds a tag key/value pair to the set based on metadata in Bosun. tagsCSV are the tag keys from the set that should be used to lookup the metadata value for each result in the set. keyCSV is the metadata key to lookup. The key will be become the new tagkey in the set. If a keyCSV has a second value (as separated by a comma) then the tag key will be renamed to correspond to that value.
+
```
captncraig: We should document rules for exactly how metadata is matched. It can be a bit tricky.

cmd/bosun/expr/funcs.go
```diff
@@ -1182,6 +1199,51 @@ func Rename(e *State, T miniprofiler.Timer, set *Results, s string) (*Results, e
...
+func csv(s string) []string {
+	return strings.Split(s, ",")
+}
+
+func MetaTag(e *State, T miniprofiler.Timer, set *Results, lookupTagsCSV, keyCSV string) (*Results, error) {
+	keys := csv(keyCSV)
+	res := Results{}
+	lookupTags := make(map[string]struct{})
+	for _, k := range csv(lookupTagsCSV) {
+		lookupTags[k] = struct{}{}
```
captncraig: I know `struct{}` is technically more performant or whatever, but bools are so much cleaner looking. If we're gonna do that, maybe we should define a `stringSet` type somewhere with proper `Contains` and `Put` methods. But just use map[string]bool

docs/expressions.md
captncraig: Namely, it will get all metadata entries for which `keyCSV` tags are a subset.

docs/expressions.md
captncraig: I see now, it does not operate quite like I expected. Perhaps more examples with different usage patterns.

cmd/bosun/expr/funcs.go
```diff
@@ -100,6 +100,17 @@ func tagRename(args []parse.Node) (parse.Tags, error) {
 	return tags, nil
 }
 
+func tagMetaTag(args []parse.Node) (parse.Tags, error) {
+	tags, err := tagFirst(args)
+	if err != nil {
+		return nil, err
+	}
+	newTagList := csv(args[2].(*parse.StringNode).Text)
```
captncraig: This is both ugly and unsafe. I really can't understand what it is doing, but I trust that it does. Maybe a simple comment to the effect of "third arg contains the new tag name, possibly comma separated with metadata field"

cmd/bosun/expr/funcs.go
```diff
@@ -1182,6 +1199,51 @@ func Rename(e *State, T miniprofiler.Timer, set *Results, s string) (*Results, e
...
+		lookupTags[k] = struct{}{}
+	}
+	for _, result := range set.Results {
+		ts := make(opentsdb.TagSet)
+		for k := range lookupTags {
+			if value, ok := result.Group[k]; ok {
+				ts[k] = value
+				continue
+			}
+			return nil, fmt.Errorf("tag key %v not found in result", k)
```
captncraig: Is there a way to detect this at parse time?

cmd/bosun/expr/funcs.go
```diff
@@ -1182,6 +1199,51 @@ func Rename(e *State, T miniprofiler.Timer, set *Results, s string) (*Results, e
...
+		}
+		mds, err := e.Metadata.GetMetadata("", ts)
+		if err != nil {
+			return nil, err
+		}
+
+		for _, md := range mds {
+			if md.Name == keys[0] {
```
captncraig: Do we need to detect multiple different values returned (just looking up iface data by host)? zero values returned?

cmd/bosun/expr/funcs.go
```diff
@@ -1182,6 +1199,51 @@ func Rename(e *State, T miniprofiler.Timer, set *Results, s string) (*Results, e
...
+				outkey := keys[len(keys)-1]
+				if _, ok := result.Group[outkey]; ok {
+					return nil, fmt.Errorf("can not add tag key %v because %v is already in the set", outkey, outkey)
+				}
+				value, err := opentsdb.Clean(md.Value.(string))
+				if err != nil {
+					return nil, err
+				}
+				result.Group.Merge(opentsdb.TagSet{outkey: value})
```
captncraig: Why merge vs `result.Group[outkey] = value`? I honestly don't know.

PR #2095, opened 2017-06-02
Add per-region details to hbase collector

clinta:
This change adds collection of per-region metrics from hbase with tags for namespace, region and table.


captncraig:
Unfortunately, this seems to add a lot more datapoints than we anticipated. Decided to enable it by an additional flag in the scollector config called `HbaseRegions`. See #2127

PR #2106, opened 2017-06-16
Add motherboard thermal zones temperatures

simnv:
win.hw.temps.current and win.hw.temps.crittrippoint params


captncraig:
This seems straight forward. A couple questions:

1. Is this general purpose enough to be useful for everybody by default?
2. Is this wmi api availible on all windows machines, or just certain hardaware? Will it get default or invalid data if they don't have the right drivers or motherboard?

PR #2114, opened 2017-06-29
use Multi-Stage docker builds to simplify build process

ChristianKniep:
Leveraging multi-stage builds of 17.06 provides a simpler way of building the image.
It skips the complete bash-script as it just copies the artifacts out of the build-step.

I moved the `Dockerfile` to the root, because I fancy the `COPY . $GOPATH/src/bosun.org/` command over fetching the latest repository via `RUN git clone` or have a little bash script that copies the current working directory next to the `Dockerfile`.
But that might be just my taste, if you guys want me to move it bach to `cmd/bosun/docker/` I can do that and create a reduced `docker.sh`.


captncraig:
I really like this idea. We've talked a lot about refactoring the docker stuff. Multi-stage is a good first step.

I'm ok with moving the dockerfile to the root, because it kinda has to be that way. I'd like to move toward a more granular setup though. I'd prefer the main image to be **just** bosun, and its needed config. None of the tsdb stuff, or other dependencies.

Perhaps an additional image can inherit that one and add tsdb, but I'd almost rather have a directory of docker-compose files for various db scenarios (bosun+redis+tsdb, or bosun+influx) or whatever.

Does that make any sense?

captncraig:
Additionally, I've kinda been waiting on all of this until docker hub supports multi-stage natively. Seems they are dragging their feet (and pushing docker cloud which I am not sure about) Watching [this issue](https://github.com/docker/hub-feedback/issues/1039) for details

blyry:
It looks like multistage has been in for awhile now, bump?

Issue #1819, opened 2016-07-10
Autoforget on silence not working


kylebrandt:
At least in some cases I believe we have observed that alerts that have had the autoforget option set do not forget the alert when it goes unknown.

The logic that used to do this was in https://github.com/bosun-monitor/bosun/commit/18fb20e1c96a223235fdb4671b3953dc6579fc13#diff-e2044b69834f7b8060d01256fc22c495L203

I think maybe the following in runHistory in check.go might need a case added for auto-forget:

```
	// finally close an open alert with silence once it goes back to normal.
	if si := silenced(ak); si != nil && event.Status == models.StNormal {
		go func(ak models.AlertKey) {
			slog.Infof("auto close %s because was silenced", ak)
			err := s.ActionByAlertKey("bosun", "Auto close because was silenced.", models.ActionClose, ak)
			if err != nil {
				slog.Errorln(err)
			}
		}(ak)
	}
	s.Unlock()
	return checkNotify, nil
```

captncraig:
The intention of [this](https://github.com/bosun-monitor/bosun/blob/master/cmd/bosun/sched/check.go#L173) statement is to not create a new incident if the alert key is silenced. Currently active alerts that are then silenced could still go unknown. I believe that was intended behavior at the time we re-worked the incident system. 

Is the intention that the following sequence would forget the alert:?

1. Alert goes crit
2. Silenced
3. goes unknown

It may make sense to forget at 3, but I fear we lose information. The silence should suppress unknown notification.

gbrayut:
I think this happened again, but I'm not 100% sure the silence had autoforget checked because it doesn't show up on the tab anymore (see https://github.com/bosun-monitor/bosun/issues/1897 )

gbrayut:
Happened again. After a 2 week silence with forget we had some alerts go unknown

mlosapio:
I believe I'm seeing this same issue. 


- Alerts are silenced prior to a host being decommissioned with a 6 hour window and "Auto Forget" == True
- Host terminates - no alerts sent. 
- 6 hour window expires and unknown alert triggers an alert.

talonx:
Seeing this on  0.6.0-beta1 too. We're using the API to call silence with the forget option. The silent mode stays for the duration specified and then comes back.

talonx:
Can this be merged? It's blocking our production use of Bosun.

PR #2120, opened 2017-07-11
fix scheduled auto-forget

mlosapio:
cmd/bosun: Resolved issue #1819. Unknown issues need to have an incident.id in order to auto-forget if the host has been permanently decommissioned


talonx:
Can this be merged? It's blocking our production use of Bosun.

PR #2124, opened 2017-07-20
cmd/scollector: allow google to take a base64 encoded json service-account key

captncraig:


cmd/scollector/collectors/google_analytics.go
```diff
@@ -186,16 +186,28 @@ func getPageviews(md *opentsdb.MultiDataPoint, svc *analytics.Service, site conf
 // googleAPIClient() takes in a clientid, secret, a base64'd gob representing
 // the cached oauth token, and a list of oauth scopes.  Generating the token is
 // left as an exercise to the reader.
-func googleAPIClient(clientid string, secret string, tokenstr string, scopes []string) (*http.Client, error) {
+// Or use a base 64 encoded service account json key. Provide json key OR oauth client info.
+func googleAPIClient(clientid string, secret string, tokenstr string, jsonToken string, scopes []string) (*http.Client, error) {
```
kylebrandt: Since this is already kind of a confusing area, should we error if there is a combination of JSON and oauth client info?

PR #2129, opened 2017-07-27
Pop the free ID when assigning to a process

### LINKED ISSUE #2128: BUG: Linux process monitoring ID assignment broken after a process quits
See - https://github.com/bosun-monitor/bosun/blob/master/cmd/scollector/collectors/processes_linux.go#L355

After the first monitored process quits, the ID assigned to it is added to the list `free`. When the next process starts it is assigned the smallest ID from `free` but that ID is not removed, so all future processes receive the same ID.

bharrisau:
Fixes #2128.

### DIFF
--- a/cmd/scollector/collectors/processes_linux.go
+++ b/cmd/scollector/collectors/processes_linux.go
@@ -352,7 +352,11 @@ func (i *idPool) get() int {
 		return i.next
 	}
 	sort.Ints(i.free)
-	return i.free[0]
+
+	var newId int
+	newId, i.free = i.free[0], i.free[1:]
+
+	return newId
 }
 
 func (i *idPool) put(v int) {


captncraig:
Yes, I believe this has been causing us some trouble as well. Thanks for tracking it down. I believe the build is failing because it needs to have gofmt run on the changed file.

bharrisau:
@captncraig I did this quickly just using the github website (seems like I left whitespace on the empty lines). I don't have any Go toolchain setup locally.

Would you be able to take it from here? And check the windows and osx implementations don't have the same issue? Otherwise I may have some time next week to set things up.

bharrisau:
@captncraig Managed to make Travis green. Is this ok to merge?

captncraig:
Yeah, let's do it.

PR #2130, opened 2017-07-27
cmd/scollector/collectors/processes_linux.go - swap tracking

Liudvikas:
Hi,

Added swap tracking capability for the monitored process in linux.
Please let me know what do you think about this. Thanks!

Regards,
Liudvikas

PR #2132, opened 2017-07-27
cmd/scollector/collectors/processes_linux.go - thread and swap tracking

Liudvikas:
Hi,

Added count of thread tracking capability for the monitored process in linux.
Please let me know what do you think about this. Thanks!

Regards,
Liudvikas


captncraig:
How many datapoints do you anticipate this will add? Is it only one datapoint per process per run?

Liudvikas:
Just merged https://github.com/bosun-monitor/bosun/pull/2130 into here.

All this proposed change will add 2 more datapoints per process per run:
```
linux.proc.threads
linux.proc.mem.swap_bytes
```

PR #2139, opened 2017-08-10
Xa iterable

nicollet:
First try to implement some iterable status.

NOT READY FOR MERGE YET.

Just for review.

### DIFF
--- a/cmd/scollector/collectors/fastly.go
+++ b/cmd/scollector/collectors/fastly.go
@@ -1,6 +1,7 @@
 package collectors
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
@@ -67,6 +68,7 @@ const (
 	fastlyScheduledMaintDesc       = "The number of currently scheduled maintenances. Does not include maintenance that is current active"
 	fastlyActiveScheduledMaintDesc = "The number of currently scheduled maintenances currently in progress. Includes the 'in_progress' and 'verifying'"
 	fastlyActiveIncidentDesc       = "The number of currently active incidents. Includes the 'investingating', 'identified', and 'monitoring' states."
+	fastlyMaxDuration              = 3 * time.Minute
 )
 
 var (
@@ -76,7 +78,9 @@ var (
 func c_fastly_status(baseAddr string) (opentsdb.MultiDataPoint, error) {
 	var md opentsdb.MultiDataPoint
 	c := statusio.NewClient(baseAddr)
-	summary, err := c.GetSummary()
+	ctx, cancel := context.WithTimeout(context.Background(), fastlyMaxDuration)
+	defer cancel()
+	summary, err := c.GetSummary(ctx)
 	if err != nil {
 		return md, err
 	}
--- a/cmd/scollector/collectors/iterable.go
+++ b/cmd/scollector/collectors/iterable.go
@@ -47,7 +47,6 @@ const bosunPrefix = "iterable.status."
 func cIterableStat(URL string) (opentsdb.MultiDataPoint, error) {
 	var md opentsdb.MultiDataPoint
 	c := statusio.NewClient(URL)
-	c.SetScheme("http")
 	summary, err := c.GetSummary() // TODO: add timeout
 	if err != nil {
 		return md, err
--- a/cmd/scollector/collectors/iterable.go
+++ b/cmd/scollector/collectors/iterable.go
@@ -2,68 +2,92 @@ package collectors
 
 import (
 	"context"
-	"fmt"
 	"time"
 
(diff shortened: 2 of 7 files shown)


vendor/github.com/bosun-monitor/statusio/statusio.go
```diff
@@ -15,14 +17,26 @@ import (
...
+	if len(baseAddr) != len(url) {
+		return "http", baseAddr
+	}
+	return "https", strings.TrimPrefix(url, "https://")
 }
 
 // NewClient creates a new statusio client for the *public api*.
 // baseAddr is i.e. status.example.copm/api/v2/
 func NewClient(baseAddr string) *Client {
+	scheme, baseAddr := extractScheme(baseAddr)
```
kylebrandt: Can just use `url.Parse` https://play.golang.org/p/mMnQDFb3ot

cmd/scollector/scollector.toml
```diff
@@ -0,0 +1,4 @@
+Host = "http://devbosun.ds.stackexchange.com"
```
kylebrandt: This file shouldn't be part of the PR.

cmd/scollector/collectors/fastly.go
```diff
@@ -67,6 +68,7 @@ const (
 	fastlyScheduledMaintDesc       = "The number of currently scheduled maintenances. Does not include maintenance that is current active"
 	fastlyActiveScheduledMaintDesc = "The number of currently scheduled maintenances currently in progress. Includes the 'in_progress' and 'verifying'"
 	fastlyActiveIncidentDesc       = "The number of currently active incidents. Includes the 'investingating', 'identified', and 'monitoring' states."
+	fastlyMaxDuration              = 3 * time.Minute
```
kylebrandt: In general I'm not clear on this change and the purpose virtue of using context here. I think maybe http already has the timeout set on the client with https://github.com/bosun-monitor/bosun/blob/master/cmd/scollector/main.go#L90

This perhaps something @captncraig probably knows better than me.

cmd/scollector/collectors/iterable.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/bosun-monitor/statusio"
+)
+
+func init() {
+	registerInit(func(c *conf.Conf) {
+		collectors = append(collectors, &IntervalCollector{
+			F: func() (opentsdb.MultiDataPoint, error) {
+				return cIterableStat(c.Iterable.StatusBaseAddr)
+			},
+			name:     "cIterableStat",
```
kylebrandt: For better or worse our convention is underscores and lowercase, for example https://github.com/bosun-monitor/bosun/blob/master/cmd/scollector/collectors/fastly.go#L44 , so should be `c_iterable_status`

vendor/github.com/bosun-monitor/statusio/statusio.go
nicollet: Oh great! I'll change that.

cmd/scollector/scollector.toml
nicollet: Oups will remove.

cmd/scollector/collectors/iterable.go
nicollet: Yes, I hesitated there. I suggest one day we rename all of them as it makes golint output unreadable.
Collectors should be bounded to this project anyway, so that would be doable.

cmd/scollector/collectors/iterable.go
```diff
@@ -2,68 +2,92 @@ package collectors
...
 	"bosun.org/cmd/scollector/conf"
 	"bosun.org/metadata"
 	"bosun.org/opentsdb"
+	"bosun.org/slog"
 	"github.com/bosun-monitor/statusio"
 )
 
+// iterable.io collector
+// to use add this in your scollector.toml:
+// [Iterable]
```
kylebrandt: We would be better open source citizens if the component keys were selectable from the config. 

If this would mean writing a bunch of metadata that isn't useful for us right now that doesn't mean we have to expand the scope that much. We could only support the ones we care about to merge this, and then if people want to add ones down the road

captncraig: Would that make it a more generic StatusPageIO collector? Seems reasonable. Config would have a `map[string]string` perhaps, mapping component name to tag name. Or can we even just automatically add cleaned tags for each component we find?

kylebrandt: Well there is the metadata descriptions etc. Also when I wrote that statusio library I did it by reverse engineering the json api, I couldn't actually find documentation.

We can just keep it as is, I'm worried I just started scope creep.

PR #2143, opened 2017-08-13
WIP: Prometheus backend

dimamedvedev:
To test
```
docker run --rm --name prometheus -p 9090:9090 prom/prometheus
```
and run `prom("up", "5m", "", "30s")` in expression tab.

Current problems:

1.  Prometheus shows all tags when doing metric query, it'll probably break Bosun alert keys There's a way to aggregate on particular tags like `avg(metric_name) by (tag_name)` ( see https://prometheus.io/docs/querying/operators/#aggregation-operators ), but that requires more sophisticated `expr.PromTag()` function and I haven't yet figured out how to deal with `promql.ParseExpr()` output.

Todo

- [ ] vendor
- [ ] tests
- [ ] TagQuery function
- [ ] filter tags in response


kylebrandt:
done via #2366

Issue #680, opened 2015-01-30
Why use UTC, rather than a Local？


wkshare:
We in China, the bosun's time zone display is not correct。

example：
moment.utc() and time.Now().UTC()

Can remove the UTC? or Add timezone to dev.conf ?

kylebrandt:
Non UTC support isn't of any use to use at Stack - so we won't be working on this any time soon. However if there is a well tested pull request for this feature I think we would be happy to accept it.

wkshare:
thks，i know。

We try to modify。

change time.Parse() to time.ParseInLocation in *.go（funcs.go, expr.go, web.go, tsdb.go ... ...)
change moment.utc to moment.local in bosun.js
and TZ=Asia/shanghai ./bosun

After the success, wish to join configuration in dev.conf。

What you suggest?

We have 3000+ machines。2W+ qps moniting data.
We'd like to try to use the bosun：）

kylebrandt:
We would want the all js to use whatever timezone is in the config for now
(and maybe eventually users can configure it). That way the interface can
be in UTC still (or the local time).

Also be aware that we use typescript to (all the tsc files) to generate .js
and don't edit the js directly.

Dieterbe:
just linking to whshare's PR from here: #685 
big +1 from me, I would like this feature a lot :)

kylebrandt:
One this I will want to make sure is that this settable my people (or at least a whole instance). I don't want to view things in localtime, rather see them in UTC. So I don't want the web interface restricted to localtime of the client machine.

wkshare:
@kylebrandt   The web page, use the best Localtime，I 'm sure :)

gbrayut:
We have servers and employees in many different time zones and want everything displayed in UTC. If you have a pull request to allow local time on bosun that is fine, but we still want to be able to configure UTC for all our users.

nickman:
UTC is fine, considering the requirement, but it would be nice if it was labeled as UTC. Otherwise (in other time-zones) it just looks wrong.
![tz](https://cloud.githubusercontent.com/assets/597608/6416037/0e4cf6f2-be71-11e4-9080-72c47c7ae874.png)

wkshare:
@nickman i got it

Dieterbe:
closed, why? i think some good points were brought up here :?

kylebrandt:
Want to get to a more manageable length of issues. Supporting local time is
a pretty big project with no contenders to do the work any time soon.

PR #2145, opened 2017-08-20
bosun: print "UTC" next to the time when displaying graphs

nicollet:
partially solve issue 680
https://github.com/bosun-monitor/bosun/issues/680

The only relevant file changed is: cmd/bosun/web/static/js/directives.ts


cmd/bosun/main.go
```diff
@@ -282,6 +282,10 @@ func main() {
 	}()
 
 	if *flagWatch {
+		// note for vim users, I've this in my .vimrc:
```
kylebrandt: Don't think this should be a comment in main.go. Maybe add a tips section to https://github.com/bosun-monitor/bosun/blob/master/CONTRIBUTING.md

cmd/bosun/web/static/js/bosun.js
```diff
@@ -867,9 +867,9 @@ var UsernameInputController = (function () {
     function UsernameInputController(auth) {
         this.auth = auth;
     }
+    UsernameInputController.$inject = ['authService'];
```
kylebrandt: I'm not sure why this is moving. Which version of tsc are you using?

cmd/bosun/main.go
nicollet: Sure, will update.

cmd/bosun/web/static/js/bosun.js
nicollet: $ tsc --version
Version 2.4.2

kylebrandt: should be npm i -g typescript@2.3.1

kylebrandt: Then will need to run `go generate` again

nicollet:
I think those changes should look good now. I don't know if I (or you) should rebase/push -f before applying. The last commit is not very worth it as it is just a go generate with the right version.

PR #2146, opened 2017-08-22
Xa lint

nicollet:
## Lint the collectors, the easy parts

- first commit: rename functions so that they are camelCase,
- second commit: easy small fixes.

There are tons of other fixes which are needed, mainly comments, but not only.


kylebrandt:
Renaming these is problematic because because people use the `-f` cli to filter certain collectors. So I don't think fixing lint or dual naming things is really worth it.

nicollet:
I only have renamed the functions name in go, not their 'name:' field for those who have one.

```
| -                                     return c_fastly(client)
| +                                     return cFastly(client)
|                               },
|                               name:     "c_fastly",
```

So our users will use the tool the same way, but contributors who use golint will have a less hard time.

Issue #2148, opened 2017-08-23
Influx derivative query conflicts with bosun rules about grouping and where clauses in the query string.


BRNTZN:
Using influx derivative:

`influx("db", '''SELECT  DERIVATIVE(mean("someValue")) FROM "table"''', "30d", "", "3d")
`
Results in:
`fatal: main.go:121: couldn't read rules: conf: path/to/conf/rules.conf:xx:y: at <crit = $avgincreasel...>: expr: derivative aggregate requires a GROUP BY interval`

Tried adding the necessary group by clause to the query:
`influx("db", '''SELECT  DERIVATIVE(mean("someValue")) FROM "table" group by time(3d)''', "30d", "", "3d")
`
Results in:
`fatal: main.go:121: couldn't read rules: conf: path/to/conf/rules.conf:xx:y: at <crit = $avgincreasel...>: expr: aggregate functions with GROUP BY time require a WHERE time clause`

Then tried adding an arbitrary where clause:
`influx("db", '''SELECT  DERIVATIVE(mean("someValue")) FROM "table" where time < now() group by time(3d)''', "30d", "", "3d")
`
This time bosun actually starts up however when checking the alert:
```
2017/08/23 07:51:21 info: check.go:588: check alert myalert start with now set to 2017-08-23 07:51:21.392396985
2017/08/23 07:51:21 error: check.go:712: influx query must not contain time in WHERE
2017/08/23 07:51:21 error: check.go:629: Error checking alert myalert: influx query must not contain time in WHERE
2017/08/23 07:51:21 info: check.go:636: check alert myalert done (584.145µs): 0 crits, 0 warns, 0 unevaluated, 0 unknown
2017/08/23 07:51:21 info: alertRunner.go:89: runHistory on total.message.count.increase took 48.016µs
```

arwineap:
It looks like this is actually coming from the influxdata/influxql module in vendor

https://github.com/bosun-monitor/bosun/blob/master/vendor/github.com/influxdata/influxdb/influxql/ast.go#L1793

earwin:
The cause of this error is that InfluxQL requires `GROUP BY time()` for several functions, and while these are normally added by bosun, it only happens _after_ the initial parsing is done.

earwin:
The only good approach to a fix I see, taking into account more fanciness that comes into InfluxDB, like subqueries, is to:
* remove `from`, `to`, `groupbytimeinterval` parameters from `influx()` function
[...]

PR #2149, opened 2017-08-23
Issue2148fix

BRNTZN:
How I fixed issue #2148 

Did not change any tests and this is probably not the best way to go about it but I'm a noob at go.


BRNTZN:
Fixed the build

PR #2153, opened 2017-08-31
Extend cadvisor collector with additional metrics

skygeo:
Increase the capabilities of the cadvisor collector.
This includes:
* Extra cpu metrics: quota, shares and period
* Extra memory metrics: cache, rss, swap, failcnt, limit, reservation, swaplimit
* New option TagsFromLabels. Enabling adding custom tags based on labels.
* New option IgnoreTags. Enable removing tags.
* Update cadvisor vendor client

### DIFF
--- a/cmd/scollector/collectors/cadvisor.go
+++ b/cmd/scollector/collectors/cadvisor.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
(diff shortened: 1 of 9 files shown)


skygeo:
Hi guys. Could you have a look at the PR? It would be nice to have it reviewed one day :-)

PR #2154, opened 2017-08-31
ElasticIndexDetails IP field is not an array

pho:
The elasticsearch collector was failing due to:
`scollector[13949]: error: interval.go:65: elasticsearch: json: cannot unmarshal string into Go struct field .ip of type []string`

On my elasticsearch 5.x node:

`elastic-1:~# curl localhost:9200/_nodes/_local/stats`
`... ,"ip":"172.16.0.19:9300",....`

So this PR fixes it, although I have a single node cluster, so I dont really know if this response would be different with more nodes


mhenderson-so:
This commit would break the ElasticSearch collector for versions below 5.x, where that field _is_ an array. I (by coincidence) submitted PR #2155 which fixes this issue in a way that does not break the collector for older Elastic versions (cc @captncraig)

pho:
Such coincidence haha

Thank you! I'll close this PR then

PR #2157, opened 2017-08-31
cmd/scollector: Hide TSDB credentials in logs

dschneller:
Upon startup, scollector reports which OpenTSDB host it is connected to.
In case there is authentication required, the credentials were logged as
part of the URL. This patch replaces them with dummies.

### DIFF
--- a/cmd/scollector/main.go
+++ b/cmd/scollector/main.go
@@ -256,7 +256,7 @@ func main() {
 	}
 	cdp, cquit := collectors.Run(c)
 	if u != nil {
-		slog.Infoln("OpenTSDB host:", u)
+		slog.Infoln("OpenTSDB host:", hideUrlCredentials(u))
 	}
 	collect.UseNtlm = conf.UseNtlm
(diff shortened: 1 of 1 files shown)


cmd/scollector/main.go
```diff
@@ -403,6 +403,17 @@ func parseHost(host string) (*url.URL, error) {
 	return u, nil
 }
 
+func hideUrlCredentials(u *url.URL) *url.URL {
+	// Copy original url, replace credentials, e. g. for logging
+	if u.User != nil {
+		u2 := new(url.URL)
```
captncraig: This should only do anything if the user is already set. Don't want to add xxx if there is no user to begin with.

dschneller: I am not sure I understand. That's why there is `if u.User != nil` check -- only when a User was set in the original URL, a copy will be made, with the xxx instead of the original credentials. So unless I am overlooking something, this already does what you ask for?

captncraig: wow, I spaced that one. Yeah, you're right.

PR #2163, opened 2017-09-26
iterable: use status.io api

nicollet:
* add context.Context: timeout for fastly and iterable.

* https support,
* tunings through configuration,
* variables and consts are not global,
* less reliance on states outside of functions.

* has tests

* only monitors according to the plan

https://docs.google.com/document/d/1ZO4fKLGU0WCC9eldOLIvLU80iNEFupcUS9kswclm2Q8/edit?ts=59b9b782#
(section Iterable System Health)

### DIFF
--- a/cmd/scollector/collectors/fastly.go
+++ b/cmd/scollector/collectors/fastly.go
@@ -1,6 +1,7 @@
 package collectors
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
@@ -67,6 +68,7 @@ const (
 	fastlyScheduledMaintDesc       = "The number of currently scheduled maintenances. Does not include maintenance that is current active"
 	fastlyActiveScheduledMaintDesc = "The number of currently scheduled maintenances currently in progress. Includes the 'in_progress' and 'verifying'"
 	fastlyActiveIncidentDesc       = "The number of currently active incidents. Includes the 'investingating', 'identified', and 'monitoring' states."
+	fastlyMaxDuration              = 3 * time.Minute
 )
 
 var (
@@ -76,7 +78,9 @@ var (
 func c_fastly_status(baseAddr string) (opentsdb.MultiDataPoint, error) {
 	var md opentsdb.MultiDataPoint
 	c := statusio.NewClient(baseAddr)
-	summary, err := c.GetSummary()
+	ctx, cancel := context.WithTimeout(context.Background(), fastlyMaxDuration)
+	defer cancel()
+	summary, err := c.GetSummary(ctx)
 	if err != nil {
 		return md, err
 	}
--- a/cmd/scollector/collectors/fastly.go
+++ b/cmd/scollector/collectors/fastly.go
@@ -68,7 +68,7 @@ const (
 	fastlyScheduledMaintDesc       = "The number of currently scheduled maintenances. Does not include maintenance that is current active"
 	fastlyActiveScheduledMaintDesc = "The number of currently scheduled maintenances currently in progress. Includes the 'in_progress' and 'verifying'"
 	fastlyActiveIncidentDesc       = "The number of currently active incidents. Includes the 'investingating', 'identified', and 'monitoring' states."
-	fastlyMaxDuration              = 3 * time.Minute
+	fastlyMaxDuration              = 1 * time.Minute
 )
 
 var (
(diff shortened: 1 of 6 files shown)


cmd/scollector/collectors/fastly.go
```diff
@@ -67,6 +68,7 @@ const (
 	fastlyScheduledMaintDesc       = "The number of currently scheduled maintenances. Does not include maintenance that is current active"
 	fastlyActiveScheduledMaintDesc = "The number of currently scheduled maintenances currently in progress. Includes the 'in_progress' and 'verifying'"
 	fastlyActiveIncidentDesc       = "The number of currently active incidents. Includes the 'investingating', 'identified', and 'monitoring' states."
+	fastlyMaxDuration              = 3 * time.Minute
```
kylebrandt: Do we know that we need this timeout?

In main.go we set the defaultclient timeouts to for http and collect to one minute.

cmd/scollector/collectors/iterable.go
```diff
@@ -0,0 +1,95 @@
...
+//   # TsdbPrefix = "iterable.status."
+//   # MaxDuration = 3 # seconds
+
+type IterComp map[string]string
+
+func init() {
+	const (
+		defaultStatusBaseAddr = "https://iterable.statuspage.io"
+		defaultTsdbPrefix     = "iterable.status."
+		defaultMaxDuration    = 3 // Seconds
```
kylebrandt: If we do continue with the timeout, 3 second sseems like a short default

cmd/scollector/collectors/iterable.go
```diff
@@ -0,0 +1,95 @@
...
+	"bosun.org/slog"
+	"github.com/bosun-monitor/statusio"
+)
+
+// iterable.io collector
+// to use add this in your scollector.toml:
+// [Iterable]
+//   StatusBaseAddr = "https://iterable.statuspage.io"
+//   # TsdbPrefix = "iterable.status."
+//   # MaxDuration = 3 # seconds
```
kylebrandt: if we keep this feature, I think is should be a duration. Example of how this is done in Bosun with this type https://github.com/bosun-monitor/bosun/blob/master/cmd/bosun/conf/system.go#L560

cmd/scollector/collectors/iterable.go
```diff
@@ -0,0 +1,95 @@
...
+func iterable(ctx context.Context, iter conf.Iterable, compKey IterComp) (opentsdb.MultiDataPoint, error) {
+	var md opentsdb.MultiDataPoint
+	c := statusio.NewClient(iter.StatusBaseAddr)
+	summary, err := c.GetSummary(ctx)
+	if err != nil {
+		return md, err
+	}
+	for _, comp := range summary.Components {
+		if key, ok := compKey[comp.Name]; ok {
+			// TODO: should Add() support a timeout?
```
kylebrandt: remaining TODO

kylebrandt:
@captncraig strange we are vendoring something in the repo... Should we even be here? But I guess if so, the commit should include the changes the lib as well.

cmd/scollector/collectors/fastly.go
nicollet: Will change it to one minute. Note that if c.GetSummary() was called several times or if it was doing several http requests, the whole function would have a global 1 minute boundary. 1 minute makes sense here since this check is launched every minute.

cmd/scollector/collectors/iterable.go
nicollet: I can do that if needed. We may not want to put too much efforts into it though, if we have sensible defaults.

kylebrandt: If we are keeping the the option I would like it to be a duration for consistency across Bosun and collector config (where possible). 

If we just take a default and don't make an option in the toml file, then yes, don't really care that much about the type.

cmd/scollector/collectors/iterable.go
nicollet: What would you suggest? Should we trust data that is coming from a webpage that has taken more than several seconds to render? Probably not.

cmd/scollector/collectors/iterable.go
nicollet: Will remove. Add() is used everywhere and is quite complicated. We have to assume that it will not block, at least for now.

nicollet:
For the vendoring thing: I have the same set of patches for the library too.

PR #2167, opened 2017-09-29
Purge old Rendered Templates

kylebrandt:


cmd/bosun/database/test/util/purge_rendered.go
```diff
@@ -0,0 +1,57 @@
...
+			if st.Open {
+				continue
+			}
+			if len(st.Actions) == 0 || st.Actions[len(st.Actions)-1].Time.Before(earliestOk) {
+				ids = append(ids, st.Id)
+			} else {
+				continue
+			}
+			fmt.Println(st.Id, len(st.Events), st.Open)
+			s, err := redis.String(myConn.Do("GET", fmt.Sprintf("renderedTemplatesById:%d", st.Id)))
```
kylebrandt: renderedTemplatesKey(st.Id)

cmd/bosun/database/state_data.go
```diff
@@ -447,3 +516,58 @@ func (d *dataAccess) transact(conn redis.Conn, f func() error) error {
...
+					continue
+				}
+				id, err := strconv.ParseInt(parts[1], 10, 64)
+				if err != nil {
+					slog.Error(err)
+					continue
+				}
+				state, err := d.getIncident(id, conn)
+				if err != nil {
+					if strings.Contains(err.Error(), "nil returned") {
```
kylebrandt: strings.Contains on error, combined with the match not being a constant doesn't seem safe in the long run.

cmd/bosun/database/state_data.go
```diff
@@ -93,16 +96,80 @@ func (d *dataAccess) GetRenderedTemplates(incidentId int64) (*models.RenderedTem
...
 	return renderedT, nil
 }
 
+func (d *dataAccess) GetRenderedTemplateKeys() ([]string, error) {
+	conn := d.Get()
+	defer conn.Close()
+
+	//ledis uses XSCAN cursor "KV" MATCH foo
+	//redis uses SCAN cursor MATCH foo
+	cmd := "SCAN"
```
kylebrandt: Maybe Redis and Ledis SCAN could be a function outside of this that can be reused (and make this func shorter)

captncraig: Yeah, we did that with some others, but the re-ordering of the args makes it tricky. I will factor out.

cmd/bosun/database/state_data.go
captncraig: Yeah, the problem there is that the error will already be wrapped at that point, and the inner error is not accessible. I guess I can also make a package function for that test though.

kylebrandt:
@captncraig Didn't find any problems testing this out, but needs conflicts resolved.

Issue #1982, opened 2016-12-12
Support Elasticsearch 5.x as a Datasource


kuchinas:
I can't get any data from elasticsearch 5.0.2. Get the error:
No ElasticSearch Node Available
On elasticsearch 2.4.2 works.

FAKERINHEART:
May be the api of el 5.0.2 is different from that of the el 2.*.*.

kylebrandt:
Yes, had to do this when we went from 1 to 2. In that case the logstash specific functions (funcs that started with ls) were deprecated, so we didn't try to support ES v1 with the new functions.

In this case, we may need to support both v2 and v5. The library we use is https://github.com/olivere/elastic , so if someone wants to try adding support for v5 via the library you are welcome ,but don't want to drop support for v2 I don't think.

Right now Stack Overflow is still at v1 (Bosun doesn't doesn't talk to our v1 ES servers) and v2, so not going to worry about this until we are looking at v5 in terms of our own development time - so for now would need to be a PR.

I imagine that the scollector collector will also need to be updated, because they mess with the monitoring APIs a lot (at least previously).

ayashjorden:
Hi,
I'm looking into this.
Current breaking changes:
1. [SearchResult.Shards was removed](https://github.com/olivere/elastic/issues/482). @kylebrandt can you elaborate on the Shards check in cmd/bosun/expr/elastic.go#L281 ?
2. SearchService.Do now requires a [context.Context](https://godoc.org/golang.org/x/net/context) object, will use contect.Background() for now.
3. `md/bosun/expr/elastic.go:437: cannot use v.Key / 1000 (type float64) as type int64 in argument to time.Unix` - thinking about using int64(v.Key/1000), but still need to better my understanding on time calculations/math-rounding in go.

Regarding maintaining support for ES-2 and ES-5, it may be required to wrap olivere/elastic usage with another interface and add config param on which one to use (ES-2 or ES-5 interface).

Appreciate comments on the above.

Jorden

kylebrandt:
1. I Responded to olivere. Basically before we were always fetching the indices before each query which was slow. This allows us to generate the indices and "hope" they exist, and if some are missing within the time range there will not be an error. Otherwise we could only ask for indices that exist, which would require querying elastic to find out. Caching is Hard, this seems to work fine.

2. Seems fine. 
3. We don't use milliseconds so that was just to convert the epoch to second. 

[...]

PR #2180, opened 2017-10-13
bosun: wip esv5 support via build tags

kylebrandt:


kylebrandt:
todo change tag from eslatest to es5

kylebrandt:
will fix #1982

PR #2183, opened 2017-10-19
cmd/bosun: add option to require message field in save

kylebrandt:


kylebrandt:
screen shot 2017-10-19 at 3 58 04 pm

PR #2196, opened 2017-11-28
Fix post notification logging

pradeepbbl:
This will add alert key and notification name to  HTTP notification failure log.

Current failure log:
```
2017/11/21 14:15:41 error: notify.go:48: sending http: bad response on notification POST: 400
```
With this change:
```
2017/11/28 15:17:18 error: notify.go:48: sending http: bad response on notification with name bosun_uptime for alert bosun_uptime{} method POST: 400
```


pradeepbbl:
looping @kylebrandt

PR #2200, opened 2017-12-12
Add editor to expr

spuzirev:
Hello.

We've lots of UX troubles trying to utilize Expr editor which is basically just textarea, therefore you don't have syntax highlight or even monospace font.
That's why I've implemented AceEditor to expr editor.


spuzirev:
Closed due to wrong parent commit

PR #2201, opened 2017-12-12
Replace textarea with aceeditor on expr

spuzirev:
Hello.

We've lots of UX troubles trying to utilize Expr editor which is basically just textarea, therefore you don't have syntax highlight or even monospace font.
That's why I've implemented AceEditor to expr editor.


mvuets:
I believe this PR can be rejected/closed in favour of already merged #2266

PR #2210, opened 2018-01-08
Native short links in bosun

captncraig:


captncraig:
Google link shortener not supported any longer. We could remove the config key if we wanted to make that a breaking change. 

Links now generate as {bosun-host}/s/123

cmd/bosun/conf/system.go
```diff
@@ -539,10 +539,12 @@ func (sc *SystemConf) AnnotateEnabled() bool {
 // MakeLink creates a HTML Link based on Bosun's configured Hostname
 func (sc *SystemConf) MakeLink(path string, v *url.Values) string {
 	u := url.URL{
-		Scheme:   "http",
-		Host:     sc.Hostname,
-		Path:     path,
-		RawQuery: v.Encode(),
+		Scheme: "http",
```
kylebrandt: hardcoded http probably not good. I think someone has a PR regarding this.

kylebrandt:
@captncraig Can you expand on not supported anymore? In other words, what happens to people that using Google shortner and start using this branch?

captncraig:
Not supported anymore as in, we don't bother making a google shortlink. Just make our own. The code to call google is removed entirely.

PR #2238, opened 2018-03-30
ICMP Improvements

clinta:
This PR is a rewrite of the icmp collector. It replaces go-fastping with [go-multiping](https://github.com/TrilliumIT/go-multiping) a library I wrote with scollector in mind.

With the current ICMP implementation, go-fastping opens a raw socket for each host being pinged, and then must process every ICMP packet the host receives to find the one it is looking for. We have some scollector hosts which ping over 1000 hosts, so every 15 seconds, 1000 raw socket listeners are started, and 1000 go-fastping threads must each process all 1000 responses to find the one they want.

This new implementation opens only one socket, and calls the appropriate handler for a given response.

This implementation also delays the first ping for a random duration up to the configured interval to even out the spikes of icmp traffic.

This PR also adds the ability to override the timeout and interval for individual hosts and on non-windows systems adds a metric for graphing the TTL on ICMP responses which can give insight into routing changes.


captncraig:
Wow, that seems pretty cool. Will need to look it over pretty closely, but it does seem to be a good improvement. I was never very pleased with how that all worked.

PR #2135, opened 2017-07-27
WIP: Arbitrary Template Keys

captncraig:
Makes notifications more individually configurable.

## BREAKING CHANGES:

Notifications no longer support `body` inline template, or `useBody` flag. Instead, you can supply `bodyTemplate` to select any pre-rendered template key on the alert's template:

 Old:
 ```
notification n {
   post = https://whatever.com/alert
   body = {"text": "{{.}}"}
}
template t {
  subject = something broke
  body = real bad
}
alert a {
  template = t
  critNotification = n
  ...
}
```

New:
 ```
notification n {
   post = https://whatever.com/alert
   bodyTemplate = postData
}
template t {
  subject = something broke
  body = real bad
  postData = {"name": "{{.Subject}}"}
}
alert a {
  template = t
  critNotification = n
  ...
}
```

Templates can now include arbitrarily named keys (not just subject and body anymore), and they will all be pre-rendered when the alert fires, with the full context of the alert.

## Features

1. Can add arbitrary keys to templates, as explained above.
2. Templates can "inherit" another template, essentially copying all of their keys/values that are not already defined:
         
        template postToChat {
            postData = {"text": "{{.Subject}}", "context": "{{.Alert.Name}}"}
        }
        template t {
          inherit = postToChat
          subject = Something bad happened
        }
3. Notifications now support `getTemplate` and `postTemplate` to select a template to use for building urls. They have `bodyTemplate` to select a template to use for post or email body (default is `body`) and `emailSubjectTemplate` to select a template to use for the email subject.

PR #2241, opened 2018-04-17
docs: Document json, makeSlice, and makeMap template functions

mvuets:
Presumably as part of the PR #2135 three new global template functions
were added: json, makeSlice, and makeMap. However they are not mentioned
nor documented at http://bosun.org/definitions. This commit attempts to
fill in the gap.


captncraig:
Thank you!

Issue #2197, opened 2017-12-01
On JSON notification issues


mvuets:
Hi. I've been asked to look into an issue of making JSON notifications (for instance to POST JSON to PD on an incident ack, etc.). My understanding is that composing JSON using a text template engine is inherently incorrect: it is clumsy and error-prone.

From this pull request #2135 and [@kylebrandt's comment](https://github.com/bosun-monitor/bosun/pull/2135#issuecomment-327512356) I see that there is an effort to improve the situation. There are three helpers `makeSlice`, `makeMap` and `| json` that could be used to assemble and serialise a data structure and avoid dealing with JSON encoding issues. Yet they don't seem to quite cut it.

One issue in particular is related to a complex data structure provided to, say, `actionBodyAck` and limited capabilites of Go's _text/template_ package: how to traverse multi-dimensional data? `.States` is an array, so I presume multiple accidents may be acknowledged at once. Here is where I fail to come up with an appropriate way to get IDs across all `.States` items. 

I can think of two potential enhancements.

1. Keep composing JSON by hand as text, but provide with a JSON escaping helper that would be a middle ground between built-in `js` and Bosun's `json`. It would encode given values (both scalar and structures) in such a way that it can be always safely used in place of an object field value:
```
jsonBody = `{
  "key": {{ V "$key" | json_val }},
  "accident_nr": {{ len .States | json_val }},
  "accident_ids": [ {{ range .States }}{{ .Id }},{{ end }} ],
  "summary": {{ .Subject | json_val }}
}
`
```

2. Provide templates with another helper that could help querying data, return it in a structured way which then can be serialised with `json`. Something alike [jq](https://stedolan.github.io/jq/). For example:
```
jsonBody = `
  {{- $accident_nr := len .States -}}
  {{- makeMap "key" (V "$key") "accident_nr" $accident_nr "summary" .Subject "accident_ids" (jq "[.States[].Id]") | json -}}
`
```

I would be glad to hear what's your vision on improving support for JSON notifications? Thanks.

kylebrandt:
I played with this a little bit last week:

- Idea: Add an `append` like command to the template language. Problem - the template language doesn't allow access to to modifying variables in the outside scope within the inner scope
 - Other idea: Make a new type slice that has methods attached to it. And append and and a method that returns something you can range over (iterator). Might work but awkward

We might just need to make sure there are some different arrangements of the state data attached to the context for the general use cases.

mvuets:
[...]

PR #2243, opened 2018-04-18
cmd/bosun, docs: Add execTmpl template function

mvuets:
In this commit I try to partly address an issue raised in the report #2197:
I want to generate a JSON on `actionBodyAck` and send it to an HTTP API.
One of the fields in the JSON payload should contain a message attached
to the performed action and also a list of ids of affected incidents,
e.g.: "User X acknowledged N incidents (ids: 1, 2, 4) with a note: M".
I need to use `makeMap` and `json` functions, which take a "pipeline"
in Go's terms, to build a properly encoded JSON; but I also need a
way to traverse `.States[]` and extract a list of ids in a form of a
"pipeline". I could traverse with a help of the `{{range}}` template
action, but there is no way I can capture the result of that execution.

This patch adds a template function `execTmpl` which executes given Go
text/template provided a data structure and returns a resulting string
which can be kept in a template variable or passed on through the
template pipeline. Another way of looking at it is in-line or
sub-templates.

    {{- $list := execTmpl "{{range .}}<li>{{.Id}}</li>\n{{end}}" .States -}}
    {{- $html := printf "%v<br/>Affected incidents:<ul>%v</ul>" .Message $list -}}
    {{- makeMap "message" $html | json -}}

At first I wanted to take advantage of Go's named sub-templates
(`{{define NAME}}...`) and pass that NAME to `execTmpl` instead of a
literal template. Unfortunately template functions do not get a context
of a template in which they are being invoked, therefore I did not have
an access to the parent template containing sub-templates. I could wrap
it in a closure, but there are too many text/template instances created
in different places in code, so I figured that I will give the current
approach a go first and see how is it accepted.

### DIFF
--- a/cmd/bosun/conf/rule/loaders.go
+++ b/cmd/bosun/conf/rule/loaders.go
@@ -1,6 +1,7 @@
 package rule
 
 import (
+	"bytes"
 	"fmt"
 	"net/mail"
 	"net/url"
@@ -33,6 +34,17 @@ func (c *Conf) loadTemplate(s *parse.SectionNode) {
 	funcs := template.FuncMap{
 		"V": func(v string) string {
 			return c.Expand(v, t.Vars, false)
+		},
+		"execTmpl": func(text string, data interface{}) string {
+			tmpl, err := template.New("inline-tmpl").Funcs(defaultFuncs).Parse(text)
+			if err != nil {
+				return "execTmpl fails to parse template: " + err.Error()
+			}
+			buf := &bytes.Buffer{}
+			if err := tmpl.Execute(buf, data); err != nil {
+				return "execTmpl fails to execute template: " + err.Error()
+			}
+			return buf.String()
 		},
 	}
 	saw := make(map[string]bool)
--- a/docs/definitions.md
+++ b/docs/definitions.md
@@ -1280,6 +1280,37 @@ alert bytes {
 }
 ```
 
+##### execTmpl(string, data) (string)
+{: .func}
+
+`execTmpl` takes a template string, applies it to the given data, and returns a resulting string. The template format is the same as the one used to describe [alert templates](definitions#templates) themselves. This function is useful when you want to capture the result of executing a template into a variable instead of outputting it directly.
+
+Example:
+
+```
+alert execTmpl {
+    template = execTmpl
+    critNotification = execTmpl
+    crit = 1
+}
+
+notification execTmpl {
+    post = https://api.example.com/jabber?send_to=somebody@example.com
+    runOnActions  = Ack
+    actionBodyAck = execTmplAck
+}
+
+template execTmpl {
+    subject = `execTmpl example`
+    body = `execTmpl example body`
+    execTmplAck = `
+        {{- $incidentIds := execTmpl "{{range .}}{{.Id}} {{end}}" .States -}}
+        {{- $text := printf "%v %v %v incident(s): %v" .User .ActionType $incidentIds .Message -}}
+        {{- makeMap "message" $text | json -}}
+    `
+}
+```
+
 ##### html(string) (htemplate.HTML)
 {: .func}
 


captncraig:
Wow, this seems... cool?

My initial reaction is "really, there isn't a better way to do that"? But I guess if you want to use json instead of manually building out json templates then yeah, maybe that makes sense. 

LGTM code-wise. Leaving to @kylebrandt to decide if it is a good idea to merge or not.

kylebrandt:
This is cool. I feel like the translation to text might make it more subject to runtime errors, and possibly make template debugging a little more difficult.  That being said it does solve the issue.

It might be more object based if we were able to attach a "scratch" object to each template that users can write to. Like what Hugo does: https://gohugo.io/functions/scratch/ 

Thoughts on that approach?

mvuets:
@captncraig, I would love to manually build out a JSON template. It would look more natural and easy to comprehend for those templating their alerts (in our case it is not a single person or a team, but hundreds of people who need to monitor their products). However there is a stumbling block that I could not sort out without patching Bosun.

Basically: how to make this template produce a valid JSON document given `.Message` is `say "hi '\xyz'!\n" & jump <sup>high</sup>`?
```
alert x {
    template = x
    critNotification = x
    crit = 1
}
notification x {
    post = https://api.example.com/jabber
    runOnActions  = Ack
    actionBodyAck = jabberActionAckBody
}
template x {
  subject=`subject`
  body = `body`
  jabberActionAckBody = `{
    "text": "{{.User}} {{.ActionType}} incidents [{{range .States}}{{.Id}}; {{end}}]: {{.Message}}"
  }`
}
```
There is an escaping problem. Neither `json` nor `js` fully cover it. Using `makeMap ... | json` helps, but then I get another problem—I cannot iterate over `.States[].Id`. See what I mean?

@kylebrandt, this patch solves my problem, but I am afraid and to be frank that it still does not solve the general case. Consider you want to generate the following JSON document out of data mentioned in the example above:
```
{ "text": "jsmith acked with a note: I took care of this",
  "incidents": [12, 24, 48] }
```

The scratch object idea is interesting. However I do not like that it introduces a mutable state into templates and takes it from dumb templating into rather some sort of scripting space.

That said neither I am a big proponent of my patch, but it seems to solve an interim problem without introducing too much complexity or new concepts, arguably. And it opened this conversation.

Overall I think that templating JSON is hard and that end-users should be handed some other sort of a templating mechanism (I mentioned some idea in the referenced GH issue). Go's _text/template_ is good enough when a user is in charge of both code and templates, and here the user's got the latter only.

kylebrandt:
@mvuets What else do you want grouped into an a single object from the States besides IDs? So far the ID is what people have been looking for, we can just pop a method or another data structure with links onto ActionNotificationContext to make these objects available.

mvuets:
@kylebrandt, accident IDs, their corresponding web page URLs, CurrentStatus, WorstStatus. What do you envision the data structure look like? The nature of this data is array (as opposite to scalar/singular in a normal notification context), so in my mind either way one inevitably has to somehow iterate over it. You could squash everything into a string, say comma-separated, but then you lose control over presentation of this data.

...just occurred to me: what if we kind of flatten-transpose {{.States}} and proxy (_strings.Join_)[https://golang.org/pkg/strings/#Join] into templates? I.e.:
```
{
  "States": {
    "Id": [24, 48],
    "AlertKey": ["disk_space", "avg_load"],
    ...
  },
  "User": "...",
  ...
}

template = `
  {{- $ids := join .States.Id ", " -}}
  {{- makeMap "text" (printf "%v %v incidents %v: %v" .User .ActionType $ids .Message) | json -}}
`
```
I think I like this idea the best so far. Hmm? (-:

kylebrandt:
@mvuets I'm more thinking squashing into objects, since what we are really doing is reshaping data here. If you could give me a few desired end objects I might be able to prose some methods and structures to create them.

So what would help is a few examples of the outputed json object.

mvuets:
@kylebrandt, here is an example of a JSON document to be sent on the ack action:
```
{
  "text": "somebody Acknowledged 2 alerts with notes \"I took care of this\", incident id's are [⛔42, ⚠️43]",
  "html": "<body xmlns=\"http://www.w3.org/1999/xhtml\">somebody Acknowledged 2 alerts with notes \"I took care of this\", incident id's are [⛔<a href=\"https://bosun.home/incident?id=42\">42</a>, ⚠️<a href=\"https://bosun.home/incident?id=43\">43</a>]</body>"
}
```
The format of that text string would be roughly `USERNAME $ACTION $ALERT_NUMBER alerts with notes \"$NOTE\", incident id's are [$INCIDENT_STATUS_ICON $INCIDENT_ID, ...]`.

mvuets:
I have just ran into another issue related to action templates: I cannot access alert variables (what can be accessed as `.Alert.Vars` under normal circumstances). I think your idea of turning `[]*models.IncidentState` into a type with a bunch of associated methods could help to bridge this gap too.

kylebrandt:
@mvuets Think it might be best to wait until go1.11 - looks go might have in block re-assignment then, see https://github.com/golang/go/commit/28c1ad9d35f27b3b57afff4ee78faac746a8ed0a 

I think can probably make some variation of append to be able to construct arbitrary slices. If continue + break also make into 1.11 (was slated for 1.10 but got pulled out) that should allow for a lot more options.

mvuets:
Yeah, that would be a nice templates update indeed. Too bad the release is going to be only around August )-:

What is your intuition about the other problem I've mentioned? Template data for action templates significantly varies from data for regular alerts. It often confuses our users. Moreover they do want to have access to `.Alert.Vars`, etc. I liked your proposal to add methods to `models.IncidentState`, so that could help mimic alert's `.Alert` in action's `.States[].Alert`. Could you give it a spin? Shall I open a separate ticket for this issue?

kylebrandt:
That being said I'm okay with still making it a type, so at the very least we can attach an IDs method that returns a slice of the incident ids, and maybe a few other straightforward methods.

PR #2246, opened 2018-04-26
fix user filter

korservick:
If we use **... AND (user:*)** filter in incident-filter - nothing return...
So, this this small patch fixes it and filter begin work like it written in  [help](https://bosun.org/usage#incident-filters)


korservick:
the speed of the merge is amazing :)

PR #2248, opened 2018-05-08
Update the aws sdk to version 1.2.10

johnewing1:
Updates the aws sdk to a version which support Elastic Container Service (ECS) task credentials. 
This allows bosun and scollector to be run containerised in an ECS cluster and be able to pick up the correct role to allow them to query the AWS apis.


johnewing1:
Closing in favour of #2262 which pulls in new version via mod

Issue #2065, opened 2017-04-25
Splay alerts randomly over a check period


mvuets:
**Problem.** I've got over 500 active alert checks pulling thousands of Graphite metrics. Most of the checks run every 5 min. I suspect that's what is causing regular high CPU load spikes. Arguably the load should be spread more evenly, hence leaving more resources room for those few important minutely checks.
![7runr](https://cloud.githubusercontent.com/assets/72705/25392881/2f478404-29da-11e7-980e-24d1c2e97eeb.png)

**Proposal.** I have an idea. But since I am very new to the Bosun land I want to validate it with you first. The scheduler could add a random jitter before the very first check run for each alert. Thus effectively shifting all checks forward in time for an arbitrary number of seconds and spreading the load.

E.g. Given three alerts and 5 min check period, minutely CPU load graph looks...
now: ▅▁▁▁▁▅▁▁▁▁▅▁▁▁▁
after: ▂▂▂▁▁▂▂▂▁▁▂▂▂▁▁
☺

Implementation wise it can be a one time random delay for at most `DefaultRunEvery * CheckFrequency` seconds before kicking off each check for the first time. Surely this jitter can be a system and/or rule configuration option (in case someone needs a rather deterministic schedule).

What do you think?

azhiltsov:
It's a great idea. Queries not only trashing the bosun machine, but graphite caches, and carbonapi as well.

grzkv:
Just submitted a PR with a solution to this https://github.com/bosun-monitor/bosun/pull/2249

Here is the test result that shows the effect of the patch on system load:
![patch_test](https://user-images.githubusercontent.com/3339062/39817558-c6eb4fae-539e-11e8-8481-3fff55487893.png)

PR #2249, opened 2018-05-09
Distributed alert checks to prevent high load spikes

grzkv:
This is a solution for #2065 

The idea behind this is simple. Every check run is slightly shifted so that the checks are distributed uniformly. 

For the subset of checks that run with the period `T`, a shift is added to every check. The shift ranges from `0` to `T-1`. The shifts are incremental. For example, if we have 6 checks every 5 mins (`T=5`). The shifts will be ` 0`, `1`, `2`, `3`, `4`, `0`. This way, without the patch 6 checks will happen at times `0`, and `5`; with the patch, two checks will happen at the time `0`, one at `1`, one at `2`, and so on. The total number of checks and check period stay the same.

Here is the test that shows the effect of the patch on system load. Note, that the majority of checks in this system have 5 mins period.
patch_test

### DIFF
--- a/cmd/bosun/conf/system.go
+++ b/cmd/bosun/conf/system.go
@@ -35,7 +35,7 @@ type SystemConf struct {
 	UnknownThreshold       int
 	CheckFrequency         Duration // Time between alert checks: 5m
 	DefaultRunEvery        int      // Default number of check intervals to run each alert: 1
-	AlertCheckDistribution bool     // Should the alert rule checks be scattered across their running period?
+	AlertCheckDistribution string   // Method to distribute alet checks. No distribution if equals ""
 
 	DBConf DBConf
 
@@ -228,7 +228,7 @@ func newSystemConf() *SystemConf {
 		CheckFrequency:         Duration{Duration: time.Minute * 5},
 		DefaultRunEvery:        1,
 		HTTPListen:             defaultHTTPListen,
-		AlertCheckDistribution: false,
+		AlertCheckDistribution: "",
 		DBConf: DBConf{
 			LedisDir:      "ledis_data",
 			LedisBindAddr: "127.0.0.1:9565",
@@ -269,6 +269,10 @@ func loadSystemConfig(conf string, isFileName bool) (*SystemConf, error) {
 	}
 	if len(decodeMeta.Undecoded()) > 0 {
 		return sc, fmt.Errorf("undecoded fields in system configuration: %v", decodeMeta.Undecoded())
+	}
+
+	if sc.GetAlertCheckDistribution() != "" && sc.GetAlertCheckDistribution() != "simple" {
+		return sc, fmt.Errorf("invalid value %v for AlertCheckDistribution", sc.GetAlertCheckDistribution())
 	}
 
 	// iterate over each hosts
@@ -404,7 +408,7 @@ func (sc *SystemConf) GetDefaultRunEvery() int {
 }
 
 // GetAlertCheckDistribution returns if the alert rule checks are scattered over check period
-func (sc *SystemConf) GetAlertCheckDistribution() bool {
+func (sc *SystemConf) GetAlertCheckDistribution() string {
 	return sc.AlertCheckDistribution
 }
 
--- a/cmd/bosun/conf/system.go
+++ b/cmd/bosun/conf/system.go
@@ -32,9 +32,10 @@ type SystemConf struct {
 	InternetProxy string
 	MinGroupSize  int
 
-	UnknownThreshold int
-	CheckFrequency   Duration // Time between alert checks: 5m
-	DefaultRunEvery  int      // Default number of check intervals to run each alert: 1
+	UnknownThreshold       int
+	CheckFrequency         Duration // Time between alert checks: 5m
+	DefaultRunEvery        int      // Default number of check intervals to run each alert: 1
+	AlertCheckDistribution bool     // Should the alert rule checks be scattered across their running period?
 
 	DBConf DBConf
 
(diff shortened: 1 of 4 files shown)


kylebrandt:
@captncraig any idea what this might do in terms of dependencies and unknown grouping? I think there may be instances where we assume things are running the same time slice.

cmd/bosun/sched/alertRunner.go
```diff
@@ -19,16 +19,25 @@ func (s *Schedule) Run() error {
...
 		ch     chan<- *checkContext
 		modulo int
+		shift  int // used to distribute alert runs
 	}
 	chs := []alertCh{}
+
+	// Every alert gets a small shift in time.
+	// This way the alerts with the same period are not fired
+	// simultaneously, but are distributed.
+	circular_shifts := make(map[int]int)
```
captncraig: Perhaps just a note here about what the keys and values represent in this map. I believe it is "run every value" to the next offset. Probably just comment that.

captncraig:
This seems like a really nice solution. I think we need a configuration value to enable (or maybe disable it). Sometimes when developing, it is extremely useful for checks to run immediately when the app starts up.

But I'm not sure if that use case is significant enough to make your behaviour an option you need to enable, or if it should be the default. I'll let @kylebrandt  make the call on that.

captncraig:
@kylebrandt has a great point I didn't consider about dependencies. Let me dig a bit on that.

cmd/bosun/sched/alertRunner.go
grzkv: The map is `run period` → `time shift to add`.
I will add a comment.
Thanks for the feedback!

captncraig:
Unknowns will work fine. That is just based on the last event timestamp. As long as it runs consistently every N times, it will be fine.

captncraig:
And it looks like the uneveluated / dependency things are run straight out of the data store from the last known run. There's a bit of a race condition involved there, so I wouldn't expect this to "break" that. Staggerubf could either increase or decrease the response time for dependency status flowing to other alerts (depending on which side runs faster), but I don't think this change will break things.

mvuets:
Would be great to have this functionality merged! Any way to help to make it happen? (-:

kylebrandt:
I'm for merging this once there is a configuration option (via the toml file) to enable and disable (default state disabled).

mvuets:
Ah, fair enough. What would you like it to be named?

grzkv:
@kylebrandt Re point 2. What exactly do you mean by jitter? Random spreading?

Or do you mean, have even more flattened spread: not only over integer cycles given by `GetCheckFrequency()`?

grzkv:
@kylebrandt Re point 1. We can make it that for the first time everything runs right away, and the shifts are applied to all subsequent runs. This will increase the complexity of code of course. Would that be better?

grzkv:
I will add the config option. Thanks for the feedback.

kylebrandt:
@grzkv I haven't fully thought it out yet. But for example if we have a Check frequency of 1 minute, checks are spread-out throughout the minute. If the Runevery is 5 minutes, the checks will be spread somewhere within a minute of that 5 minute cycle...

Here is how I think of it in the big picture:
The objective is to reduce concurrent sessions within the system so it doesn't overload Bosun and/or (more likely) the Time series database(s) used by Bosun. This is because we don't want so many queries at once that queries end up taking longer because of the amount of concurrency, and we also don't want to be impacting other users of the TSDB (i.e. Grafana or other dashboard systems, or whatever).

So the ideal way to spread out the checks depends generally on the duration of your queries a (little's law - see https://serverfault.com/a/345446/2561 for picture). If you checks are every minute, but your queries respond in say an average of 5 seconds (when not being a bunch all at once) then it would be better to spread them out through the minute, instead of hitting the system with less checks, but still in all the same time resulting in higher than ideal concurrency.

kylebrandt:
However, if this code is working well I don't want to block getting it in. We should just make the option a validated string, so if we want to try other things later the config won't change from a bool to a different type.

grzkv:
@kylebrandt You have inspired me to start reading about scheduling strategies 😄

I 100% agree with your reasoning. I thought something along the same lines. Let's proceed with the current solution for now.

I have some ideas on how we can improve further. I will share them after we merge this.

earwin:
For practical purposes, running queries sequentially without any kind of time distribution is almost always enough to deal with concurrency concerns.

Even assuming the working set for TSDB fits in memory, you're still not using up more than a single core — I didn't hear of any TSDB bothering to parallelize individual queries.
Yes, it does not look pretty on the graphs, they are spikey, but why should anyone care?

captncraig:
We also rely on fairly regular check times for cache usage on tsdb queries.
We could potentially scatter the checks and still switch out the cache at
fixed regular intervals, but I think this approach accomplishes the goal
rarther well.

kylebrandt:
IIRC within time slices we still set the time of the query to align with the runinterval time regardless of when it actually runs. have we checked that with this method when we run something a whole runinterval offset that the query time is not to the previous runinterval?

I haven't looked at that code in a while...

grzkv:
@kylebrandt Could you please elaborate on what additional checks need to be made?

cmd/bosun/conf/system.go
```diff
@@ -32,9 +32,10 @@ type SystemConf struct {
 	InternetProxy string
 	MinGroupSize  int
 
-	UnknownThreshold int
-	CheckFrequency   Duration // Time between alert checks: 5m
-	DefaultRunEvery  int      // Default number of check intervals to run each alert: 1
+	UnknownThreshold       int
+	CheckFrequency         Duration // Time between alert checks: 5m
+	DefaultRunEvery        int      // Default number of check intervals to run each alert: 1
+	AlertCheckDistribution bool     // Should the alert rule checks be scattered across their running period?
```
kylebrandt: I think this should be a string so we could implement other types of alert check distribution. Then add something to loadSystemConfig in cmd/bosun/conf/system.go to make sure it is a valid option. Valid string would be "" for default, and whatever you name this one.

grzkv: Done that. Check it out, please.

grzkv:
Changed config option to string type.
@kylebrandt please check it out.

captncraig:
So now you enable this by setting `AlertCheckDistribution  = simple` ? Seems reasonable. I'm good with this pending your ok @kylebrandt

kylebrandt:
I think just an update to the system config docs and it should be ready to merge.

grzkv:
@kylebrandt Added the docs.

mvuets:
Hi @kylebrandt! This PR is just a tiny bit away from getting merged. What can we do to help making it happen? (-:

PR #2254, opened 2018-05-17
cmd/bosun: add percentiles to window func

kylebrandt:


cmd/bosun/expr/tsdb.go
```diff
@@ -190,7 +203,11 @@ func Window(e *State, T miniprofiler.Timer, query, duration, period string, num
 				},
 			},
 		}
-		fnResult := windowFn.Call([]reflect.Value{reflect.ValueOf(e), reflect.ValueOf(T), reflect.ValueOf(callResult)})
+		fnArgs := []reflect.Value{reflect.ValueOf(e), reflect.ValueOf(T), reflect.ValueOf(callResult)}
+		if isPerc {
+			fnArgs = append(fnArgs, reflect.ValueOf(fromScalar(percValue)))
+		}
+		fnResult := windowFn.Call(fnArgs)
```
captncraig: Where is the code that actually detects the new argument and applies the percentile? I'm still confused what windowFn.Call does differently with the new parameter.

captncraig: Oh, nevermind. It actually takes another arg. And gets selected by name and reflection above. Gotcha. LGTM

Issue #2256, opened 2018-05-23
scollector - Configurable Elastic Host & Port options


chriswhite199:
Can the scollector options be extended to support configurable Elastic hostname and port values.

At the moment it only supports connections via localhost:9200, and it the Elastic instance may be listening on a specific eth interface (hence localhost may not work) or a non standard port (where there maybe multiple nodes running on the same host).

PR #2257, opened 2018-05-24
cmd/scollector: Added Elastic config options

chriswhite199:
To address: #2256


PeterGrace:
Hi @chriswhite199, I'm going to see if I can help this PR along since I'd also like this functionality in scollector.

PeterGrace:
@kylebrandt @captncraig  if either of you fine humans would be so kind as to also give your +1 before someone smacks the merge button, that'd be fantabulous.

Issue #2240, opened 2018-04-12
Support Elasticsearch 6.x as a Datasource


ghost:
I see that there is an open issue and work in progress for supporting ES 5.x as a data source https://github.com/bosun-monitor/bosun/issues/1982
I would like to vote to also have support for 6.x as that is the current release version of ES.

langerma:
as far as i have testet the esv5 option it also works for esv6

jbrann:
I am unable to run current master branch vs ElasticSearch due to config problem.  See issue 2138.  Cannot ingest config

PR #2260, opened 2018-05-27
Support connecting to multiple versions elasticsearch at runtime

bom-d-van:
With this patch bosun could connect to elasticsearch servers with different versions. Supported version are: 2.x, 5.x, 6.x. This is a **breaking** change, requiring users to update their configuration files to explicitly specify elasticsearch version with 2, 5, or 6, otherwise it would just panic (fail fast).

No compile and test errors on my machine, but I haven't test it against real elastic search servers. Publish it here to check out the community's opinions and the possibility of merging onto master. :)

Relevant tickets: https://github.com/bosun-monitor/bosun/issues/2240, https://github.com/bosun-monitor/bosun/issues/1982


bom-d-van:
Hi, the changes has been tested (thanks @pradeepbbl). And it seems working fine. Also updated doc for how to update config files.

stepashka:
Hey @kylebrandt, @captncraig!
Thanks,
Anna

kylebrandt:
There was some refactoring that has been done that will require this to be updated (how miniprofiler is done).

I'm not entirely clear on how the config is a broking change based on the documentation - is it only the Version field or is there more to it?

Since the vendoring is so big, and 2.x is EOL (see https://www.elastic.co/support/eol) can we yank that?

Supporting Multiple ES version will be nice and make upgrading easier on users.

bom-d-van:
Hi @kylebrandt 

> There was some refactoring that has been done that will require this to be updated (how miniprofiler is done).

No problem. I could rebase master and update it one more time.

> I'm not entirely clear on how the config is a broking change based on the documentation - is it only the Version field or is there more to it?

Because before this change, bosun could only talk to one version of es server but now with multiple version support if we default it to any of the versions (2/5/6), it might break the other two when the es query is run and wonder what breaks it, arguably harder to debug for users. (because bosun would use the wrong version of es client).

Therefore I explicitly panic the program on start up so users know that they have to update the configuration after upgrading the program.

bom-d-van:
A bit late but really happy maintainers finally come to this PR.

bom-d-van:
Pushed the update and all seems working (no compile or test errors). Didn't drop v2 support yet because I am not sure if all maintainers are onboard with it.

kylebrandt:
@stepashka Does Booking still care about Bosun support ES 2.x?

stepashka:
Thank you for checking, @kylebrandt! :)

Yes, we still have couple of ES 2.* clusters :(

cmd/bosun/conf/system_elastic_all.go
```diff
@@ -0,0 +1,49 @@
...
+		var cfg expr.ElasticConfig
+		switch expr.ESVersion(value.Version) {
+		case expr.ESV2:
+			cfg = parseESConfig2(value)
+		case expr.ESV5:
+			cfg = parseESConfig5(value)
+		case expr.ESV6:
+			cfg = parseESConfig6(value)
+		default:
+			panic(fmt.Errorf("conf: invalid elastic version: %s (supported versions are: v2, v5, v6)", value.Version))
```
kylebrandt: slog.Fatal is more appropriate (exit with error code, no stack trace). panic is more for programming errors and not user configuration errors.

cmd/bosun/conf/system_elastic_all.go
```diff
@@ -0,0 +1,49 @@
...
+	var cfg expr.ElasticConfig
+	switch expr.ESVersion(sc.AnnotateConf.Version) {
+	case expr.ESV2:
+		cfg = parseESConfig2(ElasticConf(sc.AnnotateConf))
+	case expr.ESV5:
+		cfg = parseESConfig5(ElasticConf(sc.AnnotateConf))
+	case expr.ESV6:
+		cfg = parseESConfig6(ElasticConf(sc.AnnotateConf))
+	default:
+		panic(fmt.Errorf("conf: invalid elastic version: %s (supported versions are: v2, v5, v6)", sc.AnnotateConf.Version))
```
kylebrandt: same here with panic vs  fatal

kylebrandt:
Can bump the middle the number of Version in version.go since this will be a breaking change.

cmd/bosun/conf/system_elastic_all.go
```diff
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	"bosun.org/cmd/bosun/expr"
+	"github.com/bosun-monitor/bosun/slog"
```
kylebrandt: import here needs to be bosun.org/bosun/slog

bom-d-van: Typo? You mean to write "bosun.org/slog", right?

kylebrandt: oops yes

kylebrandt:
I noticed that the Fatal call doesn't happen until you call an elastic. If we are going to have that, I think it chould be in cmd/bosun/conf/system.go loadSystemConfig so it happens at start.

bom-d-van:
Nice catch. I decided to do the pre-check in main because the get APIs don't return errors and loadSystemConfig does.

cmd/bosun/conf/system_elastic_all.go
```diff
@@ -20,7 +21,7 @@ func parseESConfig(sc *SystemConf) expr.ElasticHosts {
 		case expr.ESV6:
 			cfg = parseESConfig6(value)
 		default:
-			panic(fmt.Errorf("conf: invalid elastic version: %s (supported versions are: v2, v5, v6)", value.Version))
+			slog.Fatal(fmt.Errorf("conf: invalid elastic version: %s (supported versions are: v2, v5, v6)", value.Version))
```
kylebrandt: For this and the annotate error we should include:
- If it is for annotate or elastic, and also in the case of elastic which prefix it is (i.e. "default" or another named instance).
- I think since this is a breaking change, we should add a case for an empty version and return an error message saying that is a required parameter (including the info from the above bullet point)

cmd/bosun/conf/system_elastic_all.go
```diff
@@ -0,0 +1,57 @@
...
+	case expr.ESV2:
+		cfg = parseESConfig2(ElasticConf(sc.AnnotateConf))
+	case expr.ESV5:
+		cfg = parseESConfig5(ElasticConf(sc.AnnotateConf))
+	case expr.ESV6:
+		cfg = parseESConfig6(ElasticConf(sc.AnnotateConf))
+	case "":
+		slog.Fatal(fmt.Errorf("conf: [AnnotateConf]: Version is required (supported versions are: v2, v5, v6)"))
+	default:
+		slog.Fatal(fmt.Errorf("conf: [AnnotateConf]: invalid elastic version: %s (supported versions are: v2, v5, v6)", value.Version))
```
bom-d-van: @kylebrandt how about this for the error message?

kylebrandt: Good, I would just change `Version is required a field (support values for Version are: "v2", "v5", and "v5")`

Because: "Version" is a pretty generic word, so clarify that is a field, and since the values are strings, putting them in quotes maybe helps prevent a future Toml error.

kylebrandt:
The last thing on my list is optional (and kind of annoying request).

That is basically to create a new PR with two commits:

 - The vendor Commit
 - The non-vendor changes

The reason is that:
 1. It could be rebased, but then when rebased on master these commits will show up old in the commit log (unless you do some extra git trickery with timestamps.
 2. If I just squash it all into one then it is a giant commit due to the vendoring

I'm saying it is optional because I really appreciate your hard work and patience. So if you don't have the time for copying around all the files or git magic to rewrite timestamps with a rebase, then I am fine accepting option 2 and just squashing it all into one.

bom-d-van:
@kylebrandt No worries, I have pushed the changes and created another commit. Hope it's done as expected.  :)

bom-d-van:
Really thank you for the review too. It's lots of work and nice suggestions!

kylebrandt:
done via https://github.com/bosun-monitor/bosun/pull/2316

PR #2264, opened 2018-06-07
add new function sortbytags

michep:


kylebrandt:
I think there should be a 2 to 3 additional arguments:
 - Tag Keys preference, csv, ie "host,interface" sort by host's values, then interface value
 - "asc, desc" or whatever was used for `sort()`
 - This one less sure about, but may want an option for alpha/num sort types?

PR #2266, opened 2018-06-12
cmd/bosun replace textarea with aceeditor in expr

KozzyKoder:
Hello,

We have UX troubles trying to utilize Expr editor which is basically just textarea: it doesn't have syntax highlight and monospace font.
That's why we've implemented AceEditor to expr editor.

this is reimplementation of another PR
which seems to be abandoned: https://github.com/bosun-monitor/bosun/pull/2201

Thanks!


kylebrandt:
I do like changing this to the ace editor, however I'd like to see two changes:

 - The expression page supports "shift-enter" to execute the expression, would like to keep that ability
 - I think the default height should be a little smaller (although probably a few more lines then current) so by default there is more screen real estate for results.

kylebrandt:
Oh, also the bosun logo animation seems to keep going until you run an expression, I think there is stop JS function to stop that on load

KozzyKoder:
@kylebrandt fixed review issues

PR #2269, opened 2018-06-18
Added extra info to unknown templates

pradeepbbl:
Currently, we have a very limited number of variable available to the unknown templates which prevent us to match the alert and action notification template (includes incident states) and sometimes unknowns are getting unnoticed due inconsistency (when using external notification systems e.g. pagerduty).

With this patch, I tried to include the alert states to unknown which is very useful when grouping disable using [unknownMinGroupSize](http://bosun.org/definitions#unknownmingroupsize).

Please let me know your thoughts.

### DIFF
--- a/cmd/bosun/conf/unknownNotify.go
+++ b/cmd/bosun/conf/unknownNotify.go
@@ -3,6 +3,7 @@ package conf
 import (
 	"bytes"
 	"fmt"
+	"net/url"
 	"time"
 
 	"bosun.org/cmd/bosun/conf/template"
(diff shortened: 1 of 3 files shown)


pradeepbbl:
here is an example of a PagerDuty alert body template 

```
pdMonitoringBody = `{
       "service_key": "XXXXXXXXXXXXXX",
       "event_type": "trigger",
       "description": "{{.Subject | jsonStr}}",
       "client": "bosun",
       "client_url": "{{.Incident}}",
       "incident_key": "AlertName={{.Alert.Name}};IncidentId={{.Id}}"
 }` 
 ```

if you make close attention to [`incident_key`](https://v2.developer.pagerduty.com/docs/events-api#incident-de-duplication-and-incident_key) we use alert name and incident id combination to reduce duplication. With current unknown template contexts/var, it's impossible for us to prepare an unknown body that matches the above specification and due to this may unknowns are getting unnoticed by teams.

With this patch, we are able to produce the unknown body and post  to PagerDuty successfully

```
pdMonitoringUnknownBody = `{
       "service_key": "XXXXXXXXXXXXX",
       "event_type": "trigger",
       "description": "Alert goes to unknown states",
       "client": "bosun",
       "client_url": "{{ .IncidentUnknownLink .States.Id }}",
       "incident_key": "AlertName={{.Name}};IncidentId={{.States.Id}}"
}`
```

@kylebrandt / @captncraig  could you please let us know your thoughts?

pradeepbbl:
can this be merge soon :)

stepashka:
Hey @kylebrandt, @captncraig!
Thanks,
Anna

PR #2271, opened 2018-06-26
cmd/bosun: added silence and status icons and ack state text on incident page

KozzyKoder:
Hello,

we found that on incident page are missing several icons that represents alert state.
On dashboard alerts and incidents has "silence" and "active" icons. So for the sake of unification we suggest to add these icons also to incident page.

also we made several improvements in incident page:
1) Show text "need acknowledgement" next to incident title when it needs.
2) Silence icon also a link that leads to edit Silence rule.

image

What your opinion about that? Thanks!


kylebrandt:
Merged in https://github.com/bosun-monitor/bosun/commit/865107d5e70241041c8351882c8f2e7cc2495e22 

(Did a rebase and go generate into new branch since the static.go required a new `go generate` - need to do something better there eventually.  You still get authorship even though this method confused github a little. I wonder if this method is a bad idea since you branch is just kind of left hanging).

Issue #2251, opened 2018-05-16
Send alert name in Graphite request header


grzkv:
Bosun can add [custom headers in Graphite requests](https://bosun.org/system_configuration#graphiteconfheaders). However, these are static.

We need to send alert names as header values. This would improve diagnostics. This will also add the ability to track load alerts put on Graphite.

The main question is the syntax in the config file. The simplest way to proceed is to add some sort of "magic string" as the header value. When encountered, it will be converted to alert name. For example:
```toml
[GraphiteConf]
	Host = "localhost:80"
	[GraphiteConf.Headers]
		Static-Header = "Content"
		Alert-Name-Header = "{alert-name}"
```
`{alert-name}` will be expanded to something like `cpu.is.too.high` in the request header.

Please, share your thoughts.

mvuets:
@kylebrandt, @captncraig, do you find this idea reasonable in general? Our team is ready to implement it shall you give a green light.

captncraig:
Why not just add that globally for all outbound graphite requests? Is there any downside?

mvuets:
I should probably clarify the original question. We want each outbound HTTP query to Graphite to contain a header that identifies the origin (e.g. alert name) of the query. With a global header we can see what portion of Graphite traffic comes from Bosun, but we cannot easy tell which queries belong to what alerts. In other words we want to tag each HTTP query with a Bosun alert name in order to identify offensive (too heavy) queries on the Graphite side. Does it make sense?

grzkv:
@captncraig It makes sense. It does not seem to me as a big overhead.

johnewing1:
This would be a useful feature for opentsdb as well.

An other thought would be to pass through the referrer in an x-header as well. This would be useful for when bosun is being used in conjunction with the grafana bosun data source to help identify dashboards making heavy queries.

azhiltsov:
Setting a header will be a real life saver on very busy environments.

stepashka:
Hey @kylebrandt @captncraig !
Any objections to merging this?

gksinghjsr:
This is actually a much needed improvement. Would be nice to have this merged.

PR #2274, opened 2018-06-29
Alert name macros in Graphite request header

sipidronov:
As a followup for #2251 here's an implementation of passing Expression-specific data (alert-name in this case) into graphite context so that the config like 
`[GraphiteConf]`
`Host = "localhost:80"`
`[GraphiteConf.Headers]`
`X-Meow = "Mix {alert-name}"`

Results in the following request to Graphite for "alert os.super-duper-high.cpu":
`GET /render/?format=json&from=1530262155&target=%2A.cpu.%2A.cpu.idle%29&until=1530262455 HTTP/1.1`
`Host: localhost:80`
`User-Agent: Bosun/0.7.0-dev`
`X-Bosun-Server: c02w708ghv2t`
`X-Meow: Mix os.super-duper-high.cpu`
`Accept-Encoding: gzip`

If the machinery looks good I'd expect macros substitution to be extended and optimised.


stepashka:
Hey Alexander @sipidronov!
This patch seems to have conflicts with upstream, could you review these changes by any chance?
Thanks,
Anna

sipidronov:
Will re-submit PR.

PR #2277, opened 2018-07-04
Fix delay close alert notification

pradeepbbl:
We have noticed bosun sending [`actionBodyClose`](http://bosun.org/notifications) instead of [`actionBodyDelayedClose`](http://bosun.org/notifications) when alerts are getting delayed closed. This is causing confusion because users are getting closed notifications immediately whereas it's still active in bosun, also when bosun actually closes the alert (after delay timeout) it's not sending any notifications and can only be viewed in the incident page.

In this pull request we try to solve both problems (sending delay close and notify when actual close happen), could you please review and let us know your thoughts?

### DIFF
--- a/cmd/bosun/sched/check.go
+++ b/cmd/bosun/sched/check.go
@@ -147,10 +147,18 @@ func (s *Schedule) runHistory(r *RunHistory, ak models.AlertKey, event *models.E
 					if err != nil {
 						return
 					}
+
 					incident, err = data.GetIncidentState(incident.Id)
 					if err != nil {
 						return
 					}
+
+					// notify action cancel close
+					err = s.ActionNotify(models.ActionCancelClose, "bosun", "cancelled delayed close due to severity increase", []models.AlertKey{ak})
+					if err != nil {
+						return
+					}
+
 					// Continue processing alert after cancelling the delayed close
 					break
 				}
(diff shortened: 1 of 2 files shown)


pradeepbbl:
@kylebrandt / @captncraig  could you please review this pull request

stepashka:
Hey @kylebrandt, @captncraig!
Thanks,
Anna

kylebrandt:
Probably need @captncraig to review the behavior on this one.

In terms of how the default behavior was supposed to go in my head, at least in regard to emails:

 - DelayedClose means "I'm pretty sure I have fixed the incident because I did x, y, z." The notification happens then because that is what other people (at least in my story I constructed :-/) need to know - the things has probably been dealt with and the monitoring system just needs to catch up.
 - From there, you can just assume it is closed, if the problem still exists after the close duration then a new incident will be created

At least one problem with the story is that notifications (at least how I have constructed them) don't give you any information about the previous incident on the same alertkey. So it the person getting the new incident might not know about the previous one. 

I'm okay customizing this behavior I'm just worried about changing default behavior here and maybe adding some complexity. Basically the falls under the idea that "less notifications are better"  - but if @captncraig thinks the complexity is fine I'm okay with being able to customize this behavior.

svagner:
@tdinucci, @muffix We are using this change more than 1 year already. Can you please review this MR?

tdinucci:
@svagner we will consider but as no test coverage cannot accept as it stands

svagner:
@tdinucci , I agree there is really required some amount of tests. I'll make tests then:) Thanks!

PR #2290, opened 2018-08-08
cmd/bosun: change azure to use id from resource queries

kylebrandt:


cmd/bosun/expr/azure.go
```diff
@@ -188,7 +186,7 @@ func AzureQuery(prefix string, e *State, T miniprofiler.Timer, namespace, metric
 			"",
 			filter,
 			insights.Data,
-			namespace)
```
kylebrandt: Oh an turns out this is just an optional param, that is why this line is changed

PR #2299, opened 2018-09-04
Improve logs

svagner:
# problem
Sometimes people would complain that notifications about close/ack were not delivered to them
it seems that it's not possible to easily debug that now with current state of logs.

There is no log entries about successfull sending acks and closes. Only log failures (e.g. 400 response)
- notifications option 'print = true' also seems to not do anything for action notifications (even for failed notifications that are logged)
- logs don't allow you to understand if 'Send notifications' checkbox was on or off

# goal
- log successful actions (similar to how it logs alert notifications)
- print new logs according to 'notification' settings
- improve error log format - write extra data about notification that has delivery problem
- log actions if there wasn't set checkbox notification

# changes:
- add logging if there was action without notification
- add logging for any actions/alerts
- add extra felds to log messages

### DIFF
--- a/cmd/bosun/conf/notify.go
+++ b/cmd/bosun/conf/notify.go
@@ -20,6 +20,12 @@ import (
 	"github.com/jordan-wright/email"
 )
 
+const (
+	sendLogSuccessFmt = "%s; name: %s; transport: %s; dst: %s; body: %s"
+	sendLogErrorFmt   = "%s; name: %s; transport: %s; dst: %s; body: %s; error: %s"
+	httpSendErrorFmt  = "bad response for '%s' %s notification using template key '%s' for alert keys %v method %s: %d"
+)
+
 func init() {
 	metadata.AddMetricMeta(
 		"bosun.email.sent", metadata.Counter, metadata.PerSecond,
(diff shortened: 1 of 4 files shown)


stepashka:
Hey @kylebrandt, @captncraig!
Thanks,
Anna

cmd/bosun/conf/notify.go
```diff
@@ -38,33 +38,36 @@ func init() {
...
+	Print  bool
+	Name   string
 	Errors []string
 }
 
 func (p *PreparedNotifications) Send(c SystemConfProvider) (errs []error) {
 	if p.Email != nil {
 		if err := p.Email.Send(c); err != nil {
-			slog.Errorf("sending email: %s", err)
+			slog.Errorln("status: error; name: " + p.Name + "; subject: " + p.Email.Subject + "; transport: email; address: " + strings.Join(p.Email.To, ",") + "; body: " + p.Email.Body + "; error: " + err.Error())
```
kylebrandt: slog.Errorf instead, and then if the fmt strings are the same you can put the in a higher level var or a constant

cmd/bosun/web/web.go
```diff
@@ -789,6 +789,8 @@ func Action(t miniprofiler.Timer, w http.ResponseWriter, r *http.Request) (inter
 		if err != nil {
 			return nil, err
 		}
+	} else {
+		slog.Infoln(fmt.Sprintf("action without notification. user: %s, type: %s, keys: %v, ids: %v", data.User, data.Type, data.Keys, data.Ids))
```
kylebrandt: slog.Infof

cmd/bosun/conf/notify.go
svagner: Hi! Thanks for advice. I've also aggregated format for error string generation and moved this format to const

cmd/bosun/web/web.go
svagner: Changed, thanks!

svagner:
Hi @kylebrandt ! Thanks for your comments. I've pushed changes. Could you please review again? Thanks!

PR #2300, opened 2018-09-04
docs: transpose t() explanation

kylebrandt:
from the stackoverflow documentation archive: https://archive.org/details/documentation-dump.7z


kylebrandt:
still need to link from expr page

PR #2303, opened 2018-09-06
Fix opentsdb tags in aggr function

hermanschaaf:
Turns out I was doing the Tags function wrong, and tags should have `=*` like in this PR. 

Without this fix, math operations cause an error like `expr: opentsdb: bad tag: color`. The tests included here were failing before the fix with such an error.

CC @kylebrandt


hermanschaaf:
Sorry, I did a bad merge when adding an additional fix and got this branch in a bad state. Will re-open the PR shortly

hermanschaaf:
Re-opened here with additional fixes and some refactoring: https://github.com/bosun-monitor/bosun/pull/2306

PR #2306, opened 2018-09-11
Fix math operations on results of aggr

hermanschaaf:
This PR fixes a bug in `aggr` that causes math operations on the resulting series to fail. It also adds tests for such cases, and refactors the `aggr` tests to be [table-driven](https://github.com/golang/go/wiki/TableDrivenTests).

Without this fix, math operations cause an error like expr: opentsdb: bad tag: color. The tests included here were failing before the fix with such an error.

CC @kylebrandt

### DIFF
--- a/cmd/bosun/expr/expr_test.go
+++ b/cmd/bosun/expr/expr_test.go
@@ -359,7 +359,7 @@ func TestSeriesOperations(t *testing.T) {
 				Results: ResultSlice{
 					&Result{
 						Value: Series{
-						// Should be empty
+							// Should be empty
 						},
 						Group: opentsdb.TagSet{"key": "a"},
 					},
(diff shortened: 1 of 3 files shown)


hermanschaaf:
The tests are failing due to timeouts, doesn't seem to have any real issues.

hermanschaaf:
@kylebrandt how does this look to you? Would be nice to get the fix in 🙇

kylebrandt:
Looks like one of the test files just needs a gofmt

hermanschaaf:
Or not... maybe I need to upgrade to Go1.10 on this machine

hermanschaaf:
Alright, it's fixed now  👍 Sorry for the delay

Issue #2297, opened 2018-08-29
Inconsistent behaviour with over() compared to bandTSDB() and window()


DCBoland:
Using `over()` results in a seriesSet containing a series starting now, i.e. `{shift=0s}`. This is unexpected and inconsistent with the behaviour of `bandTSDB()` and `window()`, which start `period` ago. This is an issue when trying to calculate a historical distribution, median etc. using the new `aggr()` function, as the current value will be included.

The above inconsistency appears to be because `over()` subtracts `period` from now at the end of its loop, whereas the other functions do the subtraction at the start of their loops:
https://github.com/bosun-monitor/bosun/blob/4bec6475bb8f41f0f683910b90c11c8ab4a633e3/cmd/bosun/expr/tsdb.go#L386

Potential solutions:
- Change `over()`'s behaviour to be in line with the other functions
- Add parameters to set a start time or offset.

I'm assuming the first option would cause too many problems to be acceptable?

Context:
[aggr() function PR](https://github.com/bosun-monitor/bosun/pull/2294)

kylebrandt:
`over()` was created for making week over week graphs, for which you want the current week. So I don't think we should change the behavior.

So instead I think we should just add another function that behaves like band and window, I'm not sure what to call it.

hermanschaaf:
If we do create another function, can we add another parameter for explicitly specifying the end time, rather than using `period` for this? This would allow us to more easily account for minor metric delays, and also achieve what @DCBoland described. Though we may need one like this for `band` and another for `over` (with time shifting included)

DCBoland:
While looking at the above, I spotted the undocumented `shiftBand` function, which appears to return an `over`-like seriesSet, with the `period` shift behaviour from `band` – exactly what I needed!

The above commit is along the lines of @hermanschaaf's suggestion. Adds an `eduration` parameter to generic versions of `band` and `over`, named `bandQuery` and `overQuery`, to set the end time as is done with `Query`.  The original functions just wrap these new ones, with `period` given as the `eduration` to match the existing behaviour.

kylebrandt:
@DCBoland Nice I like it, adds what people want without messing with current behavior (or at least it shouldn't). 

Needs expr documentation in docs.md (can you add for the shiftBand ... easter egg :p ... as well?)

Then if you add some tests it will be easier for me to look over the code and merge it since I don't have test things locally as much.

PR #2310, opened 2018-09-19
Add tsdb query functions with end duration parameter.

DCBoland:
This PR adds two functions for querying against openTSDB – `overQuery` and `bandQuery`. These functions slightly generalise the existing `over` and `band` functions to include an end duration parameter `eduration`, which specifies when a query should end, as is done with `q`. This enables queries for a historical distribution that does not include recent outliers – related discussion in #2297 .

The existing `over` and `band` functions are changed to wrap the more general functions. The undocumented and very related `shiftBand` function was found during this work, and is added to the documentation.

The third commit f61259a refactors the functions to reuse `bandTSDB`, as `shiftBand` used this function to achieve comparable functionality to `over`.

Tests were written based on the existing function of `over`, `band` and `shiftBand` to ensure same behaviour remains.

### DIFF
--- a/cmd/bosun/expr/expr_test.go
+++ b/cmd/bosun/expr/expr_test.go
@@ -122,6 +122,19 @@ var queryTime = time.Date(2000, 1, 1, 12, 0, 0, 0, time.UTC)
 
 func TestQueryExpr(t *testing.T) {
 	queries := map[string]opentsdb.ResponseSet{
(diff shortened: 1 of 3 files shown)


kylebrandt:
cmd/bosun/expr/expr_test.go needs a fmt but other than that I think this good

PR #2312, opened 2018-09-20
cmd/bosun: ui - incident view: fix durations

kylebrandt:


kylebrandt:
before / after:

selection_066

Issue #1978, opened 2016-12-06
Update influxQueryDuration to work around issues with missing time in WHERE/GROUP BY


gbrayut:
See https://github.com/bosun-monitor/bosun/pull/1976#issuecomment-265023594 for context

```
# This should work but fails with: moving_average aggregate requires a GROUP BY interval (since that expression isn't added until after the call to influxql.ParseStatement)
influx("telegraf",'''SELECT moving_average(min(usage_idle),6) FROM cpu WHERE cpu = 'cpu-total' GROUP BY host''',"5m","","1m")

# Adding ,time(1m) to the group by causes: aggregate functions with GROUP BY time require a WHERE time clause (same issue as above)
influx("telegraf",'''SELECT moving_average(min(usage_idle),6) FROM cpu WHERE cpu = 'cpu-total' GROUP BY host,time(1m)''',"5m","","1m")
```

PR #2328, opened 2018-10-08
Fix influxdb nested functions

krise3k:
Fix https://github.com/bosun-monitor/bosun/issues/1978 by updating influx client to `v1.6.3` - latest stable version

`influxql` has been moved to separate package so I get the same version as is used in influx - https://github.com/influxdata/influxdb/blob/v1.6.3/Gopkg.toml#L35


arwineap:
This solves https://github.com/bosun-monitor/bosun/issues/2148 for me. It also seems stable  :+1:

PR #2338, opened 2018-10-29
cmd/scollector: httpunit - rescode metric and url metadata

kylebrandt:
also checks nils to avoid possibility of panics

### DIFF
--- a/cmd/scollector/collectors/httpunit.go
+++ b/cmd/scollector/collectors/httpunit.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"time"
 
+	"bosun.org/slog"
+
 	"bosun.org/metadata"
 	"bosun.org/opentsdb"
 	"github.com/BurntSushi/toml"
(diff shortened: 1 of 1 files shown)


cmd/scollector/collectors/httpunit.go
```diff
@@ -46,16 +48,32 @@ func cHTTPUnit(plans *httpunit.Plans) (opentsdb.MultiDataPoint, error) {
...
+		}
 		tags := opentsdb.TagSet{
 			"protocol":     r.Case.URL.Scheme,
 			"ip":           r.Case.IP.String(),
 			"url_host":     r.Case.URL.Host,
 			"hc_test_case": r.Plan.Label,
 		}
+		metadata.AddMeta("", opentsdb.TagSet{"hc_test_case": r.Plan.Label}, "url", r.Case.URL.String(), false)
+		if r.Result == nil || r.Result.Result != nil {
+			Add(&md, "hu.error", true, tags, metadata.Gauge, metadata.Bool, descHTTPUnitError)
```
captncraig: Is there a better name for this metric than `hu.error`? Seems unintuitive what 0/1 mean. Maybe its just me.

cmd/scollector/collectors/httpunit.go
```diff
@@ -46,16 +48,32 @@ func cHTTPUnit(plans *httpunit.Plans) (opentsdb.MultiDataPoint, error) {
...
+			continue
+		}
 		tags := opentsdb.TagSet{
 			"protocol":     r.Case.URL.Scheme,
 			"ip":           r.Case.IP.String(),
 			"url_host":     r.Case.URL.Host,
 			"hc_test_case": r.Plan.Label,
 		}
+		metadata.AddMeta("", opentsdb.TagSet{"hc_test_case": r.Plan.Label}, "url", r.Case.URL.String(), false)
+		if r.Result == nil || r.Result.Result != nil {
```
captncraig: what is this logic? If the result is nil or the result's result is not nil? Seems odd to set error and not continue in some cases, which then unsets error? Seems odd.

cmd/scollector/collectors/httpunit.go
captncraig: Also no metadata.

kylebrandt: Existing metric I'm not looking to rename at the moment.

kylebrandt: I'm not seeing what you referencing with "no metadata?"

captncraig: nvm me. I forget how things work.

Issue #2335, opened 2018-10-18
Can't use ledis with 0.8-preview because of migrations


johnseekins:
Every time I restart bosun using the new 0.8 preview and a ledis backend (since we're just testing), Bosun refuses to start until I've deleted the `ledis_data` folder.

```
2018/10/18 18:47:19 info: main.go:297: Interrupt: closing down...
2018/10/18 18:47:19 info: alertRunner.go:85: Stopping alert routine for zfsarcstats
2018/10/18 18:47:19 info: alertRunner.go:85: Stopping alert routine for load
2018/10/18 18:47:19 info: main.go:299: done
root@use1-hadoop-5:~/bosun# ./bosun 
2018/10/18 18:47:27 enabling syslog
2018/10/18 18:47:27 info: migrate.go:137: checking migrations
2018/10/18 18:47:28 info: migrate.go:85: Adding fields for previous incidents and next incident on all incidents in order to link incidents together. This is a one time operation that can take several minutes.
2018/10/18 18:47:28 info: migrate.go:88: migrating 0 incidents
2018/10/18 18:47:28 fatal: main.go:153: state_data.go:287:  command not found
root@use1-hadoop-5:~/bosun# rm -rf ledis_data/
root@use1-hadoop-5:~/bosun# ./bosun 
2018/10/18 18:47:34 enabling syslog
2018/10/18 18:47:34 info: migrate.go:137: checking migrations
2018/10/18 18:47:34 info: migrate.go:149: schema version not found in db
2018/10/18 18:47:34 info: migrate.go:151: assuming new installation because allIncidents key not found
2018/10/18 18:47:34 info: migrate.go:152: writing current schema version
2018/10/18 18:47:34 info: search.go:194: Loading last datapoints from redis
2018/10/18 18:47:34 error: search.go:197: search_data.go:189: redigo: nil returned
2018/10/18 18:47:34 info: search.go:201: Done
```

seanlook:
#2324 

I encounter the same issue when I build bosun myself and not resolved yet

kylebrandt:
@johnseekins To be clear, does it only happen until you delete previous data from a previous version of Bosun. Or does it happen when you start with a fresh db, and then restart again?

kylebrandt:
I think there are two levels of issues:
[...]

PR #2339, opened 2018-10-30
cmd/bosun: bump db schema version

kylebrandt:
this should have been done as part of the inital migration commit. #2335

### DIFF
--- a/cmd/bosun/database/database.go
+++ b/cmd/bosun/database/database.go
@@ -19,7 +19,7 @@ import (
 	"github.com/captncraig/easyauth/providers/token/redisStore"
 )
 
-var SchemaVersion = int64(1)
+var SchemaVersion = int64(2)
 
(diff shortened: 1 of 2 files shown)


captncraig:
Lgtm

cmd/bosun/database/database.go
```diff
@@ -19,7 +19,7 @@ import (
 	"github.com/captncraig/easyauth/providers/token/redisStore"
 )
 
-var SchemaVersion = int64(1)
+var SchemaVersion = int64(2)
```
kylebrandt: I think perhaps I could just wrap this in an init() and do the len of `tasks` for the value of `SchemaVersion`, but I have some feeling that perhaps there was a reason I didn't do this in the first place - also I don't feel that motivated to mess with it right now.

PR #2343, opened 2018-11-09
close influx http handler

edwardhey:
close the http connection, without close, will cause the influxdb cannot accept any request.


jsamaroo:
This looks similar to #2332

PR #2344, opened 2018-11-13
Added support for multi graphite data sources.

pradeepbbl:
This pull request will allow users to configure multiple Graphite backends similar to Elasticsearch. This requires a change in the existing Graphite config.

old config:
```
[GraphiteConf]
   Host = "localhost:80"
   [GraphiteConf.Headers]
           X-Meow = "Mix"
```

new config:
```
[GraphiteConf]
   [GraphiteConf.default]
       Host = "localhost:80"
       [GraphiteConf.default.Headers]
              X-Meow = "Mix"
```

Thanks,


captncraig:
So this is definitely a breaking change? I'm ok with that we just need to be careful to highlight that.

pradeepbbl:
@captncraig yup you are right it's a breaking change and error out in the startup, have also made a change to bosun config startup error to refer http://bosun.org/system_configuration for details.

kylebrandt:
@pradeepbbl I think the CacheKey() needs to include the prefix or you might get false cache hits.

pradeepbbl:
> @pradeepbbl I think the CacheKey() needs to include the prefix or you might get false cache hits.

@kylebrandt  thanks for reviewing, as suggested I have added Prefix to CacheKey.

pradeepbbl:
@kylebrandt can this be merged now?

PR #2345, opened 2018-11-13
First cluster implementation

svagner:
# Goal
The goal of this change is to add high availability for bosun service. Currently it’s possible to have only one active bosun node running checks and sending alert. If this node goes down or becomes unavailable there’s a high chance that checks won’t run for some time, alerts are not fired, incidents aren’t created and notifications — delivered. It then take some time for a sysadmin to switch all the checks to another node. So in the end we’re minimising human intervention needed in failure scenarios and reducing possible downtime of service.

# Changes overview
We implemented cluster for bosun.
Cluster would only have one ‘leader’ at a time, all other nodes are followers (so this is an implementation of a model with with 1 master and multiple standby nodes).
‘Master’ node executes the checks and sends notifications, ‘follower’ nodes don’t do neither (they run with ‘no-checks’ and ‘quiet-mode’ options enabled). This also adds a new (optional) dependency raftdb to store state and perform leader election.

# Configuration
Config example:
```
#node 1
[ClusterConf]
    	MetadataStorePath = "/var/db/bosun_raft"
    	RPCListen = "10.0.0.1:10000"
```
```
#node 2
[ClusterConf]
    	MetadataStorePath = "/var/db/bosun_raft"
    	RPCListen = "10.0.0.2:10000"
    	Members = ["10.0.0.1:10000"]
```
```
#node 3
[ClusterConf]
    	MetadataStorePath = "/var/db/bosun_raft"
    	RPCListen = "10.0.0.3:10000"
    	Members = ["10.0.0.1:10000"]
# or
# Members = ["10.0.0.1:10000", "10.0.0.2:10000", "10.0.0.3:10000"]
```
- MetadataStorePath - is a path to store cluster data e.g. raft database and snapshots (snapshots are empty all time in current implementation). If this path doesn’t exist we try to create it automatically
- RPCListen - host and port for serf listener (https://github.com/hashicorp/serf). Host should be the address that will be available for other nodes of the cluster (so shouldn’t be loopback address!). Port will be a primary port for cluster. There will also be another opened extra port for raft protocol (https://github.com/hashicorp/raft). Raft port will be RPCListen + 1.

Additionally an api endpoint was implemented to expose statistics and current cluster status:
```
# curl http://127.0.0.1:8070/api/cluster/status
{
  "State": 2,
  "StateString": "Leader",
  "Nodes": [
	{
  	"Address": "127.0.0.1:10022",
  	"ID": "127.0.0.1:10022",
  	"State": "Follower"
	},
	{
  	"Address": "127.0.0.1:10012",
  	"ID": "127.0.0.1:10012",
  	"State": "Leader"
	}
  ],
  "Stats": {
	"applied_index": "6",
	"commit_index": "6",
	"fsm_pending": "0",
	"last_contact": "0",
	"last_log_index": "6",
	"last_log_term": "4",
	"last_snapshot_index": "0",
	"last_snapshot_term": "0",
	"latest_configuration": "[{Suffrage:Voter ID:127.0.0.1:10022 Address:127.0.0.1:10022} {Suffrage:Voter ID:127.0.0.1:10012 Address:127.0.0.1:10012}]",
	"latest_configuration_index": "6",
	"num_peers": "1",
	"protocol_version": "3",
	"protocol_version_max": "3",
	"protocol_version_min": "0",
	"snapshot_version_max": "1",
	"snapshot_version_min": "0",
	"state": "Leader",
	"term": "4"
  }
}
```


captncraig:
I love the concept, and I'm excited to try it out. A few questions before I dive in:

- Does this do anything to sync alert rules and config between nodes? That could be a nice upgrade if possible.
- Is there any way to specify node priorities? We have two nodes currently in different data centers, and will almost always want the primary dc node to be the master.
- I assume this is like most other raft implementations in terms of consensus requirements. So we'd need 3 nodes minimum to be able to do any automated failover? Is there any tunability to the quorom requirements and so forth?

svagner:
Thanks for questions! Current realisation is just very basic first step. It means that we will not sync any data between nodes yet. Basically we are managing alert rules and config file by configuration management (puppet), and it makes configs equal. Manage of system config from application (like sync from another nodes) can be bad idea, because users can define for slaves different endpoints as instance. I also don't see normal possibility to sync alert rules between nodes with current implementation of those rules. Now we are managing rules via config file. I don't think that rewrite some config files from application is a good idea, because it can make management of configs very difficult (like it was made in redis sentinel - it is really very difficult to manage config like that from chef/ansible/puppet). May be we will think about moving rules to some db storage then?
About priorities of nodes it can be difficult, because raft algorithm - https://github.com/hashicorp/raft/issues/211. But there is exists 2 different models - wan and lan. I think for different datacenters it could work. I will take a look on this. 
It is indeed just another implementation of raft from hashicorp. Yes, for automated failover we should have at least 3 nodes. With another way it is possible to split brain:) Here is some description of this (from authors of library) - https://www.consul.io/docs/internals/consensus.html.

stepashka:
Hey @kylebrandt @captncraig !
Any objections to merging this?

svagner:
Hi @captncraig ! Is there some problem with this MR?:( We really need this, because currently it is possible to have only one node without any possibility of HA...

stepashka:
Hey, @captncraig , @kylebrandt !
Any objections to this?
Manual failover is quite inconvenient and does increase time to recover :(

stepashka:
Hey @svagner ! I've just checked and it seems this has conflicts now. Shall we resolve them?

svagner:
New implementation is in https://github.com/bosun-monitor/bosun/pull/2472

PR #2346, opened 2018-11-16
Added a new func `last` to templates,  identical to reduction functions `last()`

pradeepbbl:
This will allow users to extract the last value from a `seriesSet` in templates.

Example:

```
template last {
    body = `
       <table border="1">
              <th>Host</th>
              <th>State</th>
              {{range $s := .EvalAll .Alert.Vars.state }}
              <tr>
                  <td>{{ $s.Group.host }}</td>
                   {{ $state := last $s.Value }}
                   {{ if eq $state 0.0 }}
                   <td bgcolor="LIGHTGREEN">UP</td>
                   {{ end }}
                   {{ if eq $state 1.0 }}
                   <td bgcolor="YELLOW">UNKNOWN</td>
                   {{ end }}
                   {{ if gt $state 1.0 }}
                   <td bgcolor="RED">DEAD</td>
                   {{ end }}
             </tr>
             {{end}}
      </table>
    `
}

alert last {
    template = last
    $query = "aliasByNode(sys.host.*.status,2)"
    $state = graphite($query, "5m", "", "host")
    crit = last($state) > 1
    warn = last($state) > 0
}
```

### DIFF
--- a/cmd/bosun/conf/rule/rule.go
+++ b/cmd/bosun/conf/rule/rule.go
@@ -397,6 +397,16 @@ var defaultFuncs = template.FuncMap{
 	"pct": func(i interface{}) string {
 		return fmt.Sprintf("%.2f%%", i)
 	},
+	"last": func(series expr.Series) (a float64) {
+		var last time.Time
+		for k, v := range series {
+			if k.After(last) {
+				a = v
+				last = k
+			}
+		}
+		return
+	},
 	"replace": strings.Replace,
 	"short": func(v string) string {
 		return strings.SplitN(v, ".", 2)[0]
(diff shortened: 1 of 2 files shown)


kylebrandt:
Since you could want other things besides the last value, this could add a lot of template funcs.

Generally the pattern I use is to just set a variable, `$lastState = last($state)` and then just `.EvalAll .Alert.Vars.lastState` - any reason that doesn't work for your cases?

pradeepbbl:
It make more sense to me and I agree with your point on adding junk/duplicated func to temaplate. I tested with `.Eval` instead of `.EvalAll`  which gave only last value set. I will give it a try on Monday and will close this pull request if the results were expected.

Thanks for the feedback :)

pradeepbbl:
Closing this as the workaround suggested working as expected.

PR #2358, opened 2018-12-14
Show past inactive alerts in bosun silence test results.

pradeepbbl:
Currently, when you start creating a silencing rule after you press `test` button we expect bosun to show the alerts that will be affected by that silencing rule but at present, the list is incomplete and is only showing active alerts with `open/acked` incidents but won't show alerts with no incidents i.e. past alerts.

This is sometimes confusing (i.e scheduling maintenance) users as the `test` result was empty, with this patch we tried to fix this behavior by showing past closed alerts along with open incidents also added few styling fixes to match bosun incident sates.

i.e current behavior

screen shot 2018-12-14 at 6 39 32 pm

with this patch

screen shot 2018-12-14 at 6 45 33 pm

@kylebrandt / @captncraig let us know your thoughts

Thanks,


kylebrandt:
I think this might end up being expensive on the dom and the DB since it his allIncidents. Maybe it can range back only X incidents in a cheaper redis call, and also limit too much from loading into the DOM...?

pradeepbbl:
> Hi @kylebrandt,
> 
> Thanks for the review, as pointed I have replaced the allIncidents call and will work on DOM (to limit the result set to 5)
> 
> meanwhile, let me know your thoughts?
> 
> Thanks,

the results set fixed to 10

cmd/bosun/database/state_data.go
```diff
@@ -274,6 +275,41 @@ func (d *dataAccess) GetAllIncidentIdsByAlertKey(ak models.AlertKey) ([]int64, e
...
+
+	incidents, err := redis.Strings(conn.Do("KEYS", key))
+
+	if err != nil {
+		return nil, slog.Wrap(err)
+	}
+
+	state := []*models.IncidentState{}
+
+	for _, inc := range incidents[0:10] {
```
pradeepbbl: only range over 10 item

Issue #1974, opened 2016-11-30
Remove previous container entries from Bosun which are not alive anymore


sahil-sharma:
Greetings Bosun Team,

Let's assume I have five container ID's in Bosun items (sending metrics via scollector). And, after some time my two containers were down and replaced by two new containers (obviously with new container ID's) and are appearing in Bosun items dashboard but the previous container ID's are still being visible in the dashboard. Isn't cumbersome when you have 100+ containers? How can we differentiate between the alive and dead containers?
Kindly do the needful or update me if there is any option.
Thanks in advance!

gbrayut:
Lack of datapoints usually means the container isn't running. After a while the tags will stop showing up in the autocomplete drop down (See https://github.com/bosun-monitor/bosun/issues/1969 for some details) but we don't have any sort of data retention or removal policies.

Also the container monitoring in Bosun is pretty basic. We don't use them much here so there are probably some rough edges in that area.

EricFortin:
While it helps on the UI side, the cache in Bosun keeps growing forever and backup to redis gets bigger and bigger with time. This is not so much an issue for size as it is for number of operations I think. Bosun memory consumption also expand with time as it internally copies the data before backing it up. Is there some way to age data ?

bobais:
Hi, I have related requirement. Possibility to delete old data - lets say 2yo+ data are not interesting anymore and in order to have dataset as small as possible I would like to remove them. Is there any way how to do it? Is there some feature request I can vote for? Thanks in advance.

kylebrandt:
This newish option in master takes care of the majority of memory growth

http://bosun.org/system_configuration#maxrenderedtemplateage

PR #2362, opened 2019-01-08
cmd/bosun: purge old incident state

### LINKED ISSUE #2363: Incident state stored forever
Once an incident is raised and closed, its state is stored forever.  This can chip away at memory available in Redis over time.  There should be a mechanism to cleanup old `incidentById:*` keys, similar to what `MaxRenderedTemplateAge` provides for rendered templates.

neilfordyce:
Based on the ideas introduced in https://github.com/bosun-monitor/bosun/pull/2167, this allows purging of keys under incidentById:*.  State for closed incidents which are beyond retention will be removed if the MaxClosedIncidentAge option is specified.

This fixes #2363.  It's discussed in #1974.

### DIFF
--- a/cmd/bosun/conf/conf.go
+++ b/cmd/bosun/conf/conf.go
@@ -71,6 +71,7 @@ type SystemConfProvider interface {
 	GetAuthConf() *AuthConf
 
 	GetMaxRenderedTemplateAge() int
+	GetMaxClosedIncidentAge() int
 
 	GetExampleExpression() string
 
--- a/cmd/bosun/conf/system.go
+++ b/cmd/bosun/conf/system.go
@@ -67,6 +67,7 @@ type SystemConf struct {
 	AuthConf *AuthConf
 
 	MaxRenderedTemplateAge int // in days
+	MaxClosedIncidentAge   int // in days
 
 	EnableSave      bool
 	EnableReload    bool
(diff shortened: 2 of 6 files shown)


kylebrandt:
I think the PreviousIds will need to be walked and and any links to purged incident states will need to be updated.  Do you have any ideas on how much memory this taking for you? Moving the rendered templates out was the bulk of it for us.

neilfordyce:
Thanks for the feedback.  

We have a Bosun instance which has been running for around 2 years.  We deployed the changes in this PR with a six month closed incident retention and reclaimed about 0.5GB.  For comparison, when we added a one month rendered template retention we removed about 1GB.  We decided to add this after doing a `--bigkeys` scan on our Redis DB and finding a single incidentById key which was almost 50MB.  We have a few alerts with high cardinality groups, which causes the "Events" list in the incident blob to blow up.

There is an attempt to cleanup links to deleted incidents in `deleteIncidentReferences()`, but maybe I missed something else.  This should take care of the references returned by `GetAllIncidentIdsByAlertKey()`.  Please let me know if there is more to cleanup and I can have a look.

kylebrandt:
@neilfordyce I don't think that covers it ... (although relational stuff in redis is confusing).

models.IncidentState has `PreviousIds []int64` property. This property contains a list of all incident ids that came before it. So an incident that hasn't been removed would have a incident entry in that link that is a dead link unless the incidents that are not getting deleted have that property updated.

neilfordyce:
Thanks for the feedback.  I added some cleanup for the references in PreviousIds of alerts with the same alert keys.

Issue #2259, opened 2018-05-26
elasticsearch collector only submits index data if scollector is run on a master node


PeterGrace:
In a docker/kubernetes environment where elasticsearch may be running inside of a docker container, it doesn't make sense to also run scollector inside of every docker container.  As a result, and in correlation to  PR #2257, the elasticsearch collector should be amended to be able to pull index data from non-master hosts.

PR #2368, opened 2019-01-13
cmd/scollector: pull index data from non-master hosts

Jakski:
Related to issue #2259.
I think, there's no reason to require target node to be master - correct me, if I'm wrong. I have setup, where masters have no HTTP interface for querying statistics, so ability to collect data from non-master hosts is crucial.


Jakski:
@kylebrandt Thank you for explanation. I added option to force collecting cluster-wide metrics(including indices) when current node isn't master. By default it retains old behavior.

PR #2370, opened 2019-01-16
 cmd/bosun: test validity of outputed promql

kylebrandt:


kylebrandt:
@captncraig enhance tests a little for review... I know this makes for a big vendor but I'll use the same vendor for raw style support anyways.

PR #2373, opened 2019-01-16
cmd/bosun: (wip) raw promql query support

kylebrandt:
Example:

```
$step = 1m
promras(''' sum(rate(container_fs_reads_total[$step]) + rate(container_fs_writes_total[$step])) by (namespace) ''', "2m", "2h", "")
```

top level expression must be an aggregation operation with `by ()`.

Todo:
 - [x] Docs
 - [x] Reduce code duplication

Follow up from conversation with @pdf in https://github.com/bosun-monitor/bosun/pull/2366

### DIFF
--- a/cmd/bosun/expr/prom.go
+++ b/cmd/bosun/expr/prom.go
@@ -10,6 +10,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/prometheus/prometheus/promql"
+
 	"bosun.org/cmd/bosun/conf/template"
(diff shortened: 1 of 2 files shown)


pdf:
Thanks for the super-speedy addition @kylebrandt <3

I have some minor corrections, should I comment here first to confirm or send a PR?

kylebrandt:
@pdf PR would be good since this has already been merged :-) Thanks for reviewing.

PR #2377, opened 2019-01-19
Minor fixes for prom(m)ras

pdf:
Some very small changes.


pdf:
I was also thinking that the multi calls could be made to fail early, but not without some small refactoring, and I'm not sure it's particularly important.

Issue #1895, opened 2016-09-06
Memory leak


efficks:
I'm observing a memory leak on Bosun. I'm using it with the internal redis and get the data from graphite. 
Can the internal redis be the cause of the problem?
![image](https://cloud.githubusercontent.com/assets/8975620/18280313/a31f1ef6-7426-11e6-90ed-8b7bf340dbf9.png)

Is there a setting I'm missing to clear the memory?

captncraig:
Does it drop again if you restart bosun? I believe the ledis store should be storing things on disk, and should not use unbounded memory.

efficks:
@captncraig Yes it does. It return to it's original memory usage, 221 Mib.

efficks:
I'm running Bosun on Ubuntu Server 12.04 LTS

captncraig:
ok, we should get a heap dump from you after it has run for a while. I'll have instructions shortly.

efficks:
@captncraig Thank you for your help.

captncraig:
Lets try this. Hopefully you have go installed. 

Run `go tool pprof http://your-bousn-server/debug/pprof/heap`

You should see soemthing like:
```
Fetching profile from http://ny-bosun01.ds.stackexchange.com/debug/pprof/heap
Saved profile in /Users/yourname/pprof/pprof.ny-bosun01.ds.stackexchange.com.inuse_objects.inuse_space.002.pb.gz
```
grab that zip file and upload it somewhere we can get it.

efficks:
@captncraig There it is. I runner bosun 20 hours and the memory gets from 259 MiB to 283 MiB
Not a so big value but in one week it gets huge.
If it can help you, I'm also using the API  and I don't acknowledge any alarm.
Finally, I've made this test with a real redis server. The leak continue to be present.

It's the only problem I get before pushing this nice system in production. I got the problem on two independent platforms running Ubuntu server 12.04 LTS.

Thank you for your help.
[pprof.10.133.63.22:8070.inuse_objects.inuse_space.001.pb.gz](https://github.com/bosun-monitor/bosun/files/459238/pprof.10.133.63.22.8070.inuse_objects.inuse_space.001.pb.gz)

captncraig:
[...]

PR #2382, opened 2019-02-01
cmd/bosun/web: use DefaultServeMux for /debug/pprof requests

travishegner:
When trying to get a pprof profile of the bosun server, I discovered that the path to `/debug/pprof` was being redirected back to the homepage of our bosun instance.

[Apparently](https://stackoverflow.com/questions/19591065/profiling-go-web-application-built-with-gorillas-mux-with-net-http-pprof) the `gorilla/mux` web framework clobbers the path to pprof, so this line of code allows the `http.DefaultServeMux` to handle the url at `/debug/pprof`, and opens the bosun server back up for profiling.

### DIFF
--- a/cmd/bosun/web/web.go
+++ b/cmd/bosun/web/web.go
@@ -206,6 +206,9 @@ func Listen(httpAddr, httpsAddr, certFile, keyFile string, devMode bool, tsdbHos
 	var miniprofilerRoutes = http.StripPrefix(miniprofiler.PATH, http.HandlerFunc(miniprofiler.MiniProfilerHandler))
 	router.PathPrefix(miniprofiler.PATH).Handler(baseChain.Then(miniprofilerRoutes)).Name("miniprofiler")
 
+	//use default mux for pprof
+	router.PathPrefix("/debug/pprof").Handler(http.DefaultServeMux)
+
 	router.PathPrefix("/api").HandlerFunc(http.NotFound)
 	//MUST BE LAST!
 	router.PathPrefix("/").Handler(baseChain.Then(auth.Wrap(JSON(Index), canViewDash))).Name("index")


kylebrandt:
Would the pprof tool still work with the client if required a token and was wrapped with `roleAdmin` ?

travishegner:
Sorry @kylebrandt, I'm not quite understanding your question.

Normally, one would just run the command `go tool pprof http://bosun:8070/debug/pprof/heap` in order to analyze go's current runtime heap, for example. I'm not certain that the `go tool` command currently allows for passing it's http request with a token, or any other form of authentication for that matter.

I suppose it could be possible to add an (authenticated) option in the web interface somewhere to toggle the url handler for `/debug/pprof` on and off, so that it can be disabled until needed at runtime, but, the `/debug/pprof` url itself would still be un-authenticated while it was enabled.

Judging by some older issues (#1895), and the fact that `main.go` is importing `_ net/http/pprof`, I was submitting this pull request under the impression that the `/debug/pprof` was intended to be available (unathenticated) by default. I then further assumed that a relatively recent switch to the `gorilla/mux` multiplexer simply clobbered the url that the blind import sets up by default.

If the `/debug/pprof` information is deemed too sensitive (I can't make that judgement call, or recommend anything here, I'm still learning what is available in it) to allow the unauthenticated url open by default (as it was open in past versions), then by all means disregard this request. I'd recommend also not importing `_ net/http/pprof` as well, to make it clear that you don't want to expose that url at all. Then you'll simply have to require that someone wanting to run the profiling tools against the bosun server builds their own version which imports the `pprof` packages and adds a router/handler for the pprof url.

kylebrandt:
That was before we added auth to Bosun. So this idea is more of a nice to
have if it is easy to add a header to the go tool pprof tool.

travishegner:
I don't believe it's possible to add a header to `go tool pprof`. It does simply work with files though, so it'd likely be trivial to just use `curl` with an auth header or token to pull down the profiling snapshot, and then just open the snapshot with the `go tool` command.

We don't run authentication on our instance of bosun, so that's a little out of my scope to try and implement.

PR #2387, opened 2019-02-14
Redis sentinel support for bosun

svagner:
# Problem
Currently we can have only one redis server. If we want to have some HA for bosun redis only way that we have HAProxy. But it require another service installation. So support of sentinels will allow us to have HA for redis without any extra services.

# Implementation
 Added 2 new options - `RedisSentinels` (array), `RedisMasterName`. If `RedisMasterName` was set we try to connect to redis sentinels (`RedisSentinels` option) for receive current redis master.

### DIFF
--- a/cmd/bosun/conf/conf.go
+++ b/cmd/bosun/conf/conf.go
@@ -42,7 +42,8 @@ type SystemConfProvider interface {
 	GetEmailFrom() string
 	GetLedisDir() string
 	GetLedisBindAddr() string
-	GetRedisHost() string
+	GetRedisHost() []string
+	GetRedisMasterName() string
 	GetRedisDb() int
 	GetRedisPassword() string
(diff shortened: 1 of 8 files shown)


svagner:
@captncraig , could you please review this mr?:)

svagner:
@kylebrandt , could you please review it as well? We can't create normal HA architecture for bosun installation without this patch:(

kylebrandt:
@svagner From a quick read it looks like this only changes behavior if you set these options, otherwise everything is as is (non-breaking change)? Did you try that out to make sure?

svagner:
@kylebrandt Yes, it isn't breaking changes:) I've rechecked those changes with 3 test configuration - ledis, redis stand alone, redis+sentinel. All tests were passed:)

PR #2396, opened 2019-04-05
cmd/bosun: pass alert variables to unknown notification templates

harudark:
#2275 allows to pass alert variables to action notifications. 
This PR allows this for unknown and multiple unknown notifications.

### DIFF
--- a/cmd/bosun/conf/unknownNotify.go
+++ b/cmd/bosun/conf/unknownNotify.go
@@ -11,11 +11,16 @@ import (
 	"bosun.org/slog"
 )
 
+type UnknownNotificationIncidentState struct {
+	*models.IncidentState
+	AlertVars Vars
+}
+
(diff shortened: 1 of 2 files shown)


stepashka:
hey, @kylebrandt , @captncraig 
do you have any objections for merging this? we'd greatly appreciate that

svagner:
It would be very useful to have this change

svagner:
@muffix, @tdinucci We working with this change for a while and didn't noticed any problems. Do you have any objectives for this change? Can we merge it with the main?

tdinucci:
@svagner we won't be able to accept this PR as it stands because there are no tests. We will still take a look though and consider it as soon as we can.

svagner:
@tdinucci , thanks for the explanation! It makes great sense:)

PR #2397, opened 2019-04-08
Change errors page

svagner:
Current page can contains a lot of data, so it calls problem with errors while page loading.
New page contain only errors without messages. Error messages are available on separate page.

Also this commit add new loading page indicator.


svagner:
Looks like there is failing generate of static files. I'm assuming that it is related with version that is installed on my machine...

PR #2403, opened 2019-04-17
basic esv7 support

langerma:
i played around with the code and added basic esv7 support


harudark:
@langerma, CI failed because it couldn't find ES client package.
it looks like ES client v3, v5, v6 package in vendor have been removed in c476432.
Also for v6 and v7, may I know why do you use gopkg.in url instead of github.com?
Thank you.

langerma:
it was just a quick and dirty hack. and i do not really know how to use the vendoring thingy in go so i just went with the es client versions from gopkg

PR #2413, opened 2019-06-10
cmd/bosun: Add support for multi-file configs

12yuens2:
Adds support for using a config-directory to store multiple config files.

Config files are loaded in alphabetical order and includes subdirectories. Only `.conf` files are included. This allows prefixing config files with numbers for exact ordering. For example:
```
config-directory/00-globals.conf
config-directory/01-macros.conf
config-directory/02-alerts/service1.conf
config-directory/02-alerts/service2.conf
config-directory/03-monitoring.conf
```

The web UI has also been updated to show the concatenated config with comments separating the different files. 
```
### FROM config-directory/00-globals.conf

template header {
    body = ...
}

### END config-directory/00-globals.conf

### FROM config-directory/01-macros.conf

macro common {
    warnNotification = ...
}

### END config-directory/01-macros.conf

```
The ability to save the config has been updated so that saving on the web UI saves to the corresponding config file. Config files that are outside the bosun working directory, do not already exist and do not have a `.conf` file extension will not be saved.


dwo:
@muffix @johnewing1 do we have plans to get this in here? Was wondering what the status of this was...

dwo:
I noticed yesterday that when using `bosun -t` to test a (deliberately broken) config file, using this patch gives slightly less useful output:
```
+ /bosun/bosun -c test.toml -t
2020/03/02 17:09:15 fatal: main.go:150: couldn't read rules: conf: /src/bosun-instance/bosun-config-directory:130:4: at <$deliberate = $break...>: unknown variable $breaking_of_config
```

Ideally this would tell you exactly which file and line the problem was at.

Issue #2425, opened 2019-07-25
ES 7.x support and go modules introduction


harudark:
I'd like to add ElasticSearch 7.x support in Bosun. However, ES v7 client(https://github.com/olivere/elastic) requires to use go modules.

I tried to use govendor to add ES v7 client but that would change files used by ES v6 client.
I also tried to use go modules and it works. But it looks like a big change for the project and I'm not sure about how upstream or other people think about it.

Do you have any suggestion about this? Thank you.

johnewing1:
We've reworked the build to move to modules but we had to make a few other changes to the build steps so ended up doing so in our own PR. Would you mind rebasing #2427 from master so that it just represents the elastic changes. Thanks.

harudark:
@johnewing1, thanks for taking care of Bosun. 👍 
I'll rebase and let you know when I've finished.

johnewing1:
merged. Thanks for the contribution

PR #2426, opened 2019-07-30
build: migrate from govendor to go modules

harudark:
Migrate from govendor to go modules in order to support ES 7
See #2425


johnewing1:
This has been addressed in PR#2432

PR #2428, opened 2019-08-07
Support runtime ElasticSearch version query

harudark:
With this commit, Bosun is able to get Elastic cluster version by parsing simple HTTP query response, which means it can get version during runtime, rather than the value defined in configuration file.

This feature is useful during Elastic cluster migration. Bosun can keep working without downtime.

However, due to the fact Bosun parses configuration file very frequently, it's not suggested to enable this all the time because it will send a lot of additional queries to Elastic cluster.


harudark:
This PR doesn't work in production.

PR #2432, opened 2019-11-12
Migrate to Go modules

muffix:
Migrates the project's build system to Go modules


README.md
```diff
@@ -2,24 +2,18 @@
 
 Bosun is a time series alerting framework developed by Stack Exchange. Scollector is a metric collection agent. Learn more at [bosun.org](http://bosun.org).
 
-[![Build Status](https://travis-ci.org/bosun-monitor/bosun.svg?branch=master)](https://travis-ci.org/bosun-monitor/bosun/branches)
+[![Build Status](https://travis-ci.org/bosun-monitor/bosun.svg?branch=master)](https://travis-ci.org/bosun-monitor/bosun)
```
johnewing1: What's the reason for changing the build status badge ?

build/validate.sh
```diff
@@ -102,9 +97,5 @@ if [ "$BUILDMSG" == '' ]; then
 	BUILDSTATUS=success
 fi
 
-if [ "$TRAVIS" != '' ]; then
```
johnewing1: What's the thinking behind removing the status reporting ? Have we replaced it elsewhere ?

build/validate.sh
```diff
@@ -69,7 +64,7 @@ fi
 
 echo -e "\nTesting that bosun starts and stops cleanly"
 #TODO: save linux build from above? cant seem to find where it is though.
-cd $GOPATH/src/bosun.org/cmd/bosun
+cd cmd/bosun
```
johnewing1: Is there a reason to remove the full path from GOROOT rather than rely on relative pathing ?

build/validate.sh
muffix: The Travis integration should already do that for us: https://blog.travis-ci.com/2018-05-07-announcing-support-for-github-checks-api-on-travis-ci-com

IIRC, setting the status with an API call was required before Travis moved to Github apps.

build/validate.sh
muffix: The only idea was to move out of the GOPATH since we're now using Go modules. Happy to add in an absolute path if you think that's clearer. 👍

johnewing1: Tried this locally and got 
validate.sh: line 67: cd: cmd/bosun: No such file or directory
because I ran it from within the build directory
Probably worth putting the full path back in so that you can test the full linter setup locally without having to run from a specific directory

README.md
muffix: Thanks, that was accidentally committed. Fixed and rebased the branch on the latest master.

PR #2433, opened 2019-11-15
Remove the deprecation notice

johnewing1:
Remove the deprecation notice and update with announcement of new maintainers

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -21,7 +21,7 @@ order: 1
 	<div class="col-md-12">
 		<div class="admonition tip">
 				<p class="admonition-title">Project Status (2019-11)</p>
-				<p><a href="skyscanner.net">Skyscanner</a> is delighted to announce that as of 18th November 2019, we'll be officially taking over Bosun's ownership and maintainership from <a href="http://stackexchange.com">Stack Exchange</a> (home of stackoverflow.com).</p>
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -19,9 +19,12 @@ order: 1
 
 <div class="row">
 	<div class="col-md-12">
-		<div class="admonition warning">
-				<p class="admonition-title">Project Status (2019-02)</p>
-				<p>Bosun is now in maintenance/sunsetting status and is looking for new maintainer(s), please see <a href="/call">this notification</a>.</p>
+		<div class="admonition tip">
+				<p class="admonition-title">Project Status (2019-11)</p>
+				<p><a href="skyscanner.net">Skyscanner</a> is delighted to announce that as of 18th November 2019, we'll be officially taking over Bosun's ownership and maintainership from <a href="http://stackexchange.com">Stack Exchange</a> (home of stackoverflow.com).</p>
```
hermanschaaf: Could be wrong, but I think this should be `https://skyscanner.net` otherwise browsers may interpret it as a relative link?

dwo: `https://www.skyscanner.net/` if you want to avoid a needless redirect.

johnewing1: Updated

PR #2456, opened 2020-02-25
Show parentheses when outputting expressions

### LINKED ISSUE #2455: Feature request: show parentheses in formatted expressions
# Short description

When showing evaluated expressions (e.g. in the computations column when
testing an expression or an alert), the expressions are ambiguous (and
misleading) because they include no parentheses, even where they are
necessary.

E.g. evaluate the following expression:

```
(1 + 2) / 3
```

and the computations column will show:

```
1 + 2 / 3 = 1
1 + 2     = 3
```

Compare to evaluating the expression:

```
1 + (2 / 3)
```

which will show:

```
1 + 2 / 3 = 1.66667
2 / 3     = 0.66667
```

Here's a screenshot of the first example. There are no parentheses shown even though they are necessary to understand the expression.

image

# How this feature will help you/your organisation
This confuses engineers trying to debug their expressions. Since the macro-style expansion of variables is already confusing (see #635), this compounds their difficulty in understanding whether their expression is being handled as they intended.

# Possible solution or implementation details
Always show parentheses around binary expressions. (I'll submit a PR.)

annettejanewilson:
# Description

## Problem

When showing evaluated expressions (e.g. in the computations column when
testing an expression or an alert), the expressions are ambiguous (and
misleading) because they include no parentheses, even where they are
necessary.

E.g. evaluate the following expression:

```
(1 + 2) / 3
```

and the computations column will show:

```
1 + 2 / 3 = 1
1 + 2     = 3
```

Compare to evaluating the expression:

```
1 + (2 / 3)
```

which will show:

```
1 + 2 / 3 = 1.66667
2 / 3     = 0.66667
```

## Solution

The simplest fix is to wrap all binary expressions in parentheses. This
_does_ emit redundant parentheses, but it's much simpler than figuring
out when precedence rules allow the safe elision of parentheses. So for
example this expression:

```
1 + 2 + 3 + 4
```

will be rendered:

```
(((1 + 2) + 3) + 4)
```

I think any extra verbosity here is worth paying for being explicit and
unambiguous.

An alternative solution would be to apply the parentheses in the
_parent_ node - unconditionally if the parent node is a unary expression
and the child is a binary expression, and conditionally if the parent
node is a binary expression and the child is a binary expression: in
this case parentheses are required on the left child node if it binds
less tightly than the parent's operator, and are required on the right
child node if it binds less or equally tightly than the parent's
operator. I have now implemented this and added it to the PR.

Fixes #2455 - Feature request: show parentheses in formatted expressions

## Type of change

From the following, please check the options that are relevant.

# How has this been tested?

Please describe the tests that you ran to verify your changes. 
Provide instructions so we can reproduce. 
Please also list any relevant details for your test configuration

- [x] Pasted the above expressions in the expression pane
- [x] Updated the expected results in the unit tests
- [ ] Test against live data in the rule editor (not done)
- [ ] Test anywhere else these expressions appear in the UI (not done)

# Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] I have ~added~ updated tests that prove my fix is effective or that my feature works
- [x] Any dependent changes have been merged and published in downstream modules


annettejanewilson:
I would still like to propose this. It solves a real problem.

PR #2461, opened 2020-03-03
Fix TestNewManager

### LINKED ISSUE #2460: Bug: TestNewManager failed on some build Linux machines
# Expected behavior

The unit tests should be able to finish successfully on any machines if code works correctly.
We should not expect that hostname for test Linux machine was configured with some particular rules because it can be not true:
- https://www.debian.org/doc/manuals/debian-reference/ch03.en.html#_the_hostname
- https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/networking_guide/ch-configure_host_names

# Current behavior
```
=== RUN   TestNewManager
    TestNewManager: manager_test.go:13: Expected hostname to be 'host01.mydomain.com' but it was 'host01'
--- FAIL: TestNewManager (0.00s)
```

## Steps to reproduce

1. set the hostname for Linux node as host01.mydomain.com
2. run `./build/validate.sh` script

## Context

```
# cat /etc/hostname
host01.mydomain.com
# cat /proc/sys/kernel/hostname
host01.mydomain.com
```

svagner:
## Please follow the guide below

# Description
Fix test `TestNewManager`
Use only full hostname for tests (see os.Hostname())

Fixes #2460

## Type of change

# How has this been tested?

- [x] TestNewManager

# Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] Any dependent changes have been merged and published in downstream modules


host/manager_test.go
```diff
@@ -19,7 +19,10 @@ func validateManager(m Manager, expectedHostname string, t *testing.T) {
 }
 
 func TestNewManager(t *testing.T) {
-	m, err := NewManager(false)
+	// we should save fqdn for NewManager since os.Hostname will return /proc/sys/kernel/hostname and it can be fqdn
+	// underhood NewManager will use same call, but if we don't save fqdn we are trying to split that name
+	// In that can this test will fail anyway
+	m, err := NewManager(true)
```
muffix: `os.Hostname()` calls `sysctl kern.hostname` - It only corresponds to the contents of `/proc/sys/kernel/hostname` on some Linux systems and may or may not do so in the future. 

I agree that the test should pass, but I'm a bit worried that this is changing the semantics of the test. This is now not testing the splitting anymore. Should the expected host name be different instead?

tdinucci:
Yes, thanks for the PR 👍 

I think we probably need two tests actually:

* One where `preserveFullHostName == true`
* One where `preserveFullHostName == false` but this test must then split the copy of `expectedHostname` in the test on the first `.` (which the current test isn't doing).

If you would like to update your PR to cover both of these scenarios that would be great.  Otherwise I'll be happy to do this.

host/manager_test.go
svagner: I've rewritten test to make check more predictable:)

tdinucci:
Once you bring your branch inline with master we should be able to merge

svagner:
Thanks @tdinucci ! Branch was updated:)

PR #2481, opened 2020-05-13
Add test coverage report generation

johnewing1:
Add targets for generating coverage reports. Initial version of sonar config to exclude tests and vendored packages

## Please follow the guide below

- You will be asked some questions, please read them **carefully** and answer honestly
- Put an `x` into all the boxes [ ] relevant to your *pull request* (like that [x])
- Use *Preview* tab to see how your *pull request* will actually look like

---

# Description

Add two features.
Short cut for generating and viewing coverage reports by running `make coverage`
Added step to generate coverage report as part of checks for use by sonarcloud

## Type of change

From the following, please check the options that are relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

# How has this been tested?
Makefile steps verified locally

# Checklist:

- [x ] This contribution follows the project's [code of conduct](https://github.com/bosun-monitor/bosun/blob/master/CODE_OF_CONDUCT.md)
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -125,8 +125,8 @@ coverage:
 	$(GOTEST) -covermode=count -coverprofile=coverage.out  $$($(GOLIST) ./... | grep -v integration)
 	$(GOTOOL) cover -html=coverage.out
 
-.PHOHY: checks
-checks: goimports-check vet generate tidy-check test-coverprofile
+.PHONY: checks
+checks: goimports-check vet generate tidy-check
 
 .PHONY: clean
 clean:
--- a/Makefile
+++ b/Makefile
@@ -116,8 +116,17 @@ staticcheck:
 test:
 	$(GOTEST) -v ./...
 
+.PHONY: test-coverprofile
+test-coverprofile:
+	go test -covermode=count -coverprofile=coverage.out `go list ./... | grep -v integration` -json > test-report.out
+
+.PHONY: coverage
+coverage:
+	go test -covermode=count -coverprofile=coverage.out `go list ./... | grep -v integration`
+	go tool cover -html=coverage.out
+
 .PHOHY: checks
-checks: goimports-check vet generate tidy-check
+checks: goimports-check vet generate tidy-check test-coverprofile
 
 .PHONY: clean
 clean:
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ GOLIST=$(GOCMD) list
 GOMOD=$(GOCMD) mod
 GOTEST=$(GOCMD) test
 GOVET=$(GOCMD) vet
-
+GOTOOL=$(GOCMD) tool
 BINARY_NAME=bosun
 BINARY_UNIX=$(BINARY_NAME)_unix
 BINARY_MAC=$(BINARY_NAME)_mac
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -111,8 +111,17 @@ staticcheck:
 test:
 	$(GOTEST) -v ./...
 
+.PHONY: test-coverprofile
+test-coverprofile:
+	go test -covermode=count -coverprofile=coverage.out `go list ./... | grep -v integration` -json > test-report.out
```
muffix, line 116: The Makefile defines `GOTEST` and `GOLIST` as a variables so you could use that here to take advantage of the global settings (if we make any).
Might be worth getting rid of the legacy backticks in the command (https://github.com/koalaman/shellcheck/wiki/Sc2006)

```suggestion
	$(GOTEST) -covermode=count -coverprofile=coverage.out $$($(GOLIST) ./... | grep -v integration) -json > test-report.out
```

Makefile
```diff
@@ -111,8 +111,17 @@ staticcheck:
...
+
+.PHONY: coverage
+coverage:
+	go test -covermode=count -coverprofile=coverage.out `go list ./... | grep -v integration`
```
muffix, line 120:
```suggestion
 	$(GOTEST) -covermode=count -coverprofile=coverage.out $$($(GOLIST) ./... | grep -v integration)
```

Makefile
```diff
@@ -111,8 +111,17 @@ staticcheck:
...
+	go tool cover -html=coverage.out
```
muffix, line 121: Worth defining a `GOTOOL` variable at the top as we did for the others?

Makefile
```diff
@@ -111,8 +111,17 @@ staticcheck:
...
+
 .PHOHY: checks
-checks: goimports-check vet generate tidy-check
+checks: goimports-check vet generate tidy-check test-coverprofile
```
muffix, line 124: Not sure if it really fits into the `checks` target here. Should this be the one that `doesn't` generate the output file instead?
Do we need to add the one that does to the Github CI action?

johnewing1: Both new actions generate an output file. 
test-coverprofile creates a json one which sonar cube can understtadnd
coverage creates one which the go cover tool can use

johnewing1: We should also add the CI step, working out how to do that

Makefile
johnewing1: updated

Makefile
johnewing1: done

Makefile
johnewing1: done

Makefile
```diff
@@ -111,8 +111,17 @@ staticcheck:
...
+.PHONY: test-coverprofile
+test-coverprofile:
+	go test -covermode=count -coverprofile=coverage.out `go list ./... | grep -v integration` -json > test-report.out
+
+.PHONY: coverage
+coverage:
+	go test -covermode=count -coverprofile=coverage.out `go list ./... | grep -v integration`
+	go tool cover -html=coverage.out
+
 .PHOHY: checks
```
johnewing1, line 123: Noticed a typo here from the original makefile H/N 

Would it be cleaner to have a single PHONY directive at the top of the file ?
Reduces repetition

muffix: Yeah, maybe. I don't really mind. 👍

Makefile
muffix: Ah, right. I'm still ot convinced it really fits into the `checks` target which so far only calls targets that perform static analysis and exit with a non-zero exit code on failure. The coverprofile would just create a file, but exit with 0 either way.

johnewing1: Good point, I've removed it from that target.
Also added what I think is the required CI step

PR #2497, opened 2020-11-09
Update the documentation for dropbool

### LINKED ISSUE #2496: Bug: dropbool() documentation inconsistent with behaviour
# Context

The documentation for `dropbool()` currently states
> Drop datapoints where the corresponding value in the second series set is non-zero. (See Series Operations for what corresponding means).

The name of the function suggests that the described behaviour is the intended one, but the behaviour is inconsistent with the documentation.
Currently, the points are **kept*( (rather than dropped) where the corresponding value in the second series set is non-zero.

We should update the documentation to describe the actual behaviour and consider a breaking change that gives the function the desired (and from the name more intuitive behaviour) in a future version.

# Steps to reproduce

1. Run the following query: `dropbool(series("foo=bar", 0, 1, 1, 1, 2, 1, 3, 1), series("", 0, 1, 1, 2, 2, 3, 3, 2) != 2)`
2. Observe the output: `{"0": 1, "2": 1}` - expected is `{"1": 1, "3": 1}` with the current documentation.

muffix:
---

# Description

Fixes the documentation which described the behaviour of `dropbool`
incorrectly.
`dropbool()` currently drops the values where the value in the second
series set IS zero instead of, as described, dropping those where the
value IS NOT zero. To avoid a breaking change which would give the
function the more intuitive behaviour from its name, this only updates
the documentation.

We should consider fixing this with a breaking change in a future
release.

Fixes #2496

## Type of change

- [x] This change requires a documentation update

# How has this been tested?

- Query `dropbool(series("foo=bar", 0, 1, 1, 1, 2, 1, 3, 1), series("", 0, 1, 1, 2, 2, 3, 3, 2) != 2)` returns `{"0": 1, "2": 1}`
- Query `dropbool(series("foo=bar", 0, 1, 1, 1, 2, 1, 3, 1), series("", 0, 1, 1, 2, 2, 3, 3, 2))` returns `{"1": 1, "2": 1, "3": 1}`

The behaviour is the inverse of the previously described behaviour. The description is fixed with this PR.

# Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/docs/expressions.md
+++ b/docs/expressions.md
@@ -1130,7 +1130,7 @@ Remove any NaN or Inf values from a series. Will error if this operation results
 ## dropbool(seriesSet, seriesSet) seriesSet
 {: .exprFunc}
 
-Drop datapoints where the corresponding value in the second series set is non-zero. (See Series Operations for what corresponding means). The following example drops tr_avg (avg response time per bucket) datapoints if the count in that bucket was + or - 100 from the average count over the time period.
+Drop datapoints where the corresponding value in the second series set is zero. (See Series Operations for what corresponding means). The following example drops tr_avg (avg response time per bucket) datapoints if the count in that bucket was + or - 100 from the average count over the time period.
 
 Example:
 


neilfordyce (approved):
Well spotted.  dropbool keeps things where it's non-zero and drops where it is zero.

PR #2523, opened 2024-05-29
Update project status message on bosun.org

neilfordyce:
# Description
A long overdue update for https://bosun.org/


chlos (approved):
Update the status as at 2024-05

PR #2526, opened 2025-02-13
Update project status page

neilfordyce:
So long Bosun. Thanks for helping us steady the ship.


danielgblanco (approved):
Thanks Neil, and to everyone that contributed to this project over the years!

